PR #2, opened 2017-04-19
GT-263

ryancarlson:
- Adds in initial core data model
- Sets basic architecture model (controllers & managers)
- Loads languages from remote and stores them in local datastore
- LanguagesManager provides two method for retrieving, `loadFromDisk()`, `loadFromRemote()`


ryancarlson:
@pablomarti this PR will allow you to use the `LanguagesManager` to retrieve languages from the server. Hay tres actualmente. Once the languages are retrieved using `loadFromRemote()` you can then access them from the device by calling `loadFromDisk()`

I just added the `LanguagesViewController` to be able to visualize the data and trigger the loads. You can use it or not use it. 

I would like to follow roughly this project structure of separating models views and controllers in separate groups, and then specifying sub groups like "Languages" as we go.

I would also like to use `LanguagesManager` as the delegate and datasource for the table view to keep the view controllers smaller and have that code be more portable.

Podfile
```diff
@@ -11,7 +11,7 @@ target 'godtools' do
   pod 'Alamofire', '~> 4.0'
```
pablomarti: Can we update to 4.4?

godtools/AppDelegate.swift
```diff
@@ -16,7 +17,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
-        // Override point for customization after application launch.
+        MagicalRecord.setupCoreDataStack(withAutoMigratingSqliteStoreNamed: "godtools-5.sqlite")
```
pablomarti: This should only happen once, right, but what do you think about moving this to a method (always on AppDelegate) and calling the method from here? (I believe we should separate the things in order to have atomic methods)

godtools/Base.lproj/Main.storyboard
```diff
@@ -1,22 +1,39 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
pablomarti: Just for the record, I removed this file, and also from the info.plist

godtools/Controllers/LanguagesViewController.swift
```diff
@@ -0,0 +1,48 @@
...
+        didSet {
+            tableView.delegate = languagesManager
+            tableView.dataSource = languagesManager
+        }
+    }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        languagesManager.loadFromDisk().catch(execute: { error in
```
pablomarti: This is really cool! Would you please move each load to a separate method? Also, I believe we should create something for error handle, in order to make it more easy and automatic

pablomarti: Well nevermind about the loads, this is an example

godtools/GodToolsConstants.swift
```diff
@@ -0,0 +1,14 @@
+//
```
pablomarti: I also made a file for constants, but, instead of making a class I made a `struct`, what do you think? Also, the file I created is called `GTConstants`; I'm adding the prefix `GT` to every file, can we refactor and add the prefix `GT` to every file?

godtools/Managers/LanguagesManager.swift
```diff
@@ -0,0 +1,79 @@
...
+    
+    var languages = [Language]()
+    
+    override init() {
+        super.init()
+        serializer.registerResource(LanguageResource.self)
+    }
+    
+    func loadFromDisk() -> Promise<[Language]> {
+        languages = Language.mr_findAll() as! [Language]
```
pablomarti: I like all of this! really easy to follow, I believe we can make an "abstract" class, so a lot of these functionalities can be set there, and it should only call and pass the arguments for process the data (when retrieving with MagicalRecord, and when calling to the server with AlamoFire). That would help when making the other managers, it is going to be easier and faster, what do you think?

godtools/Managers/LanguagesManager.swift
```diff
@@ -0,0 +1,79 @@
...
+        
+        languages = languages.sorted { (language1, language2) -> Bool in
+            return language1.localizedName!.compare(language2.localizedName!).rawValue < 0
+        }
+        
+        return Promise(value:languages)
+    }
+    
+    func loadFromRemote() -> Promise<[Language]> {
+        UIApplication.shared.isNetworkActivityIndicatorVisible = true
```
pablomarti: Would you please set this into a method?

godtools/Managers/LanguagesManager.swift
```diff
@@ -0,0 +1,79 @@
...
+        
+        return Alamofire.request(URL(string: "\(GodToolsConstants.kApiBase)/\(path)")!).responseData().then { data -> Promise<[Language]> in
+            let jsonDocument = try! self.serializer.deserializeData(data)
+            
+            if jsonDocument.data != nil {
+                MagicalRecord.save(blockAndWait: { (context) in
```
pablomarti: As I wrote before, I believe all of this logic can be re-usable by adding it in an "abstract" class. I like how the protocols and extensions are encouraged by the community of Swift 3, thats a good way to make things separated and re-usable

godtools/Managers/LanguagesManager.swift
```diff
@@ -0,0 +1,79 @@
...
+            return self.loadFromDisk()
+        }
+    }
+}
+
+extension LanguagesManager: UITableViewDelegate {
+    
+}
+
+extension LanguagesManager: UITableViewDataSource {
```
pablomarti: I like the extensions, could you please set them in separated files? So in the future we will have a group for the extensions for the datasources and delegates, separated by each resource

Podfile
ryancarlson: Yes, good idea!

godtools/Base.lproj/Main.storyboard
ryancarlson: 👍

godtools/AppDelegate.swift
ryancarlson: Okay, I did this, but I'm not sure it's totally necessary.

ryancarlson: MagicalRecord already nicely puts the setup and cleanup functions into one atomic function call.

godtools/Controllers/LanguagesViewController.swift
ryancarlson: Right. This is just an example to show how the calls work. I don't know if you plan to keep this VC or not.

godtools/GodToolsConstants.swift
ryancarlson: Sure it could be a struct. I'll change that.

Do we need to prefix all the files with GT? I know that is common practice in older iOS apps, but is it still necessary? I find it clutters the filenames and makes them harder to read.

For the ones that do need prefixed, I do agree that GT is better than GodTools, so I'll change that for this file.

godtools/Managers/LanguagesManager.swift
ryancarlson: Yes. As I wrote the second manager class, I realized the code was similar in many ways to the first.

I don't like to prematurely abstract though because it can actually create more refactoring work in the long run. I typically like to extract super classes after 3rd or sometimes 2nd implementation because by that point it's easier to see the patterns in the architecture, and it finds that good point of saving time in the future w/o creating unnecessary refactoring.

So all of that to say, yes good idea.. but I'm not going to do it just quite yet :)

godtools/Managers/LanguagesManager.swift
ryancarlson: Yes, extensions are really nice!!

godtools/Managers/LanguagesManager.swift
ryancarlson: 👍

godtools/Managers/LanguagesManager.swift
ryancarlson: I think I actually like keeping these together. However if the files grow too large, then yes I am definitely willing to split them into separate files.

I do not want delegates and data sources to be in controller files any more. No more Massive View Controllers :)

godtools/GodToolsConstants.swift
pablomarti: I'm very old school haha, I set the GT on all files, but let me know if I should change it

ryancarlson:
Thanks for your review @pablomarti ! I agree with most of your changes. I made some changes and pushed them, and I explained there are some I want to do in the future, but I want to let the project evolve a little bit more before doing abstractions.

Can you look at my commits and responses and let me know if this is ready for merging? Gracias mi amigo!

godtools/Managers/LanguagesManager.swift
pablomarti: Ok sounds good! Good idea

godtools/GTConstants.swift
```diff
@@ -8,7 +8,6 @@
 
 import Foundation
 
-class GodToolsConstants {
-    
-    static let kApiBase = "https://mobile-content-api-stage.cru.org"
+struct GTConstants {
+    let kApiBase = "https://mobile-content-api-stage.cru.org"
```
pablomarti: You can use `static` here too

godtools/Managers/LanguagesManager.swift
```diff
@@ -43,7 +43,7 @@ class LanguagesManager: NSObject {
     func loadFromRemote() -> Promise<[Language]> {
         showNetworkingIndicator()
         
-        return Alamofire.request(URL(string: "\(GodToolsConstants.kApiBase)/\(path)")!).responseData().then { data -> Promise<[Language]> in
+        return Alamofire.request(URL(string: "\(GTConstants().kApiBase)/\(path)")!).responseData().then { data -> Promise<[Language]> in
```
pablomarti: If you use `static` in the struct you are going to be able to access like `GTConstants. kApiBase`

godtools/GTConstants.swift
pablomarti: So it would be `static let kApiBase = "https://mobile-content-api-stage.cru.org"`

godtools/Managers/LanguagesManager.swift
ryancarlson: I thought I should be able to do that, but it gave me a build error the first time. It works now though. Thanks!

pablomarti:
Excellent @ryancarlson ! feel free to merge!

PR #4, opened 2017-04-21
GT-264

ryancarlson:
Allows syncing of translations and resources from the remote server.
Also cleans up the code structure a little bit more.

Translations are synced as includes on the Resources fetch and as relationships on the Languages fetch. The order of these fetches should not matter as we do a findOrCreate on the remoteId of that translation.

Note: This is using a forked version of the Spine pod. I submitted a PR and am hopeful it can/will be merged.


pablomarti:
I'm good with this, great @ryancarlson ! I don't know if you want to wait to see the response to your PR on Spine?

ryancarlson:
I'd like to move ahead for now, but yes, we should keep an eye on that PR. Thanks for the review!

pablomarti:
Ok great feel free to merge!

ryancarlson:
Gracias por revisarlo!

PR #6, opened 2017-04-21
Setup of style for platform home screen and add new tool

pablomarti:


godtools/ViewControllers/BaseViewController.swift
```diff
@@ -10,14 +10,106 @@ import UIKit
...
+    
+    func displayNavigationLeftButtons() {
+        self.navigationItem.setLeftBarButtonItems(self.navigationLeftButtons.reversed(), animated: true)
+        navigationController?.navigationBar.topItem?.title = ""
+    }
+    
+    func displayNavigationRightButtons() {
+        // Set 26px of fixed space between the two UIBarButtonItems
+        let fixedSpace:UIBarButtonItem = UIBarButtonItem(barButtonSystemItem: UIBarButtonSystemItem.fixedSpace, target: nil, action: nil)
+        fixedSpace.width = 26.0
```
ryancarlson: Should we make this a constant value in this file?

godtools/ViewControllers/BaseViewController.swift
```diff
@@ -10,14 +10,106 @@ import UIKit
...
+        
+        // Set -7px of fixed space before the two UIBarButtonItems so that they are aligned to the edge
+        let negativeSpace:UIBarButtonItem = UIBarButtonItem(barButtonSystemItem: UIBarButtonSystemItem.fixedSpace, target: nil, action: nil)
+        negativeSpace.width = -7.0
```
ryancarlson: Should we make this a constant value in this file, too?

godtools/ViewControllers/Platform/LanguageSettingsViewController.swift
```diff
@@ -16,6 +16,12 @@ class LanguageSettingsViewController: BaseViewController {
     
     var delegate: LanguageSettingsViewControllerDelegate?
     
+    override var screenTitle: String {
+        get {
+            return "Language Settings"
```
ryancarlson: Let's make sure we don't forget to externalize strings. Can we start that now so we get most of them early on?

godtools/Views/Cells/HomeToolTableViewCell.swift
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    override func setSelected(_ selected: Bool, animated: Bool) {
+        super.setSelected(selected, animated: animated)
+    }
+    
+    func setCellAsDisplayOnly() {
+        self.downloadButton.isHidden = true
+        self.greyVerticalLine.isHidden = true
+        self.titleLeadingConstraint.constant = 8
```
ryancarlson: Could this have different values depending on the state of the cell? If not, then I like to keep constants set in the interface builder.

pablomarti:
Hey @ryancarlson would you please check the changes? Also, please take a look at the localization methods. I made an extension, .localized, which is working nice, but I got an idea for avoiding to have too many outlets on the viewcontrollers, and the classes of the components like GTButton or GTLabel automatically change the localization in the method `layoutSubviews`, so we only have to take care of setting the text on the .xib and to have the same text as a key on the file *Localizable.strings*, what do you think about it? (I found it very easy to handle). I'm still struggling with a method for localization with variables.

godtools/Localizable.strings
```diff
@@ -0,0 +1,24 @@
+/* 
+  Localizable.strings
```
ryancarlson: Interesting idea. Where did you find it? Or was it something you thought of?

pablomarti: I found a snippet for the extension, and about the methods on the label and buttons classes that is something I figured out. Do you feel comfortable with it?

pablomarti: (I did that in order to avoid to have too many IBOutlets)

ryancarlson: @frett and I just looked at this, and he shared one concern I think that's important enough to change this. We can run into trouble if we need to change or correct the phrase some day. If we had to change the key portion, then translators would have to translate the new phrase key, or we would have to do admin work in OneSky to update the key.

I really like the extension so we can just call `.localized`, but I think we need to use more traditional keys (`language_settings_title`)

ryancarlson:
hey @pablomarti looks good, other than the localization thing. once we have that in a good place I'd like to merge this so I can build off it with some other things.

godtools/Localizable.strings
pablomarti: Ok cool I will change it

ryancarlson: So in the xibs and the code, I think you could still do `"language_settigs_title"`.localized and I think that would be ok.

ryancarlson:
This looks good @pablomarti
!

pablomarti:
Thanks @ryancarlson !

PR #9, opened 2017-04-26
Platform menu

pablomarti:
Adds the menu with a side effect.


godtools/FlowControllers/BaseFlowController.swift
```diff
@@ -42,4 +43,55 @@ class BaseFlowController: NSObject, UINavigationControllerDelegate {
...
+                                               name: .displayMenuNotification,
+                                               object: nil)
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(dismissMenu),
+                                               name: .dismissMenuNotification,
+                                               object: nil)
+    }
+    
+    func displayMenu() {
+        let menuViewController = MenuViewController(nibName: "MenuViewController", bundle: nil)
```
ryancarlson: One thing I've been doing to create view controllers is using the String.describe initializer. It takes a class which I think will help prevent typos. It does require though that the xib filename match the view controller filename.. which it should anyways. So it's not 100% guarantee, but I think it helps.

In this case:
````swift
let menuViewController = MenuViewController(nibName: String(describing:MenuViewController.self), bundle: nil)
````

File: godtools/FlowControllers/BaseFlowController.swift [hunk 2]
```diff
@@ -42,4 +43,55 @@ class BaseFlowController: NSObject, UINavigationControllerDelegate {
...
+        let navigationController = self.currentViewController?.navigationController
+        let src = self.currentViewController
+        let dst = menuViewController
+        let srcViewWidth = src?.view.frame.size.width
+        
+        src?.view.superview?.insertSubview(dst.view, aboveSubview: (src?.view)!)
```
ryancarlson: If you unwrap your optionals along the way, then you don't have to unwrap the result. The same dangers of unwrapping apply, but the code is cleaner in my opinion. So you'd have here:
````swift
src?.view.superview?.insertSubview(dst.view, aboveSubview: src!.view)
````
(note the first `src` reference actually makes this statement safe b/c if it's nil the second one that we force unwrap would never get evaluated)

File: godtools/ViewControllers/MenuViewController.swift [hunk 3]
```diff
@@ -0,0 +1,136 @@
...
+    fileprivate func setupStyle() {
+        self.view.backgroundColor = .clear
+        self.tableView.backgroundColor = .gtGreyLight
+        self.registerCells()
+    }
+    
+    // MARK: - Navigation Buttons
+    
+    override func configureNavigationButtons() {
+        self.addEmptyLeftButton()
```
ryancarlson: Do we need a button here? What happens if there is no button?

godtools/ViewControllers/MenuViewController.swift
```diff
@@ -0,0 +1,136 @@
...
+        self.addDoneButton()
+    }
+    
+    // MARK: - UITableViewDataSource
+    
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
```
ryancarlson: If we are going to keep the delegate and datasource methods in this file, then please declare them in extensions.

In this case I think it makes sense to leave them because the possibility of reuse is very small.

godtools/ViewControllers/MenuViewController.swift
```diff
@@ -0,0 +1,136 @@
...
+    }
+    
+    // MARK: - Helpers
+    
+    fileprivate func getSection(_ section: Int) -> String {
+        return header[section]
+    }
+    
+    fileprivate func getSectionData(_ section: Int) -> Array<String> {
+        var values = Array<String>()
```
ryancarlson:
````swift
var values = [String]()
````
also works :)

File: godtools/ViewControllers/MenuViewController.swift [hunk 6]
```diff
@@ -0,0 +1,136 @@
...
+        if section == 0 {
+            values = self.general
+        }
+        else if section == 1 {
```
ryancarlson: Please move these else's up to the previous lines.

godtools/ViewControllers/MenuViewController.swift
```diff
@@ -0,0 +1,136 @@
...
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let values = self.getSectionData(indexPath.section)
+        let value = values[indexPath.row]
+        let cell = tableView.dequeueReusableCell(withIdentifier: self.menuCellIdentifier) as! MenuTableViewCell
+        cell.value = value
+        
+        if value == "notifications" || value == "preview_mode_translators_only" {
+            cell.isSwitchCell = true
+        }
+        else {
```
ryancarlson: else to the previous line here.

godtools/ViewControllers/MenuViewController.swift
pablomarti: I need to set that empty button in order to clear the baritems (removing the common back button that appears). Anything that I did for cleaning it didn't work (really strange), so I came to that "solution"... What do you think?

pablomarti:
@ryancarlson I just uploaded the changes requested

godtools/ViewControllers/MenuViewController.swift
ryancarlson: This isn't a big deal, so we don't need to spend a bunch of time on it. I was mainly curious.
One last thing, did you try `self.navigationItem.hidesBackButton = true` ? If you did and that didn't work, let's keep this since it works and move on :)

pablomarti: Yeah, the thing is that, in some cases I want to display the back button, but without text (I set something for remove the text) but that is not working on the menu view controller (maybe for the way that I called it, in order to set the animation), so putting all things together, I found easier and cleaner to do this on this way...

ryancarlson: Okay, let's keep it this way then.

PR #11, opened 2017-04-26
Add style to languages tableview

pablomarti:
This PR adds the style to the languages tableview screen and its cells.

### DIFF
--- a/godtools/ViewControllers/Platform/LanguagesTableViewController.swift
+++ b/godtools/ViewControllers/Platform/LanguagesTableViewController.swift
@@ -17,6 +17,13 @@ class LanguagesTableViewController: BaseViewController {
     
     let languagesManager = LanguagesManager.shared
     
+    var screenTitleAux: String?
+    override var screenTitle: String {
+        get {
+            return screenTitleAux!.localized
+        }
+    }
+    
     @IBOutlet weak var tableView: UITableView! {
         didSet {
             tableView.delegate = languagesManager
@@ -56,6 +63,15 @@ class LanguagesTableViewController: BaseViewController {
     }
     
     // MARK: - Helpers
+    
+    func primaryLanguage(primary:Bool) {
+        if primary {
+            self.screenTitleAux = "primary_language"
+        }
+        else {
+            self.screenTitleAux = "parallel_language"
+        }
+    }
     
     fileprivate func registerCells() {
         self.tableView.register(UINib(nibName: "LanguageTableViewCell", bundle: nil), forCellReuseIdentifier: LanguagesManager.languageCellIdentifier)
--- a/godtools/ViewControllers/Platform/LanguagesTableViewController.swift
+++ b/godtools/ViewControllers/Platform/LanguagesTableViewController.swift
@@ -58,8 +58,7 @@ class LanguagesTableViewController: BaseViewController {
     // MARK: - Helpers
     
     fileprivate func registerCells() {
-        self.tableView .register(
-            LanguageTableViewCell.classForKeyedArchiver(), forCellReuseIdentifier: LanguagesManager.languageCellIdentifier)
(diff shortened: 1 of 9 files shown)


godtools/Views/Cells/LanguageTableViewCell.swift
```diff
@@ -9,13 +9,45 @@
...
+    
+    // MARK: - Helpers
+    
+    fileprivate func setupStyle() {
+        let selectedView = UIView()
+        selectedView.backgroundColor = UIColor.gtGreen.withAlphaComponent(0.25)
+        self.selectedBackgroundView = selectedView
+    }
+    
+    func languageExist(exist:Bool) {
```
ryancarlson: if you declared the function like this:
````swift
func languageExists(_ exist:Bool) {
````
entonces el llamador no tiene que escribir "exist:" porque "_" esta aqui.

````swift
cell.languageExists(exist: true)
//becomes
cell.languageExists(true)
````

File: godtools/Views/Cells/LanguageTableViewCell.swift [hunk 2]
```diff
@@ -9,13 +9,45 @@
...
+        if exist {
+            self.deleteButton.isHidden = false
+            self.downloadButton.isHidden = true
+        }
+        else {
```
ryancarlson: can you put the else on the same line as the closing } please?

ryancarlson: you could also simplify this if/else block to:
````swift
self.deleteButton.isHidden = !exist
self. downloadButton.isHidden = exist
````

File: godtools/ViewControllers/Platform/LanguagesTableViewController.swift [hunk 3]
```diff
@@ -17,6 +17,13 @@ class LanguagesTableViewController: BaseViewController {
     
     let languagesManager = LanguagesManager.shared
     
+    var screenTitleAux: String?
+    override var screenTitle: String {
+        get {
+            return screenTitleAux!.localized
```
ryancarlson: is `screenTitleAux` guaranteed to be present? if not, this could crash un wrapping an optional that is nil.

godtools/ViewControllers/Platform/LanguagesTableViewController.swift
```diff
@@ -57,9 +64,17 @@ class LanguagesTableViewController: BaseViewController {
     
     // MARK: - Helpers
     
+    func primaryLanguage(primary:Bool) {
```
ryancarlson: A couple things here:
1. Could you change the method name to `selectingPrimaryLanguage(..)`. It's current name makes me think that I would be passing a Language in, not a bool.
2. Using the underscore in the method signature (before "primary") prevents the caller from having to type "primary"

ryancarlson:
@pablomarti oops, I meant to pick "request changes". Please see my comments before merging.

pablomarti:
Thanks for checking @ryancarlson ! I made the corrections

godtools/Views/Cells/LanguageTableViewCell.swift
```diff
@@ -39,15 +39,9 @@ class LanguageTableViewCell: UITableViewCell {
...
-    func languageExist(exist:Bool) {
-        if exist {
-            self.deleteButton.isHidden = false
-            self.downloadButton.isHidden = true
-        }
-        else {
-            self.deleteButton.isHidden = true
-            self.downloadButton.isHidden = false
-        }
+    func languageExist(_ exist:Bool) {
```
ryancarlson: This is very small, but could you change "Exist" to "Exists" and "exist" to "exists" please? It will just read a little bit better. Then this is 👍

ryancarlson:
Just one more muy peuqueña cosa, then this is 👍

pablomarti:
Ok @ryancarlson  done! :D

ryancarlson:
Gracias! this is good! 👍 :shipit:

PR #12, opened 2017-04-28
Build the tool for showing the content from the XML files

pablomarti:
This PR aims to generate the elements that are going to be used across the tools. The elements generated here are going to be used for creating the tool of the KGP.


godtools/ViewControllers/Tract/TractViewController.swift
```diff
@@ -38,6 +53,46 @@ class TractViewController: BaseViewController {
         let activityController = UIActivityViewController(activityItems: [String.localizedStringWithFormat("tract_share_message".localized, "www.knowgod.com")], applicationActivities: nil)
         present(activityController, animated: true, completion: nil)
     }
+    
+    func getData() -> Dictionary<String, Any> {
+        let paragraphContent = [
```
pablomarti: This is some dummy data, but, when parsing the XML we should get this structure.

godtools/ViewControllers/Tract/TractViewController.swift
```diff
@@ -8,13 +8,45 @@
...
+    
+    @objc fileprivate func handleGesture(gesture: UISwipeGestureRecognizer) {
+        if gesture.direction == .right {
+            moveToPreviousPage()
+        } else if gesture.direction == .left {
+            moveToNextPage()
+        }
+    }
+    
+    func moveToNextPage() {
```
ryancarlson: Are we going to push/pop view controllers? I don't think this will work because we will need to be able to jump between views based on button presses later on. I was thinking we would do one view controller with a collection of views, or something like that.

pablomarti: I got confused, just in case, when moving to the next page is like moving from the 1st spiritual law to the 2nd, and from the 2nd to the 3rd, and each viewcontroller (TractViewController) is going  to have the collection of the views (for the tabs). Is that what you were talking about? or am I missing something?

ryancarlson: So I was thinking there would be one view controller that would hold a collection of views. Swiping to the left or right would replace the current view with the next or previous. It would be similar, I think to how you build then onboarding flow.

godtools/Views/TractElements/Card.swift
```diff
@@ -0,0 +1,27 @@
...
+
+import Foundation
+import UIKit
+
+class Card: BaseTractElement {
+    
+    let paddingConstant = CGFloat(30.0)
+    
+    override func setupView(properties: Dictionary<String, Any>) {
+        let frame = CGRect(x: 0.0, y: self.yStartPosition  + paddingConstant, width: BaseTractElement.Standards.screenWidth, height: self.height)
```
ryancarlson: Can you format this line like this:
````swift
CGRect(x: 0.0, 
       y: self.yStartPosition  + paddingConstant, 
       width: BaseTractElement.Standards.screenWidth, 
       height: self.height)
````

It makes it easier to review.

File: godtools/Views/TractElements/TractRoot.swift [hunk 4]
```diff
@@ -0,0 +1,21 @@
...
+
+import Foundation
+import UIKit
+
+class TractRoot: BaseTractElement {
+    
+    override func setupView(properties: Dictionary<String, Any>) {
+        let frame = CGRect(x: 0.0, y: self.yStartPosition, width: BaseTractElement.Standards.screenWidth, height: self.height)
+        let view = UIView(frame: frame)
+        view.backgroundColor = UIColor.yellow
```
ryancarlson: `.yellow` :)

godtools/Views/TractElements/TextContent.swift
```diff
@@ -0,0 +1,80 @@
...
+        let text: String = attributes.text
+        let backgroundColor: UIColor = attributes.backgroundColor
+        
+        label.text = text
+        label.numberOfLines = 0
+        label.lineBreakMode = .byWordWrapping
+        label.sizeToFit()
+        label.backgroundColor = backgroundColor
+        
+        if BaseTractElement.isParagraphElement(self) {
```
ryancarlson: most of lines here are the same, except for the style. Can you just set the style in the if/else block, and write the other lines just once.

godtools/Views/BaseTractView.swift
```diff
@@ -0,0 +1,28 @@
...
+
+import Foundation
+import UIKit
+import SWXMLHash
+
+class BaseTractView: UIView {
+    
+    var data: XMLIndexer?
+    
+    override func awakeFromNib() {
```
ryancarlson: I don't think we need the override method if we're not doing anything in it. It helps keep the codebase cleaner.

godtools/ViewControllers/Tract/TractViewController.swift
pablomarti: Oh ok understood, cool I can make it like that

godtools/Views/TractElements/TextContent.swift
pablomarti: Sure! Actually I will make some refactors

pablomarti:
Hey @ryancarlson this PR is still as WIP, it has a lot of changes, and it is becoming too long, but I think it is impossible to avoid because this is the core for generating the views. Would you please take a look to see how is going? Currently what is pending is:
- Add and remove the views, for when the user is advancing, it should remove the previous views and load the new ones; I did this but it isn't working well...
- Refactor and clean the code (just to follow the guidelines and make it easy to read)
- Remove the current styles (the backgrounds that were added for testing)
- Add the sketch file styles (backgrounds, borders)

pablomarti:
Hey @ryancarlson would you please check this PR? It has a bug with an animation when going back from the TractViewController to Home, I'm already solving it, but, I'm doing that in another branch so this PR can be checked and merged. Let me know what do you think, sorry for making it too big, didn't find another way. Thanks Ryan!

godtools/ViewControllers/Tract/TractViewController.swift
```diff
@@ -8,9 +8,38 @@
 
 import Foundation
 import UIKit
+import SWXMLHash
 
 class TractViewController: BaseViewController {
     
+    var viewsWereGenerated :Bool = false
```
ryancarlson: You don't need to declare `:Bool` since the variable is initialized to "false".

godtools/ViewControllers/Tract/TractViewController.swift
```diff
@@ -30,6 +67,138 @@ class TractViewController: BaseViewController {
         }
     }
     
+    fileprivate func initializeView() {
+        let navigationBarFrame = navigationController!.navigationBar.frame
+        let startingPoint = navigationBarFrame.origin.y + navigationBarFrame.size.height
```
ryancarlson: This variable name is a little confusing because it has "Point" in the name, but it's not a `CGPoint`. Could we name it "startingYPos" so it's clear later that we are dealing with the y position only as a `CGFloat`?

godtools/ViewControllers/Tract/TractViewController.swift
```diff
@@ -30,6 +67,138 @@ class TractViewController: BaseViewController {
...
+        
+        let width = self.view.frame.size.width
+        let height = self.view.frame.size.height - startingPoint
+        self.containerView.frame = CGRect(x: 0.0, y: startingPoint, width: width, height: height)
+        self.view.addSubview(self.containerView)
+        
+        buildPages(width, height)
+    }
+    
+    fileprivate func buildPages(_ width: CGFloat, _ height: CGFloat) {
```
ryancarlson: I don't think width and height should be suppressed here by adding the `_` because it is not implicit from the function name what parameters the caller should pass in. Can you remove the underscores and change it so that the caller specifies `width:` and `height:` . 

I know it will be redundant, but I think it would prevent confusion later if variable names change.

godtools/Views/TractElements/TractRoot.swift
```diff
@@ -0,0 +1,31 @@
...
+import Foundation
+import UIKit
+import SWXMLHash
+
+//  NOTES ABOUT THE COMPONENT
+//  * This component must always be initialized with: init(data: XMLIndexer, withMaxHeight height: CGFloat)
+
+class TractRoot: BaseTractElement {
+    
+    override func setupView(properties: Dictionary<String, Any>) {
```
ryancarlson: `Dictionary<String, Any>` can be declared as: `[Dictionary: Any]`

reference: https://github.com/raywenderlich/swift-style-guide#syntactic-sugar

godtools/Views/TractElements/BaseTractElement.swift
```diff
@@ -0,0 +1,175 @@
...
+//  Copyright © 2017 Cru. All rights reserved.
+//
+
+import Foundation
+import UIKit
+import SWXMLHash
+
+class BaseTractElement: UIView {
+    struct Standards {
+        static let xMargin = CGFloat(8.0)
```
ryancarlson: Can you define these the same way you did further below for consistency?
````swift
static let xMargin: CGFloat = 8.0
````

File: godtools/Views/TractElements/BaseTractElement.swift [hunk 12]
```diff
@@ -0,0 +1,175 @@
...
+        static let screenWidth = UIScreen.main.bounds.size.width
+        static let textContentWidth = UIScreen.main.bounds.size.width - xMargin * CGFloat(2)
+    }
+    
+    weak var parent: BaseTractElement?
+    var elements:[BaseTractElement]?
+    var yStartPosition: CGFloat = 0.0
+    var maxHeight: CGFloat = 0.0
+    private var _height: CGFloat = 0.0
+    var height: CGFloat {
```
ryancarlson: Are the getter and setter here still necessary? I don't see any special logic happening

godtools/Views/TractElements/BaseTractElement.swift
```diff
@@ -0,0 +1,175 @@
...
+        get {
+            return _height
+        }
+        set {
+            _height = newValue
+        }
+    }
+    var width: CGFloat {
+        if (self.parent != nil) {
+            return (self.parent?.width)!
```
ryancarlson: If you force unwrap parent, then you don't have to unwrap the whole statement.
````swift
return self.parent!.width
````

File: godtools/Views/TractElements/BaseTractElement.swift [hunk 14]
```diff
@@ -0,0 +1,175 @@
...
+        } else {
+            return BaseTractElement.Standards.screenWidth
+        }
+    }
+    var horizontalContainer: Bool {
```
ryancarlson: Did you want to just initialize this to false? Or why did you put the return below?

godtools/Views/TractElements/BaseTractElement.swift
```diff
@@ -0,0 +1,175 @@
...
+    
+    // MARK: - Initializers
+    
+    init(children: [XMLIndexer], startOnY yPosition: CGFloat, parent: BaseTractElement) {
+        let frame = CGRect(x: 0.0, y: 0.0, width: 0.0, height: 0.0)
+        super.init(frame: frame)
+        self.parent = parent
+        self.yStartPosition = yPosition
+        buildChildrenForData(children)
+        setupView(properties: Dictionary<String, Any>())
```
ryancarlson: Let's use `[String: Any]()` here

godtools/Views/TractElements/BaseTractElement.swift
```diff
@@ -0,0 +1,175 @@
...
+    }
+    
+    func setupView(properties: Dictionary<String, Any>) {
+        preconditionFailure("This function must be overridden")
+    }
+    
+    func buildChildrenForData(_ data: [XMLIndexer]) {
+        var currentYPosition: CGFloat = 0.0
+        var maxYPosition: CGFloat = 0.0
+        var elements:Array = [BaseTractElement]()
```
ryancarlson: No need to specify `Array` here since you initialize the variable as an array.

godtools/Views/TractElements/TractElementOperationFunctions.swift
```diff
@@ -0,0 +1,68 @@
...
+import Foundation
+import SWXMLHash
+
+extension BaseTractElement {
+    
+    func buildElementForDictionary(_ data: XMLIndexer, startOnY yPosition: CGFloat) -> BaseTractElement {
+        let dataContent = splitData(data: data)
+        var element:BaseTractElement?
+        
+        if dataContent.kind == "hero" {
```
ryancarlson: In swift we can switch a string now, which may be a little more readable? What do you think?

godtools/Views/TractElements/TractElementOperationFunctions.swift
```diff
@@ -0,0 +1,68 @@
...
+        } else {
+            element = TextContent(data: data, startOnY: yPosition, parent: self)
+        }
+        
+        return element!
+    }
+    
+    func splitData(data: XMLIndexer) -> (kind: String, properties: Dictionary<String, Any>, children: [XMLIndexer]) {
+        let kind = data.element?.name
+        var properties = Dictionary<String, Any>()
```
ryancarlson: `[String: Any]()` again here.

godtools/Views/TractElements/CallToAction.swift
```diff
@@ -0,0 +1,34 @@
...
+//  Created by Devserker on 4/28/17.
+//  Copyright © 2017 Cru. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+class CallToAction: BaseTractElement {
+    
+    let paddingConstant = CGFloat(30.0)
```
ryancarlson: You can specify the type `CGFloat` and then just use = 30.0 again here.

godtools/Views/TractElements/Cards.swift
```diff
@@ -0,0 +1,84 @@
...
+
+class Cards: BaseTractElement {
+    
+    enum CardsState {
+        case open, preview
+    }
+    
+    var cardsState = CardsState.preview
+    
+    var xPosition: CGFloat {
```
ryancarlson: Here's another example where 0.0 is returned from the getter. Can this be changed to just initialize it to 0.0, or can it be a let xPosition = 0.0 if it doesn't change?

godtools/Views/TractElements/Cards.swift
```diff
@@ -0,0 +1,84 @@
...
+        return self.height - self.yStartPosition
+    }
+    
+    override func buildChildrenForData(_ data: [XMLIndexer]) {
+        var elements:Array = [BaseTractElement]()
+        var cardNumber = 0
+        
+        for dictionary in data {
+            let deltaChange = CGFloat(data.count - cardNumber)
+            let yPosition = self.initialCardPosition - (deltaChange * 60)
```
ryancarlson: Can you extract 60 here and 30 below out to constants within the class? It will be nice to change them in one place and having a variable name will increase readability.

godtools/Views/TractElements/Header.swift
```diff
@@ -0,0 +1,51 @@
...
+        set {
+            super.height = newValue
+        }
+    }
+    
+    override var horizontalContainer: Bool {
+        return true
+    }
+    
+    override func setupView(properties: Dictionary<String, Any>) {
```
ryancarlson: `[String: Any]`

godtools/Views/TractElements/TextContent.swift
```diff
@@ -0,0 +1,102 @@
...
+        
+        if properties["value"] != nil {
+            text = properties["value"] as! String
+        }
+        
+        if properties["backgroundColor"] != nil {
+            let backgroundColorText = properties["backgroundColor"] as! String
+            switch backgroundColorText {
+            case "red":
+                backgroundColor = UIColor.gtRed
```
ryancarlson: you can just use `.gtRed` here and on the next few lines.

godtools/Views/TractElements/TractLabel.swift
```diff
@@ -0,0 +1,84 @@
...
+    
+    override func setupView(properties: Dictionary<String, Any>) {
+        self.frame = buildFrame()
+    }
+    
+    func buildHorizontalLine() {
+        let xPosition = Card.xPaddingConstant
+        let yPosition = self.frame.size.height - 1
+        let width = self.frame.size.width - (Card.xPaddingConstant * CGFloat(2))
+        let height = CGFloat(1.0)
```
ryancarlson:
````swift
let height: CGFloat = 1.0
````

File: godtools/Views/TractElements/TractLabel.swift [hunk 25]
```diff
@@ -0,0 +1,84 @@
...
+        
+        let horizontalLine = UIView()
+        horizontalLine.frame = CGRect(x: xPosition,
+                                      y: yPosition,
+                                      width: width,
+                                      height: height)
+        horizontalLine.backgroundColor = UIColor.gtGreyLight
```
ryancarlson: `gtGreyLight`

godtools/Views/TractElements/BaseTractElement.swift
pablomarti: Most of the times will be false, but I need for some view (and maybe another views) in the future to be true, so in the children class I override this property and thats why I need a return, also, because it is easier to do that than setting `self.horizontalContainer = true` on the children class, because in order to do that it will require to override another method (because of the moment when the property is being used)

godtools/ViewControllers/Tract/TractViewController.swift
```diff
@@ -25,7 +25,7 @@ class TractViewController: BaseViewController {
         super.viewDidLoad()
         
         getData()
-        displayTitle()
+        displayScreenTitle()
```
ryancarlson: You might not need to call this any more now that we are just overriding a superclass method to set the title element. This is already called in BaseViewController.viewWillAppear().

I originally took a more manual approach to setting the custom title on this view, but then refactored it to use the pattern you created in the view controller. I'm pretty sure this line can be removed.

pablomarti:
Thank you @ryancarlson !

PR #13, opened 2017-04-28
Wire UI to data/services (part 1)

ryancarlson:
Begins to wire some of the front end UIs into the background data model.

- Home and Add Tools show available resources, based off of DownloadedResource.shouldDownload
- Selecting an item from Add Tools "downloads" (simulated for now) and makes it appear on the Home view
   - this should be the button driving the "download", I think, but it's done there yet
- Selecting a language from the languages screen sets it as the primary/parallel language in the user settings
  - download/delete buttons are not yet wired up.
- Onboarding view is shown in viewDidLoad so it only shows once per app open, and once a user selects a primary language it is not shown again.

### DIFF
--- a/godtools/Managers/DownloadResources/DownloadedResourceManager.swift
+++ b/godtools/Managers/DownloadResources/DownloadedResourceManager.swift
@@ -25,6 +25,7 @@ class DownloadedResourceManager: GTDataManager {
         serializer.registerResource(DownloadedResourceJson.self)
         serializer.registerResource(TranslationResource.self)
         serializer.registerResource(PageResource.self)
+        serializer.registerResource(LanguageResource.self)
     }
     
     func loadFromDisk() -> Promise<[DownloadedResource]> {
@@ -57,12 +58,13 @@ class DownloadedResourceManager: GTDataManager {
                 cachedResource.code = remoteResource.abbreviation
                 cachedResource.name = remoteResource.name
                 
-                let remoteTranslations = remoteResource.translations!
+                let remoteTranslations = remoteResource.latestTranslations!
                 for remoteTranslationGeneric in remoteTranslations {
                     let remoteTranslation = remoteTranslationGeneric as! TranslationResource
                     
                     let cachedTranslation = Translation.mr_findFirstOrCreate(byAttribute: "remoteId", withValue: remoteTranslation.id!, in: context)
                     
+                    cachedTranslation.language = Language.mr_findFirst(byAttribute: "remoteId", withValue: remoteTranslation.language?.id ?? "-1", in: context)
                     cachedTranslation.version = remoteTranslation.version!.int16Value
                     cachedTranslation.isPublished = remoteTranslation.isPublished!.boolValue
                     cachedResource.addToTranslations(cachedTranslation)
--- a/godtools/Managers/DownloadResources/DownloadedResourceManager.swift
+++ b/godtools/Managers/DownloadResources/DownloadedResourceManager.swift
@@ -28,9 +28,9 @@ class DownloadedResourceManager: GTDataManager {
         serializer.registerResource(LanguageResource.self)
     }
     
-    func loadFromDisk() -> Promise<[DownloadedResource]> {
+    func loadFromDisk() -> [DownloadedResource] {
         resources = DownloadedResource.mr_findAll() as! [DownloadedResource]
-        return Promise(value: resources)
+        return resources
     }
     
     func loadFromRemote() -> Promise<[DownloadedResource]> {
@@ -46,7 +46,7 @@ class DownloadedResourceManager: GTDataManager {
                 } catch {
                     return Promise(error: error)
                 }
-                return self.loadFromDisk()
+                return Promise(value:self.loadFromDisk())
             }
     }
     
--- a/godtools/Managers/DownloadResources/DownloadedResourceManager.swift
+++ b/godtools/Managers/DownloadResources/DownloadedResourceManager.swift
@@ -63,8 +63,9 @@ class DownloadedResourceManager: GTDataManager {
                     let remoteTranslation = remoteTranslationGeneric as! TranslationResource
                     
                     let cachedTranslation = Translation.mr_findFirstOrCreate(byAttribute: "remoteId", withValue: remoteTranslation.id!, in: context)
-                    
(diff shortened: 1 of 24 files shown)


ryancarlson:
@pablomarti this is still a WIP too, but works to take the views you created and have them driven by data pulled from the API.

pablomarti:
Great @ryancarlson ! I will be checking this

godtools/Managers/DownloadResources/DownloadedResourceManager.swift
```diff
@@ -57,12 +58,13 @@ class DownloadedResourceManager: GTDataManager {
...
                 cachedResource.name = remoteResource.name
                 
-                let remoteTranslations = remoteResource.translations!
+                let remoteTranslations = remoteResource.latestTranslations!
                 for remoteTranslationGeneric in remoteTranslations {
                     let remoteTranslation = remoteTranslationGeneric as! TranslationResource
                     
                     let cachedTranslation = Translation.mr_findFirstOrCreate(byAttribute: "remoteId", withValue: remoteTranslation.id!, in: context)
                     
+                    cachedTranslation.language = Language.mr_findFirst(byAttribute: "remoteId", withValue: remoteTranslation.language?.id ?? "-1", in: context)
```
pablomarti: Would you please set the value to be searched on a variable and here just use the variable?

pablomarti: Also, would you please make a method for searching instead of calling to MR methods directly?

godtools/Managers/LanguagesManager.swift
```diff
@@ -19,13 +19,29 @@ class LanguagesManager: GTDataManager {
...
+    var selectingPrimaryLanguage = true
     
     override init() {
         super.init()
         serializer.registerResource(LanguageResource.self)
     }
     
-    func loadFromDisk() -> Promise<[Language]> {
+    func loadFromDisk(id: String) -> Language {
+        let language = Language.mr_findFirst(byAttribute: "remoteId", withValue: id)!
```
pablomarti: As I said before, here you can call a method to search by remoteId instead of calling the method of MR, but the difference is that here you don't set the context, and previously you do, is there a reason?

godtools/Managers/ToolsManager.swift
```diff
@@ -54,11 +63,13 @@ extension ToolsManager: UITableViewDataSource {
     
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
         let cell = tableView.dequeueReusableCell(withIdentifier: ToolsManager.toolCellIdentifier) as! HomeToolTableViewCell
+        cell.titleLabel.text = self.resources![indexPath.section].name
```
pablomarti: What do you think about making functions to pass the data to the cellview instead of setting the values here? it may be much cleaner I believe

godtools/ViewControllers/Platform/LanguageSettingsViewController.swift
```diff
@@ -15,30 +15,51 @@ protocol LanguageSettingsViewControllerDelegate {
...
         
-        _ = DownloadedResourceManager.shared.loadFromRemote()
+        setupPrimaryLanguageButton()
+        setupParallelLanguageButton()
+    }
+    
+    private func setupPrimaryLanguageButton() {
+        let primaryLanguage = self.languagesManager.loadPrimaryLanguageFromDisk()
+        if primaryLanguage != nil {
+            primaryLanguageButton.setTitle(primaryLanguage!.localizedName, for: .normal)
```
pablomarti: This is just coding style, but what do you think about storing the language name into a variable, so you would call the `setTitle` only once for the variable?

godtools/ViewControllers/Platform/LanguageSettingsViewController.swift
```diff
@@ -15,30 +15,51 @@ protocol LanguageSettingsViewControllerDelegate {
...
+            primaryLanguageButton.setTitle("select_primary_language".localized, for: .normal)
+        }
     }
 
-    override func didReceiveMemoryWarning() {
-        super.didReceiveMemoryWarning()
+    private func setupParallelLanguageButton() {
+        if GTSettings.shared.parallelLanguageId != nil {
+            let parallelLanguage = self.languagesManager.loadFromDisk(id: GTSettings.shared.parallelLanguageId!)
+            parallelLanguageButton.setTitle(parallelLanguage.localizedName, for: .normal)
```
pablomarti: Same here about setting the values into a variable

godtools/Managers/LanguagesManager.swift
ryancarlson: The reason I created this method is to isolate the calls MR calls. Basically I wanted to keep the MR calls for Language in the the LangaugesManager. But I missed one that you previous called out in your other comment.

I want to take another look at context. If we don't specify the context, then it should be the "default" one. In some cases I had to specify it though because I ran into errors about relating objects from two different contexts together. I will take another look and see if that can be standardized at all.

godtools/Managers/ToolsManager.swift
ryancarlson: Yes, I think that's a good idea.

godtools/ViewControllers/Platform/LanguageSettingsViewController.swift
ryancarlson: 👍

godtools/Managers/DownloadResources/DownloadedResourceManager.swift
ryancarlson: 👍

ryancarlson:
@pablomarti I pushed updates for most of your suggestions here.

I don't yet see any clear abstraction patterns that would provide much benefit yet to the object manager classes. Do you have ideas to share?

I like calling the "mr" methods on the language classes right now, especially for simple lookups where we just ask for something by one attribute. We could abstract something out, but that would either get into passing class references through things, or creating predicates where we don't need to. The "mr" methods are simpler to me.

godtools/Managers/LanguagesManager.swift
pablomarti: Ok cool

pablomarti:
Ok @ryancarlson I get it! yeah the best is to go like this, it is clear, no need to change, maybe in the future with the following models we may have things clear for including some patterns for abstraction. Feel free to merge!

PR #14, opened 2017-05-05
Finishes wiring home/languages/tool details screen to app data state

ryancarlson:


ryancarlson:
@pablomarti este PR está listo por revisar, por favor.

ryancarlson:
Thanks @pablomarti ! I'm open to moving those files into separate files in the future.

PR #15, opened 2017-05-09
Add custom styles and new components to Tract Elements

pablomarti:
- Read XML files
- Change color of texts depending on configurations of XMLs
- Set background images depending on configurations of XMLs
- Change color of navigation bar depending on configuration of XMLs
- Display progress bar of tract on the top of navigation bar
- Add transparent layer effect at the bottom of the scrollview of each card
- Use new XML element <cards>
- Display "Call to action" and add animation
- Fix transition from Tract to Home


pablomarti:
Hey @ryancarlson this is a WIP PR that should be merged to the PR #12, it is almost done, just have to work on some things related to the Call to Action, and make a new refactoring (non-destructive, just for adding order). Wanted to show you how things are going!

pablomarti:
Hey @ryancarlson this PR is finish, would you please review it? thanks!

godtools/Views/TractElements/CardsAnimations.swift
```diff
@@ -57,6 +63,16 @@ extension Cards {
         }
     }
     
+    func showCallToAction() {
+        let rootView = self.parent as! TractRoot
```
ryancarlson: Is the parent guaranteed to be a `TractRoot` here? Can you confirm please?

godtools/Views/TractElements/Paragraph.swift
```diff
@@ -27,18 +27,21 @@ class Paragraph: BaseTractElement {
...
         
-        return ("toolFrontSubTitle",
-                self.width,
-                0.0,
-                NSTextAlignment.left,
-                xMargin,
-                BaseTractElement.yMargin)
+        textStyle.style = "toolFrontSubTitle"
+        textStyle.width = self.width
+        textStyle.alignment = .left
```
ryancarlson: How will the operating system handle this for right to left languages (Arabic, Hebrew and others)? We want to make sure we support those well for this new version of God Tools.

godtools/Views/TractElements/Card.swift
```diff
@@ -76,6 +78,24 @@ class Card: BaseTractElement {
                                           y: 0.0,
                                           width: self.width,
                                           height: contentHeight)
+        self.containerView.backgroundColor = .clear
+    }
+    
+    func setupTransparentView() {
+        let width = self.scrollView.frame.size.width - 6.0
+        let height = CGFloat(60.0)
```
ryancarlson: Initialization here and other places in this file.. :CGFloat

godtools/Views/TractElements/BaseTractElement.swift
```diff
@@ -20,6 +21,25 @@ class BaseTractElement: UIView {
     
     weak var parent: BaseTractElement?
     var elements:[BaseTractElement]?
+    var didFoundCallToAction: Bool = false
```
pablomarti: Thank! I missed that!

godtools/Views/TractElements/Cards.swift
```diff
@@ -66,6 +66,11 @@ class Cards: BaseTractElement {
     override func setupView(properties: Dictionary<String, Any>) {
         self.frame = buildFrame()
         self.backgroundColor = .clear
+        
+        let image = UIImage(named: self.backgroundImagePath)
```
pablomarti: Yes we have to, I didn't do that because the options were missing on the element (is still as WIP), but will make something and add suggestions to the XML file

godtools/Views/TractElements/CardsAnimations.swift
pablomarti: Yes, <cards> should always be a child of <page> (which is root)

godtools/Views/TractElements/Paragraph.swift
pablomarti: Didn't think about this :) will be working on that

pablomarti:
Hey @ryancarlson what do you think about these changes? I will continue working on the images on the PR #17 and I will wait with the swipe direction. Is this good to go?

ryancarlson:
👍

pablomarti:
Thanks @ryancarlson !

PR #17, opened 2017-05-11
Load new elements into Tract

pablomarti:
- Fix problem of "Call to action" that wasn't disappearing 
- Fix change of current state of cards
- New XML file with image included
- Add component for loading images into cards
- Refactoring for have a single way for management the properties of <text> component
- Add component for buttons and links
- Fix position of background images on cards
- Add component for tabs on cards
- Allow to align images depending on configuration


pablomarti:
Hey @ryancarlson in this PR I worked with the background images, for the scaling. There is a situation, as you can see on this commit: https://github.com/CruGlobal/godtools-swift/pull/17/commits/4221099ffb126e32f0da3c171ba5d73641ccbe32 it will try to adjust the image using **scaleAspectFit**, and it works, but on some cases because of the ratio the width of the image may be shorter that the width of the view, so, in order to solve that problem I did this https://github.com/CruGlobal/godtools-swift/pull/17/commits/6571087ae223198d31edf94b76af97262c8e57c5 which basically only deals with the width, but the consequence of this 2nd case is that part of the image may get lost at the bottom. I believe the best in order to avoid this problems is to deal with a specific ratio in all images (currently all the images have different ratios). What do you think?

pablomarti:
This PR is done, would you please check it @ryancarlson ?

pablomarti:
Also @ryancarlson I have to say that, currently with the tabs the images are not being selected, if you run the app and try that part, it may seems that is selected, but no, I still have to work on that, and in order to do that I have to think how to update the properties of the XML objects in order to accomplish that.

ryancarlson:
@pablomarti Este PR está listo por revisar? Lo olvidé.

pablomarti:
Yes this one @ryancarlson will fix the conflict

pablomarti:
Ready @ryancarlson now is possible to check.

godtools/Views/Properties/TractButtonProperties.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+
+import UIKit
+
+class TractButtonProperties {
+    
+    var i18nId: String?
+    var type: String?
+    var value: String?
+    var width = CGFloat(300.0)
```
ryancarlson: Initialization style again here :)

ryancarlson: And other places in this file

godtools/Views/TractElements/BaseTractElement.swift
```diff
@@ -143,16 +143,29 @@ class BaseTractElement: UIView {
...
+        var previousPosition: CGFloat = 0.0
         
         for dictionary in data {
-            let element = buildElementForDictionary(dictionary, startOnY: currentYPosition)
-            if self.horizontalContainer && element.yEndPosition() > maxYPosition {
-                maxYPosition = element.yEndPosition()
+            var element: BaseTractElement?
+            let properties = splitData(data: dictionary)
+            
+            if previousElementWasImage && properties.kind == "image" {
```
ryancarlson: Can you explain a little bit what you're doing in this section? I am not following it yet.

godtools/Views/TractElements/Cards.swift
```diff
@@ -65,10 +65,32 @@ class Cards: BaseTractElement {
...
+    
+    func setupBackground() {
         self.backgroundColor = .clear
         
         let image = UIImage(named: self.backgroundImagePath)
         let imageView = UIImageView(image: image)
+        
+        let viewWidth = self.frame.size.width
+        let viewHeight = self.frame.size.height
+        var width = image?.size.width
```
ryancarlson: I would just unwrap image here. It will save having to unwrap the width and height everywhere you use it.

You could and should also do a guard let when loading the image to prevent crashes. So above you could do:
````swift
guard let image = UIImage(name: self.backgroundImagePath) else {
 return
}
````
This way the worst thing that can happen is there is no background if the image is missing. As it is now, the app will crash.

File: godtools/Views/TractElements/Cards.swift [hunk 4]
```diff
@@ -65,10 +65,32 @@ class Cards: BaseTractElement {
...
+        var height = image?.size.height
+        let ratio = width! / height!
+        
+        if height! > viewHeight || width! > viewWidth {
+            width = viewWidth
+            height = width! / ratio
+        }
+        
+        let xPosition = (viewWidth - width!) / CGFloat(2.0)
+        let yPosition = CGFloat(0.0)
```
ryancarlson: `let yPosition: CGFloat = 0.0` :)

godtools/Views/TractElements/TextContent.swift
```diff
@@ -76,29 +66,35 @@ class TextContent: BaseTractElement {
...
-                backgroundColor = .clear
+        self.xPosition = self.properties.xMargin
+        self.yPosition = self.yStartPosition + self.properties.yMargin
+    }
+    
+    func loadElementProperties(properties: [String: Any]) {
+        for property in properties.keys {
+            switch property {
+            case "value":
+                self.properties.value = properties[property] as! String?
```
ryancarlson: I think you could do this here instead:
````swift
self.properties.value = properties[property] as? String
````
which is a little bit cleaner and easier to read. Very minor thing.

File: godtools/Views/TractElements/TractButton.swift [hunk 6]
```diff
@@ -0,0 +1,121 @@
...
+    override var width: CGFloat {
+        return super.width - self.xPosition - self.xMargin
+    }
+    var buttonWidth: CGFloat {
+        return self.properties.width > self.width ? self.width : self.properties.width
+    }
+    var buttonXPosition: CGFloat {
+        return (self.width - self.buttonWidth) / 2
+    }
+    var textPadding = CGFloat(8.0)
```
ryancarlson: CGFloat

godtools/Views/TractElements/TractImage.swift
```diff
@@ -0,0 +1,81 @@
...
+                      y: self.yPosition,
+                      width: self.width,
+                      height: self.height)
+    }
+    
+    func loadImage(properties: [String: Any]) -> UIImage {
+        let resource = properties["resource"] as! String?
+        self.align = properties["align"] as! String
+        
+        let image = UIImage(named: resource!)!
```
ryancarlson: This would also cause a crash if the image cannot be loaded. Is there any way you can think of that we could fail gracefully if the image is missing?

godtools/Views/TractElements/TractTabs.swift
```diff
@@ -0,0 +1,122 @@
...
+    }
+    
+    override func setupView(properties: [String: Any]) {
+        self.frame = buildFrame()
+        setupSegmentedControl()
+    }
+    
+    func setupSegmentedControl() {
+        let width = self.width
+        let height = CGFloat(28.0)
```
ryancarlson: CGFloat

godtools/Views/TractElements/Header.swift
```diff
@@ -26,6 +26,8 @@ class Header: BaseTractElement {
         }
     }
     
+    var includesNumber = false
```
ryancarlson: Do you have plans for using this in the future?

godtools/Views/TractElements/TractImage.swift
```diff
@@ -0,0 +1,81 @@
...
+    var imageView = UIImageView()
+    
+    override func setupView(properties: [String: Any]) {
+        self.imageView = UIImageView(image: loadImage(properties: properties))
+        var width = self.imageView.frame.size.width
+        var height = self.imageView.frame.size.height
+        var xPosition = CGFloat(0.0)
+        let yPosition = CGFloat(0.0)
+        
+        if width > self.width {
```
ryancarlson: If the image is wider than the view, I agree we should down scale it like you have here. Do we still want to make sure it has a margin on the sides too?

godtools/Views/TractElements/Header.swift
pablomarti: Yes, actually I'm using it on the classes Number and Title (which are children of Header), but, thinking on a better way to handle it... There should be a better way, will think about it (will make an issue), but I believe I should do that in other branch

ryancarlson: Okay cool. That's okay if it's used in a child element. I just missed that during review.

godtools/Views/TractElements/Cards.swift
pablomarti: Thanks! I like it

godtools/Views/TractElements/TractImage.swift
pablomarti: This is difficult in the way that everything depends on the size of the image, if there is no image then the sizes will be 0... Is that good enough? or should I set some default values to show a space where the image should go?

godtools/Views/TractElements/TractImage.swift
pablomarti: For the moment I set xMarginConstant to 0

godtools/Views/TractElements/TractImage.swift
ryancarlson: I think it's good for now. We may need to change it in the future once we start dealing with real images.

ryancarlson:
Looks good @pablomarti ! Thanks for making these changes!

godtools/Views/TractElements/TractImage.swift
pablomarti: Ok

pablomarti:
Thanks @ryancarlson !

PR #19, opened 2017-05-17
Handle downloading of XML zip files from Server

ryancarlson:
Zip files are downloaded and extracted to a common resources folder.

Priority is given to primary and parallel language at the time a resource is downloaded, but is not changed if the user changes languages.

Deleting just flags that translations are not deleted and resources and languages should not be deleted, but does not currently purge resources from disk.

### DIFF
--- a/godtools/Managers/DownloadResources/DownloadedResourceManager.swift
+++ b/godtools/Managers/DownloadResources/DownloadedResourceManager.swift
@@ -116,7 +116,6 @@ class DownloadedResourceManager: GTDataManager {
     
     private func purgeTranslationsOlderThan(_ translation: Translation) {
         let context = NSManagedObjectContext.mr_default()
-        print(translation)
         
         let predicate = NSPredicate(format: "language.remoteId = %@ AND downloadedResource.remoteId = %@ AND version < %d and isDownloaded = false",
                                     translation.language!.remoteId!,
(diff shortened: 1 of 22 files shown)


ryancarlson:
@pablomarti This is ready for review I think. I still need to handle actually deleting files from the disk when doing a delete.

I'm not sure about the last commit from a UX perspective either. It may be confusing for users to have items fade off the list when they are finished downloading.

godtools/Managers/DownloadResources/DownloadedResourceManager.swift
```diff
@@ -87,6 +97,40 @@ class DownloadedResourceManager: GTDataManager {
...
+            return true
+        }
+        
+        let latestTranslation = existingTranslations.max(by: {$0.version < $1.version})
+        
+        return latestTranslation == nil || version > latestTranslation!.version
+    }
+    
+    private func purgeTranslationsOlderThan(_ translation: Translation) {
+        let context = NSManagedObjectContext.mr_default()
```
pablomarti: All of this is fine, just was thinking for the future we can break all of this in some other methods (like getting the context, or making the predicates, or blocks for when doing the operation) so new implementations will be easier

ryancarlson:
Yes, those will be in future commits and PRs, but I didn't want to let this one get too large.  We will see what Ann and Eric say about the fade out. I think it is not such a good idea after all, pero veremos.

godtools/Managers/DownloadResources/DownloadedResourceManager.swift
ryancarlson: Can you show me a sample of what you're thinking?

ryancarlson: Maybe do a commit with one change and we can compare to what is here now?

ryancarlson:
Thanks, will merge! and yes, please show me what you're thinking about in a new commit so we can compare and see how it looks.

PR #20, opened 2017-05-17
Track additional fields from API

ryancarlson:
Copyright info and total views.

Show total views data on home view cell and tool detail view

Also show number of languages based on real data on tool detail view


ryancarlson:
@pablomarti here is another small PR to be reviewed after #19. It is just pulling some more small data pieces in from the API and showing a couple more properties.

PR #21, opened 2017-05-17
Track additional fields from API 

ryancarlson:
Copyright info and total views.

Show total views data on home view cell and tool detail view

Also show number of languages based on real data on tool detail view


pablomarti:
👍

PR #23, opened 2017-05-19
Add modals and movements between cards, modals and other pages

pablomarti:
This PR adds:
- Modal element
- Bindings, so the elements will have a listener when the button dispatch a message to that listener
- Movements between cards, displaying modals, and moving to other pages


pablomarti:
Hey @ryancarlson would you please check this PR? I also added some methods for using the test data instead of using the data from the XML downloaded for the API, just for testing purposes, later we should delete those methods. Also, I made some changes in the XML that I will share to @frett and @Omicron7 to see what they think about it

godtools/XMLContent/KGP/kgp-5.xml
```diff
@@ -100,7 +99,7 @@
                 </content:text>
             </paragraph>
             <paragraph>
-                <content:button type="event" value="kgp-6">
+                <content:button type="event" value="kgp-6,send-information-modal">
```
frett: to me this would retrigger the modal you are trying to close, I don't think you want the "send-information-modal" event triggered here.

pablomarti: Yeah, what I'm doing is to have a boolean which will detect if opening or closing, but, instead of that, what should we do to determine if closing or opening? Maybe sending parameters? like `<content:button type="event" value="kgp-6,send-information-modal:close">`?

frett: just saw your comment from yesterday

frett: I think setting a separate close listener might work better. I'm more concerned about overloading the event and making it mean different things based on current application state.

so, something like: `<modal listeners="send-information-modal" dismiss-listeners="send-information-model-close">`

ryancarlson: Yeah, I would prefer what @frett suggested too. That seems like it will be cleaner in the code.

godtools/ViewControllers/Tract/TractViewController.swift
```diff
@@ -227,6 +231,23 @@ class TractViewController: BaseViewController {
     
     // MARK: - Handle page movements
     
+    func moveToPage(notification: Notification) {
+        guard let dictionary = notification.object as? [String: String] else {
```
ryancarlson: Custom Information should be passed in the `userInfo` property, not in the `object`.

godtools/XMLContent/KGP/kgp-5.xml
frett: Once Goers gets the schema validation PR completed and merged I'll work on updating the XSD for this.

godtools/Views/Buttons/GTButton.swift
```diff
@@ -79,6 +79,17 @@ class GTButton: UIButton {
         self.layoutSubviews()
     }
     
+    func designAsTractModalButton() {
```
ryancarlson: Will tract modal buttons be able to be customized or follow the default colors as defined in the manifest?

godtools/Views/TractElements/TractButton.swift
```diff
@@ -60,16 +64,28 @@ class TractButton: BaseTractElement {
     override func setupView(properties: [String: Any]) {
         loadStyles()
         loadElementProperties(properties: properties)
-        self.height = self.properties.height
         
         self.button = GTButton()
-        button.cornerRadius = self.properties.cornerRadius
-        button.backgroundColor = self.properties.backgroundColor
+        if BaseTractElement.isModalElement(self) {
```
ryancarlson: Could you extract the if and else blocks out into separate functions, named something like `configureAsModalButton` and `configureAsStandardButton` or something like that?

I think that will help the code stay easily readable.

godtools/Configurations/TractBindings.swift
```diff
@@ -0,0 +1,25 @@
...
+
+class TractBindings: NSObject {
+    
+    @nonobjc static var bindings = [String: BaseTractElement]()
+    
+    static func setupBindings() {
+        TractBindings.bindings = [String: BaseTractElement]()
+    }
+    
+    static func addBindings(_ element: BaseTractElement) {
```
ryancarlson: I would name this `addBinding` since we're only receiving one element and adding one binding.

ryancarlson: But I also imagine an element could listen for multiple events. Have you considered that in your design?

godtools/ViewControllers/Tract/TractViewController.swift
```diff
@@ -227,6 +231,23 @@ class TractViewController: BaseViewController {
...
+            return
+        }
+        
+        let destinationViewPageId = dictionary["pageId"]
+        for pageId in self.pagesIds.keys {
```
ryancarlson: I think you could do:
````swift
for (pageId, pageIndex) in self.pageIds {
..
}
````
so that you can already have the value instead of doing `self.pagesIds[pageId]!` . One of the small nice things in Swift!

File: godtools/Views/BaseTractView.swift [hunk 7]
```diff
@@ -14,19 +14,34 @@ class BaseTractView: UIView {
     
     var data: XMLIndexer?
     var colors: TractColors?
-    var configurations = TractConfigurations()
+    var configurations: TractConfigurations?
```
ryancarlson: All of these, except contentView could be defined as "let" statements with no optionals. They are defined in the init() function and are not changed elsewhere in the project
````swift
let data: XMLIndexer
let colors: TractColors
let configurations: TractConfigurations
var contentView: TractRoot?
````

This is nice for two reasons:
1. It's clear that the values are not reassigned anywhere
2. We don't have to unwrap optionals everywhere we use those properties

contentView is still a var & optional because we need to access something on `self` which requires `super.init` to be called... and let's have to be initialized before `super.init` is called.

Comment (ryancarlson) [inline godtools/Views/BaseTractView.swift]: Actually to take this one step further. The first three (data, colors, configurations) are not used anywhere outside this class, so you don't need to store them as properties on the class at all unless I'm missing something?

File: godtools/Views/TractElements/TractElementFunctions.swift [hunk 8]
```diff
@@ -84,4 +85,17 @@ extension BaseTractElement {
         return (elementName!, namespace, kind, properties, children)
     }
     
+    func elementListener() -> String {
+        return ""
+    }
+    
+    func sendMessageToElement(tag: String) {
```
ryancarlson: In the previous version of GodTools, we build the events and listeners and notifications that are posted and observed. Did you consider that here? I'd like to discuss that some.

godtools/Views/TractElements/TractModal.swift
```diff
@@ -9,5 +9,101 @@
...
+    override func setupView(properties: [String: Any]) {
+        loadElementProperties(properties: properties)
+        TractBindings.addBindings(self)
+        
+        self.frame = buildFrame()
+        setupStyle()
+    }
+    
+    func setupStyle() {
+        self.backgroundColor = UIColor.black.withAlphaComponent(0.8)
```
ryancarlson: Two thoughts:
1. You could probably use .black.withAlphaComponent(0.8)
2. Will the modals always be black, or are they subject to styling through the manifest defaults? /cc @frett

godtools/Views/TractElements/TractModal.swift
```diff
@@ -9,5 +9,101 @@
...
+    }
+    
+    override func elementListener() -> String {
+        return self.properties.listener != nil ? self.properties.listener! : ""
```
ryancarlson:
````swift
return self.properties.listener ?? ""
````
should also work :)

File: godtools/Views/TractElements/TractModal.swift [hunk 11]
```diff
@@ -9,5 +9,101 @@
...
+        }
+    }
+    
+    override func receiveMessage() {
+        if self.modalOpen {
+            for view in self.subviews {
+                view.removeFromSuperview()
+            }
+            
+            self.alpha = CGFloat(0.0)
```
ryancarlson: Did you mean to set this here and just below in the animation?

godtools/Views/TractElements/TractParagraph.swift
```diff
@@ -43,14 +51,23 @@ class TractParagraph: BaseTractElement {
...
         var xMargin = BaseTractElement.xMargin
-        if BaseTractElement.isCardElement(self) {
-            xMargin = TractCard.xPaddingConstant
-        }
         
-        textStyle.font = .gtRegular(size: 18.0)
-        textStyle.width = self.width
-        textStyle.xMargin = xMargin
-        textStyle.color = self.textColor
+        if BaseTractElement.isModalElement(self) {
```
ryancarlson: Can you create functions here too to keep the code blocks smaller?

godtools/ViewControllers/Tract/TractViewController.swift
pablomarti: Thanks for this!

godtools/ViewControllers/Tract/TractViewController.swift
pablomarti: cool!

godtools/Views/Buttons/GTButton.swift
pablomarti: I was thinking about that, but the Sketch only display this one (also with the background color of the modal). I will ask in the chat about this

godtools/Views/TractElements/TractButton.swift
pablomarti: 👍

godtools/Configurations/TractBindings.swift
pablomarti: No, thanks for check that, I will refactor this portion

godtools/Views/TractElements/TractElementFunctions.swift
pablomarti: I got confused, what do you mean? I will look for that repo

godtools/Views/TractElements/TractModal.swift
pablomarti: It is needed to have `UIColor` class, and as we talked in the channel Ann said that the style will remain for every language and tool (will leave that for the moment, unless the decision change)

godtools/Views/TractElements/TractModal.swift
pablomarti: I changed this method, so now it is: 
```

override func elementListeners() -> [String]? {
        return self.properties.listener == nil ? nil : self.properties.listener?.components(separatedBy: ",")
    }
```

godtools/Views/TractElements/TractModal.swift
pablomarti: Thanks! I fixed this

godtools/Views/TractElements/TractElementFunctions.swift
ryancarlson: Here's some of the code. (it's kind of verbose), but basically what we did, is for any listener that's defined on a page or a modal, we look for notifications that have that event as a key.

https://github.com/CruGlobal/GTViewController/blob/master/GTViewController/Classes/Nav%20Classes/GTViewController.m#L590

godtools/Views/TractElements/TractParagraph.swift
pablomarti: 👍

godtools/Views/BaseTractView.swift
pablomarti: Thanks for checking this! these properties were from a previous version, now is not needed

godtools/XMLContent/KGP/kgp-5.xml
pablomarti: 👍

godtools/Views/TractElements/TractElementFunctions.swift
pablomarti: For the record, what I told you on Hipchat:

I created a dictionary for setting all the views that have a listener (lets call it dictionaryX), so, if an interaction is generated (like hitting a button), the button will dispatch a message with the values associated, and those values are going to be searched in the dictionaryX, and it will return the elements (views) with those listeners and will process the message, thats why I didn't see the need of using notifications for that (my only concern was about sending parameters to the view, but I already have an idea that I can develop on the top of that); were you talking about that or something else that I'm missing?

pablomarti:
Hey @ryancarlson I uploaded some changes, would you please check them? my only concern right now is about the bindings, not sure about your idea

godtools/XMLContent/KGP/kgp-5.xml
pablomarti: I changed the XML here https://github.com/CruGlobal/godtools-swift/pull/23/files#diff-85a7db7ebbc41f2b16813b30512ea219R88 and here https://github.com/CruGlobal/godtools-swift/pull/23/files#diff-85a7db7ebbc41f2b16813b30512ea219R102

godtools/Configurations/TractBindings.swift
```diff
@@ -0,0 +1,46 @@
...
+        TractBindings.dismissBindings = [String: BaseTractElement]()
+    }
+    
+    static func addBindings(_ element: BaseTractElement) {
+        TractBindings.addOpeningBindings(element)
+        TractBindings.addDismissBindings(element)
+    }
+    
+    static func addOpeningBindings(_ element: BaseTractElement) {
+        if element.elementListeners() == nil {
```
ryancarlson: Not a big deal as the code is still readable, but this and the next function could both be `guard let` statements. You don't have to change this, but you can if you want to :)

godtools/Views/TractElements/TractModal.swift
ryancarlson: 👍

ryancarlson:
@pablomarti I have concerns about how we're doing events and listeners. I want to merge this PR since some others are dependent upon it, but I want to discuss an alternate approach we can take to doing events and listeners (bindings) that I think can work better in some ways.

godtools/Configurations/TractBindings.swift
pablomarti: Hey @ryancarlson I will make this change once I get Xcode working in order to not make it directly

PR #24, opened 2017-05-19
Handle updates

ryancarlson:
Handles cases where the latest translation is updated on the API and that information is synced to the device


ryancarlson:
@pablomarti this is the first of three small PRs from things I accomplished today.

ryancarlson:
Thanks! I think I remove the prints in a later PR

PR #25, opened 2017-05-19
Initial app state

ryancarlson:
Includes changes that shows three resources (KGP, Satisfied and Four Spiritual Laws) by default when the app opens. It also shows English as an available language. Once the app has connected to the API and downloaded the resources from there, it will delete these temporary objects from the database.

Later we will include some default XML and images so that English can be viewed even if the app cannot connect to the API.

### DIFF
--- a/godtools/AppDelegate.swift
+++ b/godtools/AppDelegate.swift
@@ -19,10 +19,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     var window: UIWindow?
     var flowController: BaseFlowController?
     
-    override init() {
-        super.init()
-    }
-    
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
         initializeCoreDataStack()
         Fabric.with([Crashlytics.self, Answers.self])
@@ -80,9 +76,21 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // Initializes the importer so the resources directory can be created.
         _ = TranslationZipImporter.shared
         
+        if !UserDefaults.standard.bool(forKey: GTConstants.kFirstLaunchKey) {
+            initializeAppStateOnFirstLaunch()
+        }
+        
         return LanguagesManager.shared.loadFromRemote().then { (languages) -> Promise<[DownloadedResource]> in
             return DownloadedResourceManager.shared.loadFromRemote()
+        }.then { (resources) -> Promise<[DownloadedResource]> in
+            FirstLaunchInitializer().cleanupInitialAppState()
+            return Promise(value: resources)
         }
+    }
+    
+    private func initializeAppStateOnFirstLaunch() {
+        FirstLaunchInitializer().initializeAppState()
+        UserDefaults.standard.set(true, forKey: GTConstants.kFirstLaunchKey)
     }
 }
 
--- a/godtools/AppDelegate.swift
+++ b/godtools/AppDelegate.swift
@@ -84,6 +84,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             return DownloadedResourceManager.shared.loadFromRemote()
         }.then { (resources) -> Promise<[DownloadedResource]> in
             FirstLaunchInitializer().cleanupInitialAppState()
+            TranslationZipImporter.shared.catchupMissedDownloads()
             return Promise(value: resources)
         }
     }
--- a/godtools/AppDelegate.swift
+++ b/godtools/AppDelegate.swift
@@ -74,7 +74,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     
     private func initalizeAppState() -> Promise<Any> {
         // Initializes the importer so the resources directory can be created.
-        _ = TranslationZipImporter.shared
+        TranslationZipImporter.setup()
         
         if !UserDefaults.standard.bool(forKey: GTConstants.kFirstLaunchKey) {
             initializeAppStateOnFirstLaunch()
--- a/godtools/Managers/DownloadResources/DownloadedResourceManager.swift
+++ b/godtools/Managers/DownloadResources/DownloadedResourceManager.swift
@@ -48,9 +48,6 @@ class DownloadedResourceManager: GTDataManager {
                     return Promise(error: error)
                 }
                 return Promise(value:self.loadFromDisk())
-            }.then { downloadedResources -> Promise<[DownloadedResource]> in
-                TranslationZipImporter.shared.catchupMissedDownloads()
-                return Promise(value: downloadedResources)
             }
     }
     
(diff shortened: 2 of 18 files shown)


ryancarlson:
@pablomarti here is the second PR. You could review this after #24, or just review this one since #24 is smaller and this PR includes all those commits. Either way is fine with me. :)

godtools/Managers/DownloadResources/DownloadedResourceManager.swift
```diff
@@ -52,8 +52,6 @@ class DownloadedResourceManager: GTDataManager {
     }
     
     private func saveToDisk(_ resources: [DownloadedResourceJson]) {
-        let context = NSManagedObjectContext.mr_default()
-        
         for remoteResource in resources {
             let cachedResource = DownloadedResource.mr_findFirstOrCreate(byAttribute: "remoteId", withValue: remoteResource.id!, in: context)
```
pablomarti: Hey I get lost, how does it get the context?

ryancarlson: From the superclass GTDataManager :)

godtools/Managers/TranslationZipImporter.swift
```diff
@@ -152,8 +149,52 @@ class TranslationZipImporter {
...
+        let files = try FileManager.default.contentsOfDirectory(atPath: directoryPath)
+        
+        for filename in files {
+            let referencedFile = findEntity(ReferencedFile.self, byAttribute: "filename", withValue: filename) ?? createEntity(ReferencedFile.self)
+            referencedFile?.filename = filename
+            referencedFile?.addToTranslations(translation)
+        }
+    }
+    
+    private func moveFilesFrom(directoryPath: String) throws {
```
michaelharro: moving files from a temp directory to a permanent one is covered in http://nshipster.com/nstemporarydirectory/ as well

godtools/ViewControllers/Platform/HomeViewController.swift
```diff
@@ -63,12 +63,21 @@ class HomeViewController: BaseViewController {
...
+                                               object: nil)
     }
     
-    @objc fileprivate func presentLanguageSettings() {
+    @objc private func presentLanguageSettings() {
         self.delegate?.moveToUpdateLanguageSettings()
     }
     
+    @objc private func reloadTableData() {
+        self.toolsManager.delegate = self
```
michaelharro: this could cause memory issues as ToolsManager's delegate is not a weak reference

godtools/Managers/FirstLaunchInitializer.swift
```diff
@@ -0,0 +1,101 @@
...
+//  Created by Ryan Carlson on 5/19/17.
+//  Copyright © 2017 Cru. All rights reserved.
+//
+
+import Foundation
+import CoreData
+
+class FirstLaunchInitializer: GTDataManager {
+    
+    private let magicId = "-358"
```
michaelharro: Does "-358" have meaning or is it random?

ryancarlson:
@michaelharro thanks for the review. you caught some good things in here. i'll work through these comments.

godtools/Managers/FirstLaunchInitializer.swift
ryancarlson: It's random. I just picked something negative as I knew the rails API would never generate that. I happen to like the number 358.

godtools/Managers/GTDataManager.swift
```diff
@@ -13,8 +13,18 @@ import Spine
...
 
 class GTDataManager: NSObject {
+    let documentsPath: String
+    let resourcesPath: String
     
     let serializer = Serializer()
+    let context = NSManagedObjectContext.mr_default()
+    
+    override init() {
+        documentsPath = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true)[0]
```
ryancarlson: So this one's a little interesting. It requires me to unwrap the value returned by `.first` or provide a default. Unwrapping the value is just as risky as using `[0]` and i don't know what default to provide. What do you think?

godtools/Managers/GTDataManager.swift
```diff
@@ -31,8 +41,40 @@ class GTDataManager: NSObject {
...
+    
+    func findAllEntities<T: NSManagedObject>(_ entityClass: T.Type) -> [T] {
+        return entityClass.mr_findAll(in: context) as! [T]
+    }
+    
+    func createEntity<T: NSManagedObject>(_ entityClass: T.Type) -> T? {
+        return entityClass.mr_createEntity(in: context)
+    }
+    
+    func deleteEntities<T: NSManagedObject>(_ entityClass: T.Type, matching: NSPredicate) {
```
ryancarlson: This function does match:
````swift
func findEntities<T: NSManagedObject>(_ entityClass: T.Type, matching: NSPredicate) -> [T]
````
which returns the entities "matching" the predicate. So it is consistent in that sense?

To me, `deleteAllEntities` would not take a predicate and just delete all records in that table.

File: godtools/AppDelegate.swift [hunk 7]
```diff
@@ -80,9 +76,22 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // Initializes the importer so the resources directory can be created.
         _ = TranslationZipImporter.shared
```
ryancarlson: That changes the declaration from `let` to `var` and makes `shared` and optional. Unless there's a way to have the setup `class` function call a private init method? Have you done something like the before? I'd strongly prefer the singleton not be an optional that has to be unwrapped each time.

````swift
    var shared: TranslationZipImporter?
    
    class func setup() {
        shared = TranslationZipImporter()
    }
````

File: godtools/Managers/TranslationFileRemover.swift [hunk 8]
```diff
@@ -0,0 +1,25 @@
...
+class TranslationFileRemover: GTDataManager {
+    
+    func deleteUnusedPages() {
+        for page in findAllEntities(ReferencedFile.self) {
+            do {
+                if page.translationsAsArray().filter({ $0.isDownloaded }).count == 0 {
+                    try FileManager.default.removeItem(atPath: "\(resourcesPath)/\(page.filename!)")
+                }
+            } catch {
+                Crashlytics().recordError(error, withAdditionalUserInfo: ["customMessage": "Error deleting file \(page.filename!) no longer referenced by any translations"])
```
ryancarlson: How about we consider this as a task for later? I definitely want to make sure that we're doing proper error handling throughout the project. We didn't in MH and that's made debugging painful. However there's other more pressing that need working out before we come back to this in my opinion.

godtools/ViewControllers/Platform/HomeViewController.swift
ryancarlson: I made the delegate variable weak.

ryancarlson:
@michaelharro I've made some changes, the things outstanding are:
 - reworking the lambda filtering into predicates (I think we agree this can wait until we move to Realm?)
 - reworking Crashlytics reporting (we agree this can wait?)
 - moving the files from the temporary directory differently (I don't follow what the link you provided is eluding too, is this a big deal for now?)

ryancarlson:
thanks for the review!

PR #26, opened 2017-05-19
Delete unused files

ryancarlson:
Once all translations that use any given file have been marked as "deleted" or "not downloaded", then delete that file from the disk.


ryancarlson:
@pablomarti here is the third PR I mentioned. This can be reviewed after #25

PR #28, opened 2017-05-22
Add logic to show light content (white) status bar on all view controllers

ryancarlson:
- status bar will be shown by default, with option to hide it
 - tract view controller will opt to hide it


ryancarlson:
@michaelharro can you review this small PR?

ryancarlson:
Thanks. Actually though this change suggests that the background is black and therefore the status bar content is white. The end result is this: 
simulator screen shot may 22 2017 3 24 26 pm

michaelharro:
ok, I wasn't paying much attention to the specifics. I just saw you used the correct technique and was going to do a visual check. It looks correct so it is good to go

ryancarlson:
Cool, thanks!

PR #29, opened 2017-05-23
Use convenience methods to get array of translations

ryancarlson:


ryancarlson:
Here's a tiny PR @michaelharro . The majority of the changes for deleting files no longer used was included in #25 .

PR #30, opened 2017-05-24
Remove switches for preview mode and notifications from menu

ryancarlson:
They will not be supported in this version.


ryancarlson:
@pablo here is a very easy PR, could you review please?

PR #32, opened 2017-05-24
Download and Render Banners on Home, Add Tools and Tool Detail views

ryancarlson:
This PR:

- creates BannerManager to handle the networking and persisting/retrieval of banners
- adds an Attachment class to start to track references of downloaded files for deletion later when they are unused
   - this PR only tracks banners and doesn't do any deletion
- reworks HomeTableViewCells to render well on all devices and with or without banner
   - introduces spacer views on the left and right to have nice padding
- renders banners throughout platform


ryancarlson:
@michaelharro or @pablomarti could you review these changes please?

godtools/Managers/GTDataManager.swift
```diff
@@ -15,13 +15,15 @@ import CoreData
...
     let resourcesPath: String
+    let bannersPath: URL
     
     let serializer = Serializer()
     let context = NSManagedObjectContext.mr_default()
     
     override init() {
         documentsPath = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true)[0]
-        resourcesPath = "\(documentsPath)/Resources"
+        resourcesPath = documentsPath.appending("/").appending("Resources")
```
ryancarlson: I'd like to convert these to all be URLs soon

godtools/Managers/BannerManager.swift
```diff
@@ -0,0 +1,126 @@
...
+        if FileManager.default.fileExists(atPath: bannersPath.path, isDirectory: nil) {
+            return
+        }
+        
+        do {
+            try FileManager.default.createDirectory(at: bannersPath,
+                                                    withIntermediateDirectories: false,
+                                                    attributes: nil)
+        } catch {
+            Crashlytics().recordError(error, withAdditionalUserInfo: ["customMessage": "Error creating Banners directory on device."])
```
pablomarti: Maybe we should make a function for setting the errors?

ryancarlson: @michaelharro suggested that too. I could take a look at that on Friday when I'm working on GodTools again. We would probably do something notifications based, so anywhere we catch an error posts a notification with an optional message, and then one central place listens for those notifications and reports to Crashlytics. Sound good to you guys?

godtools/ViewControllers/Platform/ToolDetailViewController.swift
```diff
@@ -42,6 +43,7 @@ class ToolDetailViewController: BaseViewController {
             .joined(separator: ", ")
         
         self.displayButton()
+        self.bannerImageView.image = BannerManager.shared.loadFor(resource!)
```
pablomarti: I like the return of a promise with the image

godtools/ViewControllers/Platform/ToolDetailViewController.xib
```diff
@@ -25,31 +26,24 @@
             <rect key="frame" x="0.0" y="0.0" width="375" height="667"/>
             <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
             <subviews>
-                <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="yUi-li-bC7">
+                <imageView userInteractionEnabled="NO" contentMode="scaleAspectFill" horizontalHuggingPriority="251" verticalHuggingPriority="251" highlightedImage="cell_banner_placeholder" translatesAutoresizingMaskIntoConstraints="NO" id="6sd-7g-yCc">
```
pablomarti: This placeholder will only affect for when highlighted, shouldn't we set the image in case the promise fails?

godtools/ViewControllers/Platform/ToolDetailViewController.swift
ryancarlson: In this case we are just loading from the disk. I don't think we need to use a promise here.

godtools/ViewControllers/Platform/ToolDetailViewController.xib
ryancarlson: Sí, good idea!

godtools/ViewControllers/Platform/ToolDetailViewController.swift
pablomarti: Yeah, sorry, I confused with `downloadFor`

godtools/Managers/BannerManager.swift
pablomarti: Sounds good!

ryancarlson:
@pablomarti i fixed the placeholder. I think this is good to go, do you agree?

pablomarti:
Yes! Great @ryancarlson !

PR #33, opened 2017-05-24
Enable transition from menu to language settings view

ryancarlson:
- move some methods from PlatformFlowController to BaseFlowController to make this happen

### DIFF
--- a/godtools/FlowControllers/BaseFlowController.swift
+++ b/godtools/FlowControllers/BaseFlowController.swift
@@ -122,6 +122,11 @@ class BaseFlowController: NSObject, UINavigationControllerDelegate {
 
 extension BaseFlowController: BaseViewControllerDelegate {
     
+    func goHome() {
+        _ = self.currentViewController?.navigationController?.popToRootViewController(animated: true)
+        resetNavigationControllerColorToDefault()
+    }
+    
     func goBack() {
         _ = self.currentViewController?.navigationController?.popViewController(animated: true)
         resetNavigationControllerColorToDefault()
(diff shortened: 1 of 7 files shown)


ryancarlson:
@pablomarti hey can you look at this one? there may be a better way with the flow controllers, but this seems to work. let me know?

pablomarti:
Hey @ryancarlson I uploaded a change, please see my last commit. Now when accessing from the menu to the Language Settings, when going back will move to Home. I have found 2 bugs:
1 - If doing this, a title appears at the left, at the side of the arrow to go back
2 - when pressing "Done" on Menu, it moves back to Home but the transition goes black, so its ugly (that is separated from the objective of this PR, but is related)

What do you think? Do we work those things on another PR or do we do that here? I review your changes and are ok, please check my commit and let me know, thanks!

godtools/FlowControllers/BaseFlowController.swift
```diff
@@ -105,6 +108,14 @@ class BaseFlowController: NSObject, UINavigationControllerDelegate {
                         dst?.view.transform = CGAffineTransform(translationX: 0, y: 0) },
                        completion: { finished in
                         _ = navigationController?.popViewController(animated: false) } )
+        
+        self.shouldRemoveMenu = false
+    }
+    
+    func removeMenuViewController() {
+        let menuPosition = (self.currentViewController?.navigationController?.viewControllers.count)! - 2
```
ryancarlson: @pablomarti rather than doing a count, wouldn't it be better to loop through the view controller stack and find one that `isKindOf: MainMenuController.self`? That would be safer than this if another view controller got pushed on to the stack somehow in the future.

PR #37, opened 2017-05-26
Work of fixing UI bugs on Cards

pablomarti:
This PR targets:
GT-303: Fix swipe up and swipe down when navigating through cards
GT-295: Add shadows to cards of tract
GT-299: Fix overlay at the bottom of the card only when content height is higher than card height

Also:

- it fixes the problem that the "Call to action" element was not showing up when swiping up the last card.
- adds new strategy for the management of properties
- makes a refactoring for standarize the setup of elements 
- allow to override colors from configuration of XML for `<page>` objects

This PR should be checked after the PR #23 is checked and merged.


godtools/Views/TractElements/BaseTractElement.swift
```diff
@@ -209,8 +209,14 @@ class BaseTractElement: UIView {
     
     func setupView(properties: Dictionary<String, Any>) {
         self.frame = buildFrame()
+        loadElementProperties(properties: properties)
```
ryancarlson: you could name the method this way:
````swift
func loadElementProperties(_ properties: properties)

File: godtools/Views/TractElements/TractButton.swift [hunk 2]
```diff
@@ -78,6 +82,24 @@ class TractButton: BaseTractElement {
...
+    override func buildFrame() -> CGRect {
+        return CGRect(x: self.xPosition,
+                      y: self.yPosition,
+                      width: self.width,
+                      height: self.height)
+    }
+    
+    override func textStyle() -> TractTextContentProperties {
+        let textStyle = super.textStyle()
+        textStyle.font = .gtRegular(size: 18.0)
```
ryancarlson: Are we supporting text size scaling for buttons? /cc @Omicron7 @frett ?

godtools/Configurations/TractColors.swift
```diff
@@ -8,10 +8,26 @@
...
 import UIKit
 
-class TractColors {
+class TractColors: NSObject {
     
     var primaryColor: UIColor?
     var primaryTextColor: UIColor?
     var textColor: UIColor?
+    
+    override init() {
```
pablomarti: I believe so... Normally, I will delete it and see what happens (because I don't remember) but I believe it was needed in order to make the other init (currently I don't have Xcode for testing)

godtools/Views/TractElements/TractButton.swift
pablomarti: It should be supported, I believe it was talked on a meeting (I have to wait for that, so I can make some changes on some properties classes)

godtools/Views/TractElements/TractLink.swift
```diff
@@ -21,8 +21,29 @@ class TractLink: TractButton {
...
     }
+    
+    override func addTargetToButton() {
+        if self.properties.type == .url {
+            self.button.addTarget(self, action: #selector(buttonTarget), for: .touchUpInside)
+        }
+    }
+    
+    override func buttonTarget() {
+        let events = self.properties.events!.components(separatedBy: ",")
```
pablomarti: You are right, it may get a nil, will work on that

godtools/Configurations/TractColors.swift
pablomarti: Confirmed, it is needed when generated custom inits.

ryancarlson: Okay, cool!

PR #39, opened 2017-05-26
When banner is nil, then set the placeholder image.

ryancarlson:
This ensures that when a resource is removed, the cell below it doesn't accidentally display the banner for the resource just removed.

The use case is, cell n has a banner and n+1 does not. Cell n's resource is removed, and when coming back to the home view cell n+1 (now cell n) renders the banner for the resource just removed because the banner passed into the config method was nil, and there was nothing to handle that.


ryancarlson:
@pablomarti here is a very small PR for a bug i just noticed.

PR #43, opened 2017-05-26
Localized name description

ryancarlson:
Retrieve localized name and description when updating translations from the database

Use these values to show throughout the app where resource.name was used.


ryancarlson:
note: this change will not appear in the app until a new translation is available for a resource-language. i may refactor that later..

PR #45, opened 2017-05-31
Add functionality to segemented control to change language of tract pages

pablomarti:


ryancarlson:
@pablomarti never mind my last comment. the issue was actually because the API was not allowing downloads of the XML. i fixed that, downloaded again, and this works.

pablomarti:
Thanks @ryancarlson sorry for making a long commit, I was making a lot of experiments, and I wasn't sure of some things that at the end got lost, so at the end I wrapped everything and did that this way, will not happen again, thanks!

PR #47, opened 2017-05-31
Fix infinite loop when merging build number bump commits

ryancarlson:
- instead of pushing to a branch and PRing, allow bump commits to go straight to develop with a token to skip the CI build if it's a version number bump commit.
 - this requires installing a new plugin to jenkins which looks for that token
 - and it requires allowing cru-jenkins-mobile to push to develop which is fine


fastlane/Fastfile
```diff
@@ -99,16 +99,11 @@ platform :ios do
 
     clean_build_artifacts
     commit_version_bump(xcodeproj: "godtools.xcodeproj",
-                        message: "Build number bump for Beta build #{lane_context[SharedValues::BUILD_NUMBER]}",
+                        message: "Build number bump for Beta build #{lane_context[SharedValues::BUILD_NUMBER]} [ci skip]",
                         force: true)
     add_git_tag
-    push_to_git_remote(local_branch: ENV['GIT_BUILD_NUMBER_BRANCH'] || git_branch,
+    push_to_git_remote(local_branch: ENV['GIT_BRANCH'] || git_branch,
```
michaelharro: If you are going to change this you will also need to change the cru_git_setup helper method that is run at the beginning of the beta and production lanes.

You should also make all these same changes to the production lane.

Once you have a working solution make these changes to the Fastfile.template so that new projects get these fixes too. It is found here https://github.com/CruGlobal/mobile-deployment-config/blob/master/ios/setup/Fastfile.template

fastlane/Fastfile
```diff
@@ -99,16 +99,11 @@ platform :ios do
 
     clean_build_artifacts
     commit_version_bump(xcodeproj: "godtools.xcodeproj",
-                        message: "Build number bump for Beta build #{lane_context[SharedValues::BUILD_NUMBER]}",
+                        message: "Build number bump for Beta build #{lane_context[SharedValues::BUILD_NUMBER]} [ci skip]",
```
michaelharro: from what I understand the ci skip feature is a part of the github pull request builder jenkins plugin which is what we use for the test jobs when PRs are created. However, we use the git plugin for detecting merged PRs on develop and master. The git plugin does not have this functionality to my knowledge. Here is the documentaion for the yaml you are able to define on each of them so you can check me on that https://docs.openstack.org/infra/jenkins-job-builder/triggers.html#triggers.github-pull-request https://docs.openstack.org/infra/jenkins-job-builder/scm.html#scm.git

PR #48, opened 2017-05-31
Fix crash b/c String cannot be converted to NSTextAlignment

ryancarlson:
This needs another look for supporting right to left languages


godtools/Views/TractElements/TractTextContentProperties.swift
```diff
@@ -35,10 +35,23 @@ class TractTextContentProperties: TractElementProperties {
             case "text-scale":
                 self.scale = properties[property] as? CGFloat
             case "text-align":
-                self.align = (properties[property] as? NSTextAlignment)!
+                self.align = convertTextAlignmentString(properties[property] as! String?)
```
pablomarti: Great! thanks for that!

PR #50, opened 2017-06-01
Refactor tract bindings

pablomarti:
Changes:
- Now instead of looking for the id of the page, it looks for the listeners of each page
- The listeners of the pages are stored as bindings before generating all of the content
- When an event is dispatched, it looks on the bindings related to the event, including the page bindings
- It is possible to move from one page to another, even if the pages are distant and one of them is currently not loaded into memory

With this approach I believe that if the app grows to the point of moving from one tool to another is going to be possible, a new binding list may be added to TractBindings in order to manage that, which will have to work with a structure that contains the bindings of the other tool (which will require to be already downloaded).


pablomarti:
For testing, I set in the dummy XML some listeners in the 5th page, one that points to the 6th, and one that points to the 2nd page (which, for when the user is seeing the 5th view the 2nd view is not on memory).

godtools/ViewControllers/Tract/TractViewControllerPageMovements.swift
```diff
@@ -17,14 +17,12 @@ extension TractViewController {
             return
         }
         
-        let destinationViewPageId = dictionary["pageId"]
-        for (pageId, pageIndex) in self.pagesIds {
-            if pageId == destinationViewPageId {
-                self.currentPage = pageIndex
-                break
-            }
+        let pageListener = dictionary["pageListener"]
```
ryancarlson: I would do a `guard let` here too. If the listener is not present for some reason, then the app would crash on the next line.

godtools/Views/TractElements/TractElementFunctions.swift
```diff
@@ -93,16 +93,20 @@ extension BaseTractElement {
         return nil
     }
     
-    func sendMessageToElement(tag: String) {
-        if TractBindings.bindings[tag] != nil {
-            guard let view = TractBindings.bindings[tag] else { return }
+    func sendMessageToElement(listener: String) {
+        if TractBindings.bindings[listener] != nil {
```
pablomarti: The reason why I did this is because there are other conditions, so if it returns it will not check the other conditions, and, just in case, there may be a case where the listener appear on different places, on `listeners` and `dismiss-listeners` attributes

pablomarti:
Hey @ryancarlson just uploaded a change and wrote a comment, what do you think?

godtools/Views/TractElements/TractElementFunctions.swift
ryancarlson: Claro! That makes sense :)

pablomarti:
Thanks @ryancarlson !

PR #53, opened 2017-06-02
Track views (2)

ryancarlson:
This PR adds the ability to record with the API that a resource has been viewed. If the request fails, then a counter is incremented on DownloadedResource object. These stored values are then synced on app re-launch. If successful the counter resets to 0.

Another PR b/c I did the initial merge wrong. @pablomarti you've already reviewed these commits and approved in the previous PR. Can I merge to develop now?


pablomarti:
Ok @ryancarlson

PR #56, opened 2017-06-02
Fix UI problems

pablomarti:
- Fix the width and position of the hero component
- Set the font size following the design
- Allow to access to the webpage from a button (new XML added for testing)
- Build class for management of pagination between Tract pages
- Do not display call to action on last page


godtools/Views/TractElements/TractButtonActions.swift
```diff
@@ -7,13 +7,21 @@
...
-        let values = self.properties.value!.components(separatedBy: ",")
-        for value in values {
-            sendMessageToElement(listener: value)
+        if self.properties.type == .event {
+            let values = self.properties.value!.components(separatedBy: ",")
+            for value in values {
+                sendMessageToElement(listener: value)
+            }
+        }
+        else if self.properties.type == .url {
```
ryancarlson: Minor thing, but can you move the else up to the line above?

PR #57, opened 2017-06-02
Changes on the UI

pablomarti:
GT-331: Move back from Tool Detail to Home after the tool is downloaded or deleted
GT-333: Do not display title if there is only one language available and do not display segmented control if the primary language is the same to the parallel language
GT-334: Remove exclamation image from Language Settings

### DIFF
--- a/godtools/ViewControllers/Tract/TractViewController.swift
+++ b/godtools/ViewControllers/Tract/TractViewController.swift
@@ -76,10 +76,8 @@ class TractViewController: BaseViewController {
     }
 
     override func displayScreenTitle() {
-        if parallelLanguageIsAvailable() {
+        if parallelLanguageIsAvailable() && determinePrimaryLabel() != determineParallelLabel() {
             self.navigationItem.titleView = languageSegmentedControl()
-        } else {
-            self.title = currentTractTitle()
         }
     }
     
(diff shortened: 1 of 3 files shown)


pablomarti:
There is a small problem, after the download is complete it should move back to Home from Tool Detail, and it is doing that, the problem is that not always the progressview goes to 100%, because, the animation is "too slow". I tried working with the notifications (current behaviour), also I updated the manager to work with Promises but didn't make a difference (so I turned back to how it was it), and I did try not to use the progress view animation but build a custom animation, but didn't work. What I did for decrease this problem was to set a higher value to the delay when moving back to Home, but, the progress view is like 80% when moving. Do you want me to continue solving this @ryancarlson or should I leave it like it is right now?

godtools/ViewControllers/Tract/TractViewController.swift
```diff
@@ -76,10 +76,8 @@ class TractViewController: BaseViewController {
     }
 
     override func displayScreenTitle() {
-        if parallelLanguageIsAvailable() {
+        if parallelLanguageIsAvailable() && determinePrimaryLabel() != determineParallelLabel() {
```
ryancarlson: Ah, yes I want to do a small fix to make sure that the user cannot choose the same language for primary and parallel. Then we don't have to do this in the future. This works for now though. 👍

PR #59, opened 2017-06-05
Add actions to menu

pablomarti:
This PR contains:
- GT-326: share GodTools with native Share sheet
- GT-329: add WebView and display GodTools website as temporal page
- GT:327: if the user has configured the email account open mailer for sending a mail, otherwise displays a webview with the contact page
- GT:328: if the user has configured the email account open mailer for sending a mail, otherwise displays a webview with the contact page
- GT:330: a viewcontroller for about, where it has 9 labels with dynamic height inside of a scrollview

### DIFF
--- a/godtools/Localizable.strings
+++ b/godtools/Localizable.strings
@@ -26,17 +26,6 @@
 "privacy_policy" = "Privacy Policy";
 "share_god_tools_share_sheet_text" = "Here you can download GodTools app: http://godtoolsapp.com";
 
-// About texts
-"about_text1" = "Text1";
-"about_text2" = "Text2";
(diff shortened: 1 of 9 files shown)


pablomarti:
Some of the texts are hardcoded, like the subject of emails and the links, because, I don't know if it is going to be needed to have localized the texts (because is the subject for the email to the GT team) and same with the links, let me know what do you think @ryancarlson, thanks!

godtools/Localizable.strings
```diff
@@ -24,6 +24,18 @@
 "share_a_story_with_us" = "Share a Story With Us";
 "terms_of_use" = "Terms of Use";
 "privacy_policy" = "Privacy Policy";
+"share_god_tools_share_sheet_text" = "Here you can download GodTools app: http://godtoolsapp.com";
+
+// About texts
+"about_text1" = "Text1";
```
ryancarlson: I think you want the "general_about" strings below.

pablomarti:
Great thanks @ryancarlson ! I didn't see the texts, now they are in place

PR #60, opened 2017-06-05
GT-294: Change style of navigation bar of tract

pablomarti:
This PR sets the navigation bar items to have the color defined in the primary text color of the manifest. The background color also changes, but first it transforms the color with an alpha to a similar color without alpha, removing the transparency.


pablomarti:
The problem right now @ryancarlson is that I'm not able to make the color to process the alpha channel, or maybe I'm missing something, so this PR is WIP.

pablomarti:
@ryancarlson I made some changes, transforming the color with transparency to a color without transparency, following: https://www.viget.com/articles/equating-color-and-transparency please let me know what do you think

PR #62, opened 2017-06-06
GT-337: Fix crash when moving through pages

pablomarti:
This PR changes the way of how the views are managed. It will always have a max of 5 items, but will have a specific space for storing each view (that, when not used will be nil). This fix the crash and helps by making the code easy to understand and handle, and allows to move from long distance views.

### DIFF
--- a/godtools/ViewControllers/Tract/TractViewControllerContent.swift
+++ b/godtools/ViewControllers/Tract/TractViewControllerContent.swift
@@ -50,16 +50,16 @@ extension TractViewController {
         let width = self.containerView.frame.size.width
         let height = self.containerView.frame.size.height
         let lastPosition = self.totalPages() - 1
-        var tmpPagesViews = [BaseTractView?](repeating: nil, count: 10)
+        var tmpPagesViews = [BaseTractView?](repeating: nil, count: totalPages())
         
         for position in range.start...range.end {
-            let pageView = self.pagesViews[position]
-            if pageView != nil {
-                tmpPagesViews[position] = pageView!
+            if let pageView = self.pagesViews[position] {
+                tmpPagesViews[position] = pageView
             } else {
                 let view = buildPage(position, width: width, height: height)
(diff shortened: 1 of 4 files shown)


godtools/ViewControllers/Tract/TractViewControllerContent.swift
```diff
@@ -47,39 +47,30 @@ extension TractViewController {
...
-                self.pagesViews.removeFirst()
-            }
-            
-        } else if firstTag > range.start {
-            var position = 0
-            for _ in range.start...firstTag {
-                let view = buildPage(range.start + position, width: width, height: height)
-                self.pagesViews.insert(view, at: position)
+        for position in range.start...range.end {
+            let pageView = self.pagesViews[position]
```
ryancarlson: This could be shortened to use `if let` which I think unwraps for you
````
if let pageView = self.pageViews[position] {
 tmpPageViews[position] = pageView 
} else {
 //el mismo code aqui
}

File: godtools/ViewControllers/Tract/TractViewControllerContent.swift [hunk 2]
```diff
@@ -106,6 +97,11 @@ extension TractViewController {
         }
         
         self.pagesViews.removeAll()
+        resetPagesView()
+    }
+    
+    func resetPagesView() {
+        self.pagesViews = [BaseTractView?](repeating: nil, count: 10)
```
ryancarlson: Are we sure the count should always be 10?

ryancarlson:
Otherwise I think this logic works!

godtools/ViewControllers/Tract/TractViewControllerContent.swift
pablomarti: Oh! that is a horrible mistake, sorry! was so focus on something I have on paper that forgot to use the variable

pablomarti:
Hey @ryancarlson just uploaded the changes for both things, thanks!

PR #63, opened 2017-06-07
GT-298: Sends an email dispatched from tract

pablomarti:
This PR adds an idea for how emails can be managed. It gets the email from xml and sends when the action is dispatched.

### DIFF
--- a/godtools/ViewControllers/Tract/TractViewControllerActions.swift
+++ b/godtools/ViewControllers/Tract/TractViewControllerActions.swift
@@ -18,20 +18,28 @@ extension TractViewController: MFMailComposeViewControllerDelegate {
                 return
             }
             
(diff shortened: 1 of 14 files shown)


godtools/ViewControllers/Tract/TractViewControllerActions.swift
```diff
@@ -0,0 +1,37 @@
...
+
+extension TractViewController: MFMailComposeViewControllerDelegate {
+    
+    func sendEmail(notification: Notification) {
+        if MFMailComposeViewController.canSendMail() {
+            guard let dictionary = notification.userInfo as? [String: Any] else {
+                return
+            }
+            
+            let composeVC = MFMailComposeViewController()
```
ryancarlson: This section and the else block below could be extracted into methods for good readability.

PR #67, opened 2017-06-07
GT-300: Send information of contact form of tract

pablomarti:
This PR adds:
- Adds a manager for followups, and the action for create a subscriber
- Gets the data from the form
- Adds global listeners
- When the listener for "followup:send" is dispatched, a subscriber is created


pablomarti:
Hey @ryancarlson @frett this is still on WIP, it is almost done, just have to check some things, and also see that the data is being saved. What do you think? And what do you think about the change in the XML for `<form>`?

pablomarti:
Hey @ryancarlson would you please check this PR? I made the changes from what @frett say about not creating the requests dynamically, so I created a manager for followups, I hope everything is ok. Also I created the place for setting the global listeners, please let me know what do you think, thanks!

pablomarti:
Hey @ryancarlson I finished merging develop into this PR and it is working. Would you please check it?

godtools/Managers/FollowUpsManager.swift
```diff
@@ -0,0 +1,30 @@
...
+
+import UIKit
+import PromiseKit
+
+class FollowUpsManager: GTDataManager {
+    
+    let path = "follow_ups"
+    
+    func createSubscriber(params: [String: String]) -> Promise<Void>? {
+        showNetworkingIndicator()
```
ryancarlson: Is the networking indicator being hidden somewhere else? I don't see it in this PR.

godtools/Managers/FollowUpsManager.swift
```diff
@@ -0,0 +1,30 @@
...
+        
+        return issuePOSTRequest(params)
+            .then { data -> Promise<Void> in
```
ryancarlson: We should do some kind of error handling, even if it is only logging the error to crashlytics.

godtools/Constants/GTGlobalTractBindings.swift
```diff
@@ -0,0 +1,27 @@
...
+    
+    enum Bindings:String {
+        case followUp = "followup:send"
+    }
+    
+    static func listen(listener: String, element: BaseTractElement) {
+        switch listener {
+        case Bindings.followUp.rawValue:
+            let manager = FollowUpsManager()
+            _ = manager.createSubscriber(params: element.getFormData()!)
```
ryancarlson: `getFormData()` can return nil, I would extract out that call and use a `guard let` assignment to avoid any crashes here.

godtools/Managers/FollowUpsManager.swift
pablomarti: Oh wow ok I will check why is not showing up

godtools/Managers/FollowUpsManager.swift
pablomarti: Ok

godtools/Constants/GTGlobalTractBindings.swift
pablomarti: Good idea, thanks!

pablomarti:
Just made the changes @ryancarlson , what do you think?

PR #69, opened 2017-06-08
Realm

ryancarlson:
This PR converts from CoreData and Magical Record for data tracking. There are several known issues at this point and this could be considered WIP.
 - error handling is really bad. we force try all of the writes with no error handling.
 - write block boundaries are very inconsistent.. but this is hard to clean up because
 - manager classes are bloated and convoluted.

As separate tasks I want to clean up the manager classes and then create consistent boundaries for write blocks.

One known practical issue is that users will seem like they have nothing downloaded when installing this build on top of a core data backed build. This is pre-production, so that can be handled by communication.


ryancarlson:
@pablomarti, I've tested this build out and it works pretty well, with the one practical issue I noted. There is more that should be done to clean up the code - which I've noted here. I wanted to open a PR though because there's a lot of work that I don't want to get bigger.. and this seems to work from what testing I have done.

godtools/Managers/DownloadResources/DownloadedResourceManager.swift
```diff
@@ -11,14 +11,14 @@ import UIKit
...
-import MagicalRecord
+import RealmSwift
 
 class DownloadedResourceManager: GTDataManager {
     static let shared = DownloadedResourceManager()
     
     let path = "/resources"
     
-    var resources = [DownloadedResource]()
+    var resources = List<DownloadedResource>()
```
pablomarti: I believe you can set `typealias DownloadResources = List<DownloadedResource>` outside of the class, so you can use that smaller `DownloadResources` everywhere instead of using `List<DownloadedResource>`

godtools/Managers/DownloadResources/DownloadedResourceManager.swift
```diff
@@ -52,79 +52,113 @@ class DownloadedResourceManager: GTDataManager {
...
+                    let remoteTranslation = remoteTranslationGeneric as! TranslationResource
+                    let languageId = remoteTranslation.language?.id ?? "-1"
+                    let resourceId = remoteResource.id ?? "-1"
+                    let version = remoteTranslation.version!.int16Value
+                    
+                    if !translationShouldBeSaved(languageId: languageId, resourceId: resourceId, version: version) {
+                        continue;
+                    }
+                    
+                    var cachedTranslation = findEntity(Translation.self,byAttribute: "remoteId",withValue: remoteTranslation.id!)
```
pablomarti: Maybe we can make methods that use the decorator pattern, like: `findEntityByRemoteId`, what do you think?

pablomarti: So, maybe, something like: `manager.findEntityByRemoteId("remoteTranslation.id!")` where manager is an instance of TranslationsManager, what do you think?

pablomarti: That would affect other parts for consistency in order to use the pattern if you decide for it

godtools/Managers/DownloadResources/DownloadedResourceManager.swift
```diff
@@ -52,79 +52,113 @@ class DownloadedResourceManager: GTDataManager {
...
-        saveToDisk()
     }
 
     private func translationShouldBeSaved(languageId: String, resourceId: String, version: Int16) -> Bool {        
         let predicate = NSPredicate(format: "language.remoteId = %@ AND downloadedResource.remoteId = %@",
                                     languageId,
                                     resourceId)
         
-        let existingTranslations: [Translation] = findEntities(Translation.self, matching: predicate)
+        let existingTranslations = findEntities(Translation.self, matching: predicate)
```
pablomarti: Same here, we can set a manager instance of TranslationsManager and just pass the matching predicate as parameter.

godtools/Managers/FirstLaunchInitializer.swift
```diff
@@ -15,84 +15,85 @@ class FirstLaunchInitializer: GTDataManager {
...
     
-    private func initializeInitialResources(language: Language?) -> [DownloadedResource?] {
-        let kgp = createEntity(DownloadedResource.self)
-        kgp?.code = "kgp"
-        kgp?.name = "Knowing God Personally"
-        kgp?.remoteId = magicId
-        kgp?.shouldDownload = true
+    private func initializeInitialResources(language: Language?) -> [DownloadedResource] {
+        let kgp = DownloadedResource()
+        kgp.code = "kgp"
```
pablomarti: Just in case, because it is hardcoded, this is never going to change, right?

godtools/Managers/LanguagesManager.swift
```diff
@@ -46,21 +46,16 @@ class LanguagesManager: GTDataManager {
...
-            return language1.localizedName().compare(language2.localizedName()).rawValue < 0
-        }
-        
+    func loadFromDisk() -> List<Language> {
+        languages = findAllEntities(Language.self, sortedByKeyPath: "localizedName")
         return languages
     }
     
-    func loadFromRemote() -> Promise<[Language]> {
+    func loadFromRemote() -> Promise<List<Language>> {
```
pablomarti: Same as before, we can set thinks easier to read if we use the `typealias` for creating something like `Languages`

godtools/Managers/DownloadResources/DownloadedResourceManager.swift
```diff
@@ -52,79 +52,113 @@ class DownloadedResourceManager: GTDataManager {
...
-                
-                cachedAttachment.sha = remoteAttachment.sha256
-                cachedAttachment.resource = cachedResource
+    func download(_ resource: DownloadedResource) {
+        do {
+            try realm.write {
+                resource.shouldDownload = true
+                TranslationZipImporter.shared.download(resource: resource)
             }
+        } catch {
```
pablomarti: This is for later as you said, it would be nice to have something generic for dealing with these, and allow to pass a block for the specific actions to do.

godtools/Managers/DownloadResources/DownloadedResourceManager.swift
```diff
@@ -52,79 +52,113 @@ class DownloadedResourceManager: GTDataManager {
...
-            for remoteTranslationGeneric in remoteTranslations {
-                let remoteTranslation = remoteTranslationGeneric as! TranslationResource
-                let languageId = remoteTranslation.language?.id ?? "-1"
-                let resourceId = remoteResource.id ?? "-1"
-                let version = remoteTranslation.version!.int16Value
+        }
+    }
+    
+    private func saveToDisk(_ resources: [DownloadedResourceJson]) {
+        try! realm.write {
```
pablomarti: This will make the app to crash, right?

godtools/Managers/DownloadResources/DownloadedResourceManager.swift
```diff
@@ -52,79 +52,113 @@ class DownloadedResourceManager: GTDataManager {
...
-            let remoteTranslations = remoteResource.latestTranslations!
-            for remoteTranslationGeneric in remoteTranslations {
-                let remoteTranslation = remoteTranslationGeneric as! TranslationResource
-                let languageId = remoteTranslation.language?.id ?? "-1"
-                let resourceId = remoteResource.id ?? "-1"
-                let version = remoteTranslation.version!.int16Value
+        }
+    }
+    
+    private func saveToDisk(_ resources: [DownloadedResourceJson]) {
```
pablomarti: Is it possible to breakdown this functions in other functions in order to make it easy to follow? (not sure because it is writing to Realm)

godtools/Managers/FirstLaunchInitializer.swift
```diff
@@ -15,84 +15,85 @@ class FirstLaunchInitializer: GTDataManager {
     private let initialPackageCodes = ["kgp", "fourlaws", "sat"]
     
     func initializeAppState() {
-        let language = initializeInitialLanguage()
-        let resources = initializeInitialResources(language: language)
-        initializeInitialTranslations(language: language, resources: resources)
-        
-        saveToDisk()
+        try! realm.write {
```
pablomarti: This will also make the app crash, right? Thats intentional or are you thinking to add a catch?

godtools/Managers/FirstLaunchInitializer.swift
```diff
@@ -15,84 +15,85 @@ class FirstLaunchInitializer: GTDataManager {
...
-        deleteEntities(Language.self, matching: predicate)
-        deleteEntities(DownloadedResource.self, matching: predicate)
-        
-        if GTSettings.shared.primaryLanguageId == magicId {
-            let primaryLanguageCode = Locale.preferredLanguages.first ?? "en"
-            let primaryLanguage = findEntity(Language.self, byAttribute: "code", withValue: primaryLanguageCode) ??
-                findEntity(Language.self, byAttribute: "code", withValue: "en")
-        
-            primaryLanguage?.shouldDownload = true
+        try! realm.write {
```
pablomarti: This one too

godtools/Managers/GTDataManager.swift
```diff
@@ -10,15 +10,15 @@ import Foundation
...
+import RealmSwift
 
 class GTDataManager: NSObject {
     let documentsPath: String
     let resourcesPath: String
     let bannersPath: URL
     
     let serializer = Serializer()
-    let context = NSManagedObjectContext.mr_rootSaving()
+    let realm = try! Realm()
```
pablomarti: Here too

godtools/Managers/LanguagesManager.swift
```diff
@@ -72,32 +67,39 @@ class LanguagesManager: GTDataManager {
         }
     }
     
-    
     func recordLanguageShouldDownload(language: Language) {
-        language.shouldDownload = true
-        saveToDisk()
+        try! realm.write {
```
pablomarti: And here, this will make it crash (I understand is a "do or die")

godtools/Managers/LanguagesManager.swift
```diff
@@ -72,32 +67,39 @@ class LanguagesManager: GTDataManager {
...
+            language.shouldDownload = true
+        }
     }
     
     func recordLanguageShouldDelete(language: Language) {
-        language.shouldDownload = false
-        for translation in language.translationsAsArray() {
-            translation.isDownloaded = false
-            translation.removeFromReferencedFiles(translation.referencedFiles!)
+        try! realm.write {
```
pablomarti: And here may also need a catch

godtools/Managers/LanguagesManager.swift
```diff
@@ -72,32 +67,39 @@ class LanguagesManager: GTDataManager {
...
-        TranslationFileRemover().deleteUnusedPages()
-    
-        saveToDisk()
     }
 
     private func saveToDisk(_ languages: [LanguageResource]) {
-        for remoteLanguage in languages {
-            let cachedlanguage = findFirstOrCreateEntity(Language.self, byAttribute: "remoteId", withValue: remoteLanguage.id!)
-            cachedlanguage.code = remoteLanguage.code
+        try! realm.write {
```
pablomarti: And here

godtools/Managers/TranslationZipImporter.swift
```diff
@@ -192,11 +192,19 @@ class TranslationZipImporter: GTDataManager {
         
         let files = try FileManager.default.contentsOfDirectory(at: directory, includingPropertiesForKeys: nil, options: .skipsHiddenFiles)
         
-        for file in files {
-            let filename = file.lastPathComponent
-            let referencedFile = findEntity(ReferencedFile.self, byAttribute: "filename", withValue: filename) ?? createEntity(ReferencedFile.self)
-            referencedFile?.filename = file.lastPathComponent
-            referencedFile?.addToTranslations(translation)
+        try! realm.write {
```
pablomarti: And here

godtools/Managers/Translations/TranslationsManager.swift
```diff
@@ -10,32 +10,36 @@ import Foundation
...
 import CoreData
+import RealmSwift
 
 class TranslationsManager: GTDataManager {
     static let shared = TranslationsManager()
     
     func translationWasDownloaded(_ translation: Translation) {
-        translation.isDownloaded = true
-        saveToDisk()
+        try! realm.write {
```
pablomarti: And here

godtools/Managers/Translations/TranslationsManager.swift
```diff
@@ -10,32 +10,36 @@ import Foundation
...
                                     NSNumber(booleanLiteral: latest.isDownloaded))
         
-        deleteEntities(Translation.self, matching: predicate)
-        
-        if saving {
-            saveToDisk()
+        if realm.isInWriteTransaction {
+            realm.delete(findEntities(Translation.self, matching: predicate))
+        } else {
+            try! realm.write {
```
pablomarti: And here

godtools/Managers/DownloadResources/DownloadedResourceManager.swift
```diff
@@ -52,79 +52,113 @@ class DownloadedResourceManager: GTDataManager {
...
+        do {
+            try realm.write {
+                resource.shouldDownload = false
+                for translation in resource.translations {
+                    translation.isDownloaded = false
+                }
+                
+                TranslationFileRemover().deleteUnusedPages()
             }
+        } catch {
```
pablomarti: Another place that needs the catch

godtools/Views/Cells/HomeToolTableViewCell.swift
```diff
@@ -76,7 +76,7 @@ class HomeToolTableViewCell: UITableViewCell {
         titleLabel.isEnabled = isAvailableInPrimaryLanguage
         titleLabel.text = resource.localizedName(language: primaryLanguage)
         
-        languageLabel.text = resource.isAvailableInLanguage(parallelLanguage) ? parallelLanguage!.localizedName() : nil
+        languageLabel.text = resource.isAvailableInLanguage(parallelLanguage) ? parallelLanguage!.localizedName : nil
```
pablomarti: Sometimes I think that we can these kind of scenarios, where maybe it would be nice to also decorate, so, instead of making this check here have a function on the class or in the resource, what do you think? (this is something that applies in a lot of places)

godtools/Managers/DownloadResources/DownloadedResourceManager.swift
```diff
@@ -29,18 +29,18 @@ class DownloadedResourceManager: GTDataManager {
...
         resources = findAllEntities(DownloadedResource.self)
         return resources
     }
     
-    func loadFromRemote() -> Promise<[DownloadedResource]> {
+    func loadFromRemote() -> Promise<List<DownloadedResource>> {
         showNetworkingIndicator()
         
-        let params = ["include": "translations,pages,attachments"]
+        let params = ["include": "translations,ages,attachments"]
```
ryancarlson: This is a good catch. We actually don't need to retrieve pages anymore, so this can be removed. I think it makes sense to do in #70 though.

godtools/Managers/DownloadResources/DownloadedResourceManager.swift
ryancarlson: I did exactly this change in #70 :)

godtools/Managers/FirstLaunchInitializer.swift
ryancarlson: Correct. We want to pre-load the app with some options so that the users can still run the app if they cannot connect to the API. So these values have to be hard-coded, but there may be a better way to do it.

godtools/Views/Cells/HomeToolTableViewCell.swift
ryancarlson: I'm not sure I understand, can you explain some more?

godtools/Managers/DownloadResources/DownloadedResourceManager.swift
ryancarlson: I've thought about that, but one downside could be potential retain loops if we aren't careful. Since all of the manager classes share the same underlying find* methods, I think this is okay. Outside the manager classes, then we should use the appropriate manager class.

godtools/Views/Cells/HomeToolTableViewCell.swift
pablomarti: I mean, instead of that line have something like languageLabel.text = `getParallelLanguageName()` which will have all of that, and will return nil if doesn't exist. This is a common thing in the code of all app, don't know what is best

godtools/Managers/GTDataManager.swift
ryancarlson: This is opening the database (and possibly initializing it. The best case scenario here is to show the user an error message. There is no real other behavior I can think of if we cannot access the database.

godtools/Views/Cells/HomeToolTableViewCell.swift
ryancarlson: I think this is more of a specialized case. Here we want to set the parallel language name only if there is a translation for that resource. We could create a method for that, but I don't think that logic is used elsewhere.

ryancarlson:
@pablomarti I responded to most of these items with changes, and left some comments and questions for others. What do you think?

godtools/Managers/GTDataManager.swift
```diff
@@ -43,52 +44,48 @@ class GTDataManager: NSObject {
...
-    }
-    
-    func deleteEntities<T: NSManagedObject>(_ entityClass: T.Type, matching: NSPredicate) {
-        entityClass.mr_deleteAll(matching: matching, in: context)
-    }
-    
-    func deleteEntity<T: NSManagedObject>(_ entity: T) {
-        entity.mr_deleteEntity(in: context)
+
+    func safelyWriteToRealm(_ writeBlock: () -> Void ) {
```
pablomarti: Really great!

godtools/Managers/GTDataManager.swift
pablomarti: 👍

PR #76, opened 2017-06-08
Handle failed downloads

ryancarlson:
- give user option to re-download resources when it appears that a download has failed.

### DIFF
--- a/godtools/ViewControllers/Tract/TractViewControllerContent.swift
+++ b/godtools/ViewControllers/Tract/TractViewControllerContent.swift
@@ -23,7 +23,7 @@ extension TractViewController {
             showErrorMessage()
             return
         }
-        
+    
         for pageNumber in range.start...range.end {
             let view = buildPage(pageNumber, width: width, height: height)
             self.pagesViews[pageNumber] = view
@@ -136,7 +136,9 @@ extension TractViewController {
         
         let actionNo = UIAlertAction(title: "no".localized,
                                       style: .cancel,
-                                      handler: nil)
+                                      handler: { action in
+                                        self.disableResource()
+        })
         
         alert.addAction(actionYes)
         alert.addAction(actionNo)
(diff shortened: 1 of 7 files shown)


ryancarlson:
@pablomarti this should finish one of the crashes ann ran into where it looked like downloads failed. This will pop the user out to the home screen and allow them to re-download the content. If they choose no, the resource will be removed.

I also prevented an ugly race condition i created that caused some crashes.

godtools/ViewControllers/Tract/TractViewControllerContent.swift
```diff
@@ -17,6 +17,13 @@ extension TractViewController {
         cleanContainerView()
         
         let range = getRangeOfViews()
+        
+        if range.end < range.start {
+            baseDelegate?.goHome()
```
pablomarti: Thanks for this!

godtools/ViewControllers/Tract/TractViewControllerContent.swift
```diff
@@ -116,4 +123,40 @@ extension TractViewController {
...
+                                        self.disableResource()
+        })
+        
+        alert.addAction(actionYes)
+        alert.addAction(actionNo)
+        present(alert, animated: true, completion: nil)
+    }
+    
+    private func redownloadResources() {
+        DownloadedResourceManager().delete(resource!)
```
pablomarti: Just in question, I was checking and get lost with this method and the following, is in some place moving back to Home? Because I can't see it, and it should be nice to move the user back to Home.

ryancarlson: Moving back to home happens already on this line: https://github.com/CruGlobal/godtools-swift/pull/76/files#diff-05386bdf576325a006af7b088c952098R22

This method is triggered if the user chooses the "yes" action on the error alert prompt.

pablomarti: Yeah! didn't see that, thanks!

PR #77, opened 2017-06-09
Load properties from manifest and display the backgrounds 

pablomarti:
This PR is a refactoring of the things that process the XML nodes. The objective was to separate properties from the objects that manage the views, and the objective was set due to the need for manage the properties in a hierarchy that can make possible to override values.

This PR adds:
- New file structure with more order in the project
- Separe the logic for processing XML operations
- Separate the logic for the management of the nodes properties (XML properties and views properties)


godtools/TractClasses/ManifestProperties.swift
```diff
@@ -0,0 +1,27 @@
...
+//  ManifestProperties.swift
+//  godtools
+//
+//  Created by Pablo Marti on 6/9/17.
+//  Copyright © 2017 Cru. All rights reserved.
+//
+
+import UIKit
+
+class ManifestProperties: XMLNode {
```
ryancarlson: I see this class is used as an optional throughout many other classes. It seems we should be guaranteed an instance of this class, even if all the properties it contains are not present. It would be better and cleaner, in my opinion if this weren't passed around as an optional and force unwrapped.

godtools/Views/TractElements/TractBackgroundProperties.swift
```diff
@@ -0,0 +1,72 @@
...
+        case "fill-x":
+            self.backgroundImageScaleType = .fillX
+        case "fill-y":
+            self.backgroundImageScaleType = .fillY
+        default:
+            self.backgroundImageScaleType = .fit
+        }
+    }
+    
+    func setupBackgroundImageAlign(_ string: String) {
```
ryancarlson: Can you explain how this will be used?

godtools/XMLManagement/XMLManager.swift
```diff
@@ -0,0 +1,83 @@
...
+
+class XMLManager: NSObject {
+    
+    func getContentElements(_ data: XMLIndexer) -> (elementName: String, namespace: String, kind: String, properties: Dictionary<String, Any>, children: [XMLIndexer]) {
+        let elementName = data.element?.name
+        let elementNameArray = elementName?.components(separatedBy: ":")
+        var namespace = ""
+        var kind = ""
+        
+        if (elementNameArray?.count)! > 1 {
```
ryancarlson: This would be nice as a small function.

godtools/TractClasses/ManifestProperties.swift
pablomarti: 👍

godtools/Views/TractElements/TractBackgroundProperties.swift
pablomarti: Just in case I change a little this code, basically there are some properties that can not be automatically setup, so there is a format for setting those properties

pablomarti:
Hey @ryancarlson sorry this PR has become too long, but I found somethings that I saw that were important to be changed... Would you please take a look to see how it is going? I'm almost done, just want to see how to completely move some properties, and I have to display properly the backgrounds (all of this is weird but is related to the backgrounds, on how they come and the hierarchy, that make me go like this)

PR #82, opened 2017-06-20
Set render font sizes

ryancarlson:


godtools/Views/TractElements/TractTextContentProperties.swift
```diff
@@ -15,11 +15,11 @@ class TractTextContentProperties: TractProperties {
     var i18nId: String = ""
     var textAlign: NSTextAlignment = .left
     // textColor
-    var textScale: CGFloat = 1.0
+    // textScale
     var value: String = ""
     
     override func defineProperties() {
-        self.properties = ["i18nId", "textScale", "value"]
```
ryancarlson: @pablomarti  so does this mean we can't scale individual text elements any more? we need to be able to do that.

godtools/Views/TractElements/TractHeaderProperties.swift
```diff
@@ -18,6 +18,11 @@ class TractHeaderProperties: TractProperties {
         self.properties = ["backgroundColor"]
     }
     
+    override func setupDefaultProperties() {
+        self.textScale = 3
```
ryancarlson: @pablomarti the header actually is not scaled by default. it should be the same size as default text (18). it's the header element in a hero that should be 30 by default.

godtools/TractClasses/TractProperties.swift
```diff
@@ -13,17 +13,21 @@ class TractProperties: XMLNode {
     var primaryColor = GTAppDefaultStyle.primaryColor.getRGBAColor()
     var primaryTextColor = GTAppDefaultStyle.primaryTextColorString.getRGBAColor()
     var textColor = GTAppDefaultStyle.textColorString.getRGBAColor()
+    var textScale: Int = 1
```
ryancarlson: Scale should be a float, not an int.

godtools/TractClasses/TractProperties.swift
```diff
@@ -13,17 +13,21 @@ class TractProperties: XMLNode {
...
+    var textSize: Int = 18
     
     required override init() {
         super.init()
     }
     
     override func getProperties() -> [String] {
         defineProperties()
-        return ["primaryColor", "primaryTextColor", "textColor"] + self.properties
+        return ["primaryColor", "primaryTextColor", "textColor", "textScale", "textSize"] + self.properties
```
ryancarlson: we should never expect a textSize to be set as a property. we define the defaults and they can be scaled.

godtools/Views/TractElements/TractTextContentProperties.swift
pablomarti: No, the text-scale will always be used, but now it is defined as a general property, not just for texts, because it can be set on many places

godtools/Views/TractElements/TractHeaderProperties.swift
pablomarti: Ok will fix this

godtools/TractClasses/TractProperties.swift
pablomarti: Ok

godtools/TractClasses/TractProperties.swift
pablomarti: Ok understood, will fix it

PR #83, opened 2017-06-20
Font size & color tweaks

ryancarlson:
- make colors inherited from the top level downward (manifest, page, text:content)
- a local text color should override anything set above it
- font scaling support works now too, only on lowest level content:text node
- default text size = 18, except for headings, now 30


ryancarlson:
@pablomarti this should handle rendering fonts with the right size and color in most cases. I still need to handle card-text-color, but that should be pretty easy

PR #84, opened 2017-06-21
Card text color

ryancarlson:
Allows for cardTextColor to be set at the page level to override textColor set for the rest of the page or manifest.

Some other small cleanup too


godtools/XMLManagement/XMLManager.swift
```diff
@@ -48,26 +48,31 @@ class XMLManager: NSObject {
             return
         }
         
-        if !object.getProperties().contains(propertyName) {
+        var objectMirror: Mirror? = Mirror(reflecting: object)
```
pablomarti: I really love this

PR #86, opened 2017-06-21
Fix and update nav bar issues

ryancarlson:
- fix typo that prevented color from being loaded from properties 'B' -> 'b'
- use machine generated images with alpha for the background image of nav bars to achieve transparency


godtools/TractClasses/ManifestProperties.swift
```diff
@@ -14,13 +14,13 @@ class ManifestProperties: TractProperties {
     var backgroundImage: String = ""
     var backgroundImageAlign: [TractImageConfig.ImageAlign] = [.center]
     var backgroundImageScaleType: TractImageConfig.ImageScaleType = .fill
-    var navBarColor = GTAppDefaultStyle.navBarColor.getRGBAColor()
-    var navBarControlColor = GTAppDefaultStyle.navBarControlColor.getRGBAColor()
+    var navbarColor = GTAppDefaultStyle.navBarColor.getRGBAColor()
```
pablomarti: Good catch

PR #88, opened 2017-06-22
Take the fact that the view slides behind the nav bar when moving cards.

ryancarlson:
- cards should not slide behind nav bar.

This means cards starting yPos should be adjusted by the nav bar height, but the height should also be reduced by the nav bar height
The internal heights also need adjusted so that the scrolls work well


ryancarlson:
@pablomarti I'm not sure this is the best way to fix the issue, but it seems work decently well in the simulator. i'm going to test on my phone. what do you think?

pablomarti:
It looks good but a side effect is that when opening a card, the other cards at the bottom totally disappear (I'm testing in the 5S simulator)

ryancarlson:
Approved Pablo's changes!

PR #95, opened 2017-06-23
Fix issues when deleting a resource or a translation

ryancarlson:
The referenced file counts were being done incorrectly. This led to deleting one language or resource deleting files that belong to something else. This could mean images missing or even crashes when XML pages were missing.

### DIFF
--- a/godtools/AppDelegate.swift
+++ b/godtools/AppDelegate.swift
@@ -27,7 +27,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                 UIApplication.shared.isNetworkActivityIndicatorVisible = false
         }
         
-        print(NSHomeDirectory())
+        #if DEBUG
+            print(NSHomeDirectory())
+        #endif
         
         return true
     }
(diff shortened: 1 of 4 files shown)


godtools/AppDelegate.swift
```diff
@@ -27,6 +27,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                 UIApplication.shared.isNetworkActivityIndicatorVisible = false
         }
         
+        print(NSHomeDirectory())
```
pablomarti: Please remove this print

godtools/Managers/TranslationZipImporter.swift
```diff
@@ -188,12 +188,13 @@ class TranslationZipImporter: GTDataManager {
         
         let files = try FileManager.default.contentsOfDirectory(at: directory, includingPropertiesForKeys: nil, options: .skipsHiddenFiles)
         
+        print(files.count)
```
pablomarti: Please remove this

godtools/AppDelegate.swift
ryancarlson: I actually want to leave this one in there for awhile. It's the quickest and easiest way i've found to be able to determine the disk location for the database and resources folder. It shouldn't hurt anything, but we can leave it in for launch.

I could also wrap it in a debug tag so it only runs on debug builds, what do you think?

godtools/Managers/TranslationZipImporter.swift
ryancarlson: Good catch.

PR #97, opened 2017-06-23
Persist card state between languages

pablomarti:


pablomarti:
Hey @ryancarlson would you please check this PR? it is working fine for me, except for the lag of 0.7 segs, but it was before also. I believe the solution of creating a parallel view so when changing the language it will change the view immediately is a good option but it should be done in another branch.

godtools/Views/TractElements/TractCards.swift
```diff
@@ -38,6 +29,13 @@ class TractCards: BaseTractElement {
         return self.height - self.elementFrame.y + TractPage.navbarHeight
     }
     
+    // MARK: - Dynamic settings
+    
+    var isOnInitialPosition: Bool = true
```
ryancarlson: The type `Bool` can be inferred here. Very very small thing :)

pablomarti: Just made the change, thanks!

PR #105, opened 2017-06-27
Load type of input, build hidden input, and display keyboard depending on type

pablomarti:


godtools/GeneralExtensions/String.swift
```diff
@@ -15,6 +15,17 @@ extension String {
         return NSLocalizedString(self, tableName: nil, bundle: Bundle.main, value: "", comment: "")
     }
     
+    func removeBreaklines() -> String {
+        let regex = try! NSRegularExpression(pattern: "\n", options: NSRegularExpression.Options.caseInsensitive)
```
ryancarlson: I'm guessing we don't need the class/struct names `NSRegularExpression.Options` here? I'm not sure though.

godtools/GeneralExtensions/String.swift
```diff
@@ -15,6 +15,17 @@ extension String {
...
+        let range = NSMakeRange(0, self.characters.count)
+        return regex.stringByReplacingMatches(in: self, options: [], range: range, withTemplate: "")
+    }
+    
+    func condenseWhitespaces() -> String {
+        let components = self.components(separatedBy: NSCharacterSet.whitespacesAndNewlines)
```
ryancarlson: Do we need `NSCharacterSet` here?

godtools/Views/TractElements/TractInput.swift
```diff
@@ -51,17 +51,30 @@ class TractInput: BaseTractElement {
...
         self.textField.backgroundColor = elementProperties.backgroundColor
         self.textField.placeholderTranslationKey = elementProperties.placeholder ?? ""
+        self.textField.text = elementProperties.inputValue()
         self.textField.frame = CGRect(x: 0.0,
                                       y: self.height,
                                       width: self.elementFrame.finalWidth(),
                                       height: elementProperties.height)
         
+        switch elementProperties.type {
+        case .email:
```
ryancarlson: 👍 !

godtools/GeneralExtensions/String.swift
pablomarti: Not needed :D

godtools/GeneralExtensions/String.swift
pablomarti: It is not needed too, just making the changes, thanks!

PR #113, opened 2017-06-29
Improvements to follow up subscriptions

ryancarlson:
- post data in JSON API format using Spine resource
- add form validation for required fields
- cache failed subscriptions on device for retry later
- show an alert and don't call next event if validation fails (weak reference used!)
- implement a mechanism where listeners can return a status: `success`, `failure`, `viewNotFound` so that event caller can handle the result


ryancarlson:
@pablomarti can you make sure i didn't introduce a retain cycle with the delegate i added here?

godtools/Managers/GTDataManager.swift
```diff
@@ -37,12 +41,19 @@ class GTDataManager: NSObject {
...
     }
-    
+        
     func issuePOSTRequest(_ params: Parameters) -> Promise<Data> {
         return Alamofire
             .request(buildURL() ?? "",
                      method: HTTPMethod.post,
                      parameters: params)
+            .validate({ (request, response, data) -> Request.ValidationResult in
+                if response.statusCode / 100 != 2 {
```
pablomarti: Hey why didn't you compare directly to 200?

godtools/Views/TractElements/BaseTractElement.swift
```diff
@@ -119,6 +124,8 @@ class BaseTractElement: UIView {
             }
         }
     }
+    
+    weak private var delegate: BaseTractElementDelegate?
```
pablomarti: 👍

godtools/ViewControllers/Tract/TractViewController.swift
```diff
@@ -152,14 +155,13 @@ class TractViewController: BaseViewController {
...
+    fileprivate func configureLanguageSegmentedControl() {
         let primaryLabel = self.determinePrimaryLabel()
         let parallelLabel = self.determineParallelLabel()
         
-        let segmentedControl = UISegmentedControl(items: [primaryLabel, parallelLabel])
-        segmentedControl.selectedSegmentIndex = 0
-        segmentedControl.addTarget(self, action: #selector(didSelectLanguage), for: .valueChanged)
-        return segmentedControl
+        languageSegmentedControl = UISegmentedControl(items: [primaryLabel, parallelLabel])
+        languageSegmentedControl!.selectedSegmentIndex = 0
```
pablomarti: The app is crashing in this line because `languageSegmentedControl` is nil, so it is not retaining the class, the `weak` is causing this (don't know why, it should not happen), but, do we really need the `weak` for this?

godtools/ViewControllers/Tract/TractViewController.swift
```diff
@@ -79,7 +81,8 @@ class TractViewController: BaseViewController {
 
     override func displayScreenTitle() {
         if parallelLanguageIsAvailable() && determinePrimaryLabel() != determineParallelLabel() {
-            self.navigationItem.titleView = languageSegmentedControl()
+            configureLanguageSegmentedControl()
+            self.navigationItem.titleView = languageSegmentedControl
```
pablomarti: Is because of this that you are setting `languageSegmentedControl` as weak?

godtools/Views/TractElements/TractForm+Actions.swift
```diff
@@ -42,7 +43,65 @@ extension TractForm {
...
+        data["language_id"] = getDelegate()?.displayedLanguage().remoteId
+        
         return data
     }
     
+    func validateForm() -> Bool {
+        var validationErrors = [String]()
+        
+        for element in formElements {
+            let input = element as! TractInput
```
ryancarlson: good idea!

godtools/ViewControllers/Tract/TractViewController.swift
ryancarlson: No, it's so that I can easily access the segmented control in the delegate method. I would have to go through the navigation controller to the navigation bar and to the title view, and then get its subview and cast it so a segmented control to check the value of it. Far easier to just hold a reference :)

godtools/ViewControllers/Tract/TractViewController.swift
ryancarlson: The weak that i put there was really bad actually. since that was a weak reference, it could get dropped immediately after creating the object before it gets added as a subview. 👎

ryancarlson:
hey @pablomarti i think i've addressed your concerns here. what do you think?

godtools/Managers/GTDataManager.swift
pablomarti: Hey cool! good one

pablomarti:
Great @ryancarlson !

PR #118, opened 2017-06-30
Fix cards problem when text of hero component is large and there are many cards

pablomarti:
On iPhone 4 and 5, currently, if the text of the `<hero>` is large and there are too many `<card>` inside of `<cards>`, the first `<card>` will not be able to have a touch gesture because the `<cards>` element gets its position down, and those `<card>` objects have a -y position (this problem may also happen in iPhone 6 and 7 if the problem with the heights overlaps).

The solution is that the position y of the `<cards>` object should be updated taking the new height of the `<hero>` component (that new height is calculated after the `<cards>` object was added, so, this required operations between elements of the same level).

### DIFF
--- a/godtools/Views/TractElements/BaseTractElement.swift
+++ b/godtools/Views/TractElements/BaseTractElement.swift
@@ -197,12 +197,12 @@ class BaseTractElement: UIView {
     
     func getPreviousElement() -> BaseTractElement? {
         guard let index = getElementPosition() else {
-            return nil
+            return self.elements?.last
         }
         if index > 0 {
             return self.parent!.elements?[index - 1]
         }
-        return nil
+        return self.elements?.last
     }
     
     func getFollowingElement() -> BaseTractElement? {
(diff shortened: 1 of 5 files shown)


pablomarti:
This is almost done, but it requires a refactoring, and a problem, sometimes the <hero> height is too small.

godtools/Views/TractElements/BaseTractElement.swift
```diff
@@ -197,12 +197,12 @@ class BaseTractElement: UIView {
     
     func getPreviousElement() -> BaseTractElement? {
         guard let index = getElementPosition() else {
-            return nil
+            return self.elements?.last
```
ryancarlson: Can you explain what is happening here and on line 205? I don't understand.

godtools/Views/TractElements/BaseTractElement.swift
```diff
@@ -197,12 +197,12 @@ class BaseTractElement: UIView {
     
     func getPreviousElement() -> BaseTractElement? {
         guard let index = getElementPosition() else {
-            return self.elements?.last
+            return self.parent?.elements?.last
```
ryancarlson: Same question here after update.

godtools/Views/TractElements/TractHero.swift
```diff
@@ -64,7 +64,6 @@ class TractHero: BaseTractElement {
         self.scrollView.contentSize = CGSize(width: contentWidth, height: contentHeight)
         self.scrollView.frame = CGRect(x: 0.0, y: 0.0, width: contentWidth, height: self.heroHeight)
         self.scrollView.showsVerticalScrollIndicator = false
-        self.scrollView.backgroundColor = .red
```
ryancarlson: 👍

godtools/Views/TractElements/BaseTractElement.swift
pablomarti: Previously, if it was impossible to get the element position (because the list was empty or the element wasn't added to the list yet) it returned nil, but, because it is currently processing the element (so the position is not set yet but will be set when the element is processed it gets the last element of the list (which will be the previous one.

godtools/Views/TractElements/BaseTractElement.swift
pablomarti: The answer is the same, but, there was a logical/dumb problem before, it was trying to get the previous element on the childrens elements (really dumb, sorry...) so now instead it is searching on the parent elements

PR #122, opened 2017-07-04
Disabled download button

ryancarlson:
When a resource exists, but has no published translations: 
 - don't show the download button on the add tools table view
 - show a disabled "unavailable" button on the tool detail view.

### DIFF
--- a/godtools/Views/Buttons/GTButton.swift
+++ b/godtools/Views/Buttons/GTButton.swift
@@ -47,17 +47,21 @@ class GTButton: UIButton {
     
     @IBInspectable var translationKey: String = "" {
         didSet {
-            self.setTitle(translationKey.localized, for: .normal)
+            let state = self.isEnabled ? UIControlState.normal : UIControlState.disabled
+            
+            self.setTitle(translationKey.localized, for: state)
         }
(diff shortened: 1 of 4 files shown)


godtools/Views/Buttons/GTButton.swift
```diff
@@ -47,10 +47,28 @@ class GTButton: UIButton {
...
+        
+        self.color = .gtWhite
+        self.tintColor = .gtWhite
+        self.borderWidth = 0.0
+        self.borderColor = .clear
+        self.backgroundColor = .gtGrey
+        self.setImage(#imageLiteral(resourceName: "download_white"), for: .normal)
+        self.translationKey = "unavailable"
+        
+        self.isEnabled = false
```
pablomarti: I believe this line should be before self.translationKey = "unavailable", because of the `didSet` of the translationKey

pablomarti: Nevermind @ryancarlson I switch my phone to spanish for testing the previous PR, and this text isn't translated yet, it is working properly 👍

ryancarlson:
Gracias!

PR #123, opened 2017-07-04
Add basic GoogleAnalytics to to project

ryancarlson:
Initially we will track a handful of screens to make sure we get the basic number of users, sessions, rough location etc.

The mechanism is for each VC to override the func `screenName()` to provide its screen name to the analytics handler.


ryancarlson:
@michaelharro would approve of this, I bet ;)

godtools/ViewControllers/BaseViewController.swift
```diff
@@ -183,4 +184,18 @@ class BaseViewController: UIViewController {
...
+        let userInfo = [GTConstants.kAnalyticsScreenNameKey: screenName()]
+        
+        NotificationCenter.default.post(name: .screenViewNotification,
+                                        object: nil,
+                                        userInfo: userInfo)
+    }
+    func screenName() -> String {
+//        assertionFailure("method must be overriden by view controller")
+        
+        return "unknown"
```
pablomarti: Everything looks good, just this, but will this be updated later?

ryancarlson: Yes, I wanted to confer with @frett to make sure we have the same naming pattern for our screens on both platforms.

ryancarlson:
@pablomarti is this okay to merge so that other PRs can move forward? I will create an issue to add other screen names to the analytics tracking.

PR #125, opened 2017-07-04
GT-362 Sync Cached follow ups

ryancarlson:
Create a function that will sync cached follow ups, attempting 2 more times before giving up.

This method is not called yet, b/c I will call it from the AppDelegate. There are at least two other branches that need merged into the delegate, and it will be conflict-city with if I do it now.


ryancarlson:
Cool. I will update that once #123 is merged. I think would introduce conflicts.

ryancarlson:
@pablomarti I added the line to app delegate to run the job on app launch. I tested this yesterday. Good to go?

PR #129, opened 2017-07-06
GT-367: Add screen names for analytics 

pablomarti:


pablomarti:
It is almost done, currently it needs:
- Screen name for "Terms of use" (there is no screen yet)
- Set the id of the tract in the name of the screen name when sending (currently the manifest is not sending an id)

ryancarlson:
@pablomarti just use the index of the view. we don't need the id or actual filename

godtools/ViewControllers/Platform/WebViewController.swift
```diff
@@ -28,5 +29,11 @@ class WebViewController: BaseViewController {
             return self.pageTitle?.localized ?? ""
         }
     }
+    
+    // MARK: - Analytics
+    
+    override func screenName() -> String {
+        return self.pageTitleForAnalytics ?? ""
```
ryancarlson: let's do "unknown" here instead of ""

godtools/ViewControllers/Tract/TractViewControllerActions.swift
```diff
@@ -42,4 +42,9 @@ extension TractViewController: MFMailComposeViewControllerDelegate {
         self.present(alert, animated: true, completion: nil)
     }
     
+    func sendPageToAnalytics() {
+        let screenName = "\(self.currentPage)"
```
ryancarlson: It should be {resource.code}-{self.currentPage}. example "kgp-2" "4sl-6"

godtools/ViewControllers/MenuViewController.swift
```diff
@@ -210,41 +216,43 @@ extension MenuViewController {
...
     fileprivate func shareGodToolsApp() {
         let textToShare = [ "share_god_tools_share_sheet_text".localized ]
         let activityViewController = UIActivityViewController(activityItems: textToShare, applicationActivities: nil)
         self.present(activityViewController, animated: true, completion: nil)
+        sendScreenViewNotification(screenName: "Share App")
     }
     
     fileprivate func shareAStoryWithUs() {
         if MFMailComposeViewController.canSendMail() {
+            sendScreenViewNotification(screenName: "Share Story")
```
ryancarlson: I would rather track this as an event later since it's not truly a screen within the app.

godtools/ViewControllers/MenuViewController.swift
```diff
@@ -210,41 +216,43 @@ extension MenuViewController {
...
-            self.delegate?.openWebView(url: url!, title: "contact_us".localized)
+            self.delegate?.openWebView(url: url!, title: "contact_us".localized, analyticsTitle: "Contact Us")
         }
     }
     
     fileprivate func shareGodToolsApp() {
         let textToShare = [ "share_god_tools_share_sheet_text".localized ]
         let activityViewController = UIActivityViewController(activityItems: textToShare, applicationActivities: nil)
         self.present(activityViewController, animated: true, completion: nil)
+        sendScreenViewNotification(screenName: "Share App")
```
ryancarlson: I would rather track this as an event later since it's not truly a screen within the app.

pablomarti:
Thanks @ryancarlson !

PR #130, opened 2017-07-06
GT-354: Fix margins on texts

pablomarti:


pablomarti:
Thanks @ryancarlson !

PR #134, opened 2017-07-07
The values in OneSky do not have a protocol. 

ryancarlson:
So I added a method to add http:// to those that don't have one.

Without a protocol, the call to: UIApplication.shared.openURL(url) will do nothing


ryancarlson:
@pablomarti here is a small fix i had to make after using the one sky localized URLs. Those values in Onesky do not have a protocol, so the URL launch fails. Before your API updated it just used hard-coded URLs where I did put in the protocol

PR #135, opened 2017-07-07
GT-372: Add share links to share Tract

pablomarti:


godtools/ViewControllers/Tract/TractViewController.swift
```diff
@@ -190,7 +194,13 @@ class TractViewController: BaseViewController {
     }
     
     override func shareButtonAction() {
-        let activityController = UIActivityViewController(activityItems: [String.localizedStringWithFormat("tract_share_message".localized, "www.knowgod.com")], applicationActivities: nil)
+        let languageCode = self.selectedLanguage?.code ?? "en"
```
ryancarlson: I like the way you did this here! 👍

godtools/ViewControllers/Tract/TractViewController.swift
```diff
@@ -84,7 +85,10 @@ class TractViewController: BaseViewController {
     
     override func configureNavigationButtons() {
         self.addHomeButton()
-        self.addShareButton()
+        
+        if self.resource?.code != "kgp-us" {
```
ryancarlson: Can you create a class property that contains a list of items we should ignore, and then just do a lookup on that array instead of checking one value? That will make it easier to add remove things as new tools get added to the app or the website.

godtools/ViewControllers/Tract/TractViewController.swift
pablomarti: Thx!

pablomarti:
Hey @ryancarlson I just uploaded the change

PR #136, opened 2017-07-07
Show onboarding screens on first launch

ryancarlson:
This was broken when logic was added to preset the primary language on first launch.


ryancarlson:
@pablomarti another small bug fix for something i broke.

ryancarlson:
thanks!

PR #137, opened 2017-07-07
Delete files XML files and images that may be on the disk from a prev…

ryancarlson:
…ious version.

Leaving the old database for now in case we need it for something


ryancarlson:
@pablomarti i'm deleting the old XMLs and images here when we finish launching the first time to free up disk space on the user's device.

PR #138, opened 2017-07-07
Button margin

ryancarlson:


ryancarlson:
@pablomarti i also made a small change to paragraph margins.. what do you think?

pablomarti:
Looks good @ryancarlson !

ryancarlson:
gracias mi amigo!

PR #139, opened 2017-07-07
App icons

pablomarti:


pablomarti:
Will remake this because of a problem with the provisioning profile

PR #140, opened 2017-07-07
Launch screen

pablomarti:


pablomarti:
For me this is ok @ryancarlson, what do you think?

PR #143, opened 2017-07-08
Prevent crash when navigation controller is nil.

ryancarlson:
Not totally sure the cases that cause this failure, but it's been reported in crashlytics a couple times.

https://fabric.io/cru/ios/apps/org.cru.godtools.beta/issues/5956b1f0be077a4dccd947b4?time=last-seven-days


ryancarlson:
hey @pablomarti on Monday can you look at this? I've seen this one or two times in Crashlytics.

PR #144, opened 2017-07-10
Continue to next language instead of returning from function.

ryancarlson:
I think I was thinking that this would just return out from the block, but I'm not sure. At any rate, this prevents new languages from being saved, which is an insta-crash whenever something new is published.


ryancarlson:
hey @pablomarti this crashes now that i've added a bunch of new translations. Can you approve this so I can ship a new build?

PR #146, opened 2017-07-10
Somehow languages are not being saved before resources are inspected.

ryancarlson:
- Fixing that bug by ensuring they are there. An early return here isn't the end of the world.


ryancarlson:
@pablomarti I encountered another crash in the same location. I don't understand this one, as I don't see a reason why the languages wouldn't be saved now.. unless the API wasn't returning them.

The solution is to just abort the check to delete older translations. It's not a big deal if they stick around a little while too long. They will eventually be cleaned up.

ryancarlson:
closed in favor of #148

PR #151, opened 2017-07-12
Fix button text when text is longer than buttons width

pablomarti:


pablomarti:
Hey @ryancarlson it looks good, it goes to the 2nd line without problem

PR #152, opened 2017-07-18
Delete languages locally when not found on remote server

ryancarlson:


ryancarlson:
@pablomarti could you review this please?

PR #153, opened 2017-07-19
GT-399 Add font to support Amharic characters

pablomarti:


pablomarti:
Hey @ryancarlson I have been trying to test this but I can't find a way to change the lang on the available languages neither on the system configs for Amharic (Ethiopic) lang. Would you please help me with this? thanks!

godtools/GeneralExtensions/UIFont.swift
```diff
@@ -0,0 +1,57 @@
...
+    
+    static func gtThin(size: CGFloat) -> UIFont {
+        return UIFont.systemFont(ofSize: size, weight: UIFontWeightThin)
+    }
+    
+    static func gtSemiBold(size: CGFloat) -> UIFont {
+        return UIFont.systemFont(ofSize: size, weight: UIFontWeightSemibold)
+    }
+    
+    func transformToAppropriateFont() -> UIFont {
```
ryancarlson: Can you make this function private and call this function inside the functions above instead of throughout the various tract classes? If that works, it's better because we will not have to repeat this function call whenever a new UI element is added.

pablomarti: Cool!

pablomarti:
Hey @ryancarlson just made the change, really good catch! It will affect also other elements because GTLabel always use one of those fonts. What do you think? I still can't find a way for testing this...

godtools/GeneralExtensions/UIFont.swift
```diff
@@ -11,30 +11,61 @@ import UIKit
...
-        return transformToAppropriateFontByLanguage(language!, textScale: 1.0)
+        if language.code == "am-ET" {
+            if weight == UIFontWeightSemibold || weight == UIFontWeightBold {
+                fontName = "NotoSansEthiopic-Bold"
+            } else {
+                fontName = "NotoSansEthiopic"
+            }
+        }
+        
+        if fontName == "" {
```
ryancarlson: Instead of setting the variable and checking it here, just return the appropriate font inside the if/else blocks above.

godtools/GeneralExtensions/UIFont.swift
```diff
@@ -11,30 +11,61 @@ import UIKit
...
+        guard let language = LanguagesManager.defaultLanguage else {
+            return UIFont.systemFont(ofSize: size, weight: weight)
         }
         
-        let languagesManager = LanguagesManager()
-        let language = languagesManager.loadPrimaryLanguageFromDisk()
+        var fontName = ""
         
-        return transformToAppropriateFontByLanguage(language!, textScale: 1.0)
+        if language.code == "am-ET" {
```
ryancarlson: Let's do `if language.code != "am"` as the primary condition. Most of the time it will be true this way and we can just return the system font as normal.

ryancarlson: Also we will use `am` as the language code in the new system, not `am-ET`

godtools/Managers/LanguagesManager.swift
```diff
@@ -19,6 +19,21 @@ class LanguagesManager: GTDataManager {
     
     var selectingPrimaryLanguage = true
     
+    static var _defaultLanguage: Language?
+    static var defaultLanguage: Language? {
```
ryancarlson: I see that this gets set when the getter is called for the first time. How does it ever get changed though? I never see the setter called.

pablomarti: I will add the setter when the primary lang is changed

godtools/GeneralExtensions/UIFont.swift
```diff
@@ -0,0 +1,80 @@
...
+        
+        return UIFont.systemFont(ofSize: size, weight: weight)
+    }
+    
+    static func shouldTransformLanguage() -> Bool {
+        guard let language = LanguagesManager.defaultLanguage else {
+            return false
+        }
+        
+        if language.code == "am-ET" {
```
ryancarlson: This should be "am".

godtools/GeneralExtensions/UIFont.swift
```diff
@@ -0,0 +1,80 @@
...
+        } else {
+            return false
+        }
+    }
+    
+    func transformToAppropriateFontByLanguage(_ language: Language, textScale: CGFloat = 1.0) -> UIFont {
+        var fontSize = self.pointSize
+        var fontName = self.fontName
+        
+        if language.code == "am-ET" {
```
ryancarlson: This should be "am" as well

godtools/Managers/LanguagesManager.swift
```diff
@@ -144,15 +159,16 @@ class LanguagesManager: GTDataManager {
         }
     }
     
-    func setSelectedLanguageId(_ id: String) {
+    func setSelectedLanguageId(_ language: Language) {
```
ryancarlson: Let's change the name of the method to `setSelectedLanguage` since we are passing in a `Language` instead of a language ID

PR #155, opened 2017-07-20
Release 5.0.1

ryancarlson:
Fixes:
 - allows languages to be deleted when delete from remote server
 - longer button titles wrap to multiple lines.


ryancarlson:
@pablomarti here is that release.

PR #157, opened 2017-07-20
Remove not needed target

pablomarti:


pablomarti:
Problem fixed @ryancarlson, my bad... The extension file that I created was with a target membership for Test, and the other files were not on that target (actually almost every file is not with that target) so I removed it

PR #158, opened 2017-07-31
Analytics name fix

ryancarlson:
- unwrap optional resource when reporting name
- report the first view "0" when a tract is first loaded


ryancarlson:
@pablomarti can you review this small PR please? just a couple things I noticed from looking through the analytics.

PR #160, opened 2017-08-01
Update URLs to match where things are actually hosted on the new godt…

ryancarlson:
…oolsapp.com site.

### DIFF
--- a/godtools/ViewControllers/MenuViewController.swift
+++ b/godtools/ViewControllers/MenuViewController.swift
@@ -224,7 +224,7 @@ extension MenuViewController {
         if MFMailComposeViewController.canSendMail() {
             sendEmail(recipient: "support@godtoolsapp.com", subject: "Email to GodTools support")
         } else {
-            let url = URL(string: "http://www.godtoolsapp.com/contact.html")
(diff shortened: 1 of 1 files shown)


ryancarlson:
This is an older commit I think I left open b/c I thought it possible that the website would be updated. It hasn't, so I think this is worth merging now. @frett ?

godtools/ViewControllers/MenuViewController.swift
```diff
@@ -239,13 +239,13 @@ extension MenuViewController {
...
         } else {
-            let url = URL(string: "http://www.godtoolsapp.com/contact.html")
+            let url = URL(string: "http://www.godtoolsapp.com/#contact")
             self.delegate?.openWebView(url: url!, title: "share_a_story_with_us".localized, analyticsTitle: "Share Story")
         }
     }
     
     fileprivate func openTermsOfUse() {
-        let url = URL(string: "http://www.godtoolsapp.com/terms")
+        let url = URL(string: "http://www.godtoolsapp.com/user-agreement/")
```
frett: I should update this in Android as well

PR #163, opened 2017-08-09
Fix screen name reporting for first view of tract controller

ryancarlson:
It was being reported twice b/c the code was trigger both in:
1. TractViewController.viewDidLoad() - this one worked properly because the method it called built the view name "resource"-"pageNum"
2. BaseViewController.viewWillAppear() - this one did not work, because TractViewController did not override the screenName() method with the logic to build the name. It fell back to "unknown" as defined in BaseViewController.screenName().

This fix eliminates the call in viewDidLoad() and just allows the call in BaseViewController.viewWillAppear() to be triggered. It also moves the logic to build the view name to the overridden screenName() method that is used for all views of a tract now. the viewWillAppear() call now uses the screenName() logic to build the right name.

### DIFF
--- a/godtools/ViewControllers/Tract/TractViewController.swift
+++ b/godtools/ViewControllers/Tract/TractViewController.swift
@@ -49,7 +49,6 @@ class TractViewController: BaseViewController {
         getResourceData()
         setupSwipeGestures()
         defineObservers()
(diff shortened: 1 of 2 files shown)


ryancarlson:
thanks!

PR #169, opened 2017-08-16
Re-add share link to us KGP

ryancarlson:
it will default to the first page of the KGP on knowgod.com


ryancarlson:
@pablomarti can you review this please?

PR #170, opened 2017-08-16
Gt 420 english resources

ryancarlson:
show english resources in tract controller when the resource is not translated into the primary language.

this includes some refactoring of how languages are referenced from the `TractViewController`


ryancarlson:
hey @pablomarti can you review this one too?

PR #172, opened 2017-10-20
Gt 428 Update Adobe Analytics

ryancarlson:
- import Adobe Analytics as a cocoapod instead of a library
- use different config files for debug and release builds
- change Adobe app ID in both config files
- remove `GodTools :` prefix in screen name reporting
- minor compatibility changes required to build in XCode9
   - RealmSwift and XML parsing library


godtools/Views/TractElements/TractTabs+ChildSetup.swift
```diff
@@ -19,7 +19,7 @@ extension TractTabs {
             for node in item.children {
                 if self.xmlManager.parser.nodeIsLabel(node: node) {
                     if let textNode = self.xmlManager.parser.getTextContentFromElement(node) {
-                        self.tabsProperties().options.append(textNode.text!)
+                        self.tabsProperties().options.append(textNode.text)
```
pablomarti: Why was the `!` removed?

godtools/XMLManagement/XMLManager.swift
```diff
@@ -26,7 +26,7 @@ class XMLManager: NSObject {
             properties[attribute.name] = attribute.text
         }
         
-        if data.element?.text != nil && data.element?.text?.trimmingCharacters(in: .whitespaces) != "" {
+        if data.element?.text != nil && data.element?.text.trimmingCharacters(in: .whitespaces) != "" {
```
pablomarti: Why was the `?` removed?

godtools/Views/TractElements/TractTabs+ChildSetup.swift
ryancarlson: This was a required as part of building in XCode 9. I'm not sure why, but the build would not pass w/o this change.

godtools/XMLManagement/XMLManager.swift
ryancarlson: Same as the other comment. This was a required as part of building in XCode 9. I'm not sure why, but the build would not pass w/o this change.

ryancarlson:
Thanks! Yeah those two changes showed up as build errors in XCode 9, but worked in XCode 8. I had to make those changes to get things working in the new environment, but I did not dig in deeply to see why.

godtools/XMLManagement/XMLManager.swift
pablomarti: Ok, thanks!

PR #173, opened 2017-10-20
Fix rendering of navigation buttons in XCode 9

ryancarlson:
The issue actually isn't specific to the "right side" buttons, though the issue was noticed there.

The reason is because the buttons that are shown on the right side of the bar have images that are signficantly larger than the left side.

This version of XCode that builds the project doesn't respect the size applied to UIBarButton item and allows the image (in customView) to expand as large as it wants to. The old builder enforced the limit at the UIBarButtonItem level

The actual fix is to apply height and width constraints to the buttons inside the UIBarButtonItem to make sure they stay 22x22 regardless of the size of the image they contain.

### DIFF
--- a/godtools/Assets.xcassets/AppIcon.appiconset/Contents.json
+++ b/godtools/Assets.xcassets/AppIcon.appiconset/Contents.json
@@ -47,6 +47,11 @@
       "idiom" : "iphone",
       "filename" : "device-icon-180.png",
       "scale" : "3x"
+    },
+    {
+      "idiom" : "ios-marketing",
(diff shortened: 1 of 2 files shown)


ryancarlson:
Hey @pablomarti here is one more small PR that came up now that I'm building in XCode 9. It was the issue, maybe you saw it, where the navigation bar icons on the right side were large and distorted. This fixes that issue by applying constraints.

pablomarti:
Ok will check it right now

ryancarlson:
thank you!

PR #178, opened 2017-10-31
Adobe analytics async

ryancarlson:
This PR moves all Adobe Analytics operations to a background queue as to not interfere with any interactions on the main thread.

There is also an update in the readme describing the current state of the branches.

After merging this branch, the develop branch will still contain code unsuitable for release. Hotfixes should be done from develop-xcode8

### DIFF
--- a/godtools/Analtyics/GodToolsAnaltyics.swift
+++ b/godtools/Analtyics/GodToolsAnaltyics.swift
@@ -116,8 +116,8 @@ class GodToolsAnaltyics {
     private func recordScreenViewInAdobe(screenName: String) {
         var properties: [String: String] = [:]
         
-        properties[AdobeAnalyticsConstants.Keys.screenName] = "\(screenName)"
-        properties[AdobeAnalyticsConstants.Keys.previousScreenName] = "\(previousScreenName)"
+        properties[AdobeAnalyticsConstants.Keys.screenName] = screenName
+        properties[AdobeAnalyticsConstants.Keys.previousScreenName] = previousScreenName
         properties[AdobeAnalyticsConstants.Keys.appName] = AdobeAnalyticsConstants.Values.godTools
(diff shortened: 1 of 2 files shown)


ryancarlson:
Hey @pablomarti this PR contains some fixes that I think should be done to make the Adobe Analytics code work better. However these changes won't solve the performance issues. I suspect those are related libraries we had to update (RealmSwift and SWXMLHash) to build the project in XCode 9. One of them either has a bug or we're using it incorrectly.

The branch `develop-xcode8` has the the commits that updated those libraries removed as well as these three commits which I think are generally helpful.

We can merge this PR, but all hot-fixes should be done from `develop-xcode8` until we have time to properly fix the performance issues in `develop`. Does that make sense?

pablomarti:
Ok @ryancarlson understood! Will check right now

godtools/Analtyics/GodToolsAnaltyics.swift
```diff
@@ -102,14 +107,17 @@ class GodToolsAnaltyics {
         
         tracker?.send(screenViewInfo)
         
-        recordScreenViewInAdobe(screenName: screenName)
+        adobeAnalyticsBackgroundQueue.async { [unowned self] () in
+            self.recordScreenViewInAdobe(screenName: screenName)
```
pablomarti: Maybe use a weak self? (I don't know what is `unowned self`, will check)

pablomarti: Oh I see, this is cool 👍

ryancarlson:
Thanks!

godtools/Analtyics/GodToolsAnaltyics.swift
ryancarlson: Yeah a `weak` self could be nil, which would mean having to unwrap the optional. I don't see any way that self could be nil here so `unowned` should work and thus removes the optional. (I don't understand what the term "unowned" really means here, but it is the term anyways)

PR #179, opened 2017-11-07
Add support for iPhoneX

pablomarti:
This PR fixes:
- GT-435: Fix navbar in renderer
- GT-436: Fix cards in renderer
- GT-438: Fix navigation bar icons on iPhone 6S


pablomarti:
Hey @ryancarlson would you please check this PR? We still need feedback from Ann and Eric.

pablomarti:
@ryancarlson if you want to test in the simulator for iPhoneX please go to the file UIDevice.swift and in the method `iPhoneX()` change the condition to be: `return modelName == "x86_64"` (which is the simulator).

godtools/Views/TractElements/TractCallToAction.swift
```diff
@@ -47,16 +47,15 @@ class TractCallToAction: BaseTractElement {
...
     
     override func loadFrameProperties() {
-        var yPosition: CGFloat {
-            var position = self.elementFrame.y + TractCallToAction.yMarginConstant
-            if position < (self.parent?.getMaxHeight())! - self.height {
-                position = (self.parent?.getMaxHeight())! - self.height
-            }
-            return position
+        let bottomConstant = TractPageContainer.marginBottom
+        let minPosition = (self.parent?.getMaxHeight())! - self.height - bottomConstant
```
ryancarlson: We should be able to guarantee that the parent is defined so force-unwrapping is okay. I'd prefer to see it written this way though, as it's easier to read:
````swift
let minPosition = self.parent!.getMaxHeight() - self.height - bottomConstant

File: godtools/Views/TractElements/TractPage.swift [hunk 2]
```diff
@@ -15,9 +15,15 @@ import SWXMLHash
 
 class TractPage: BaseTractElement {
     
+    var pageContainer: TractPageContainer?
+    
     // MARK: - Setup
     static var navbarHeight: CGFloat = 0.0
     
+    static var statusbarHeight: CGFloat {
+        return UIDevice.current.iPhoneX() ? CGFloat(44) : CGFloat(0)
```
ryancarlson: These should probably use the constants I created in `TractViewController` . The would have to be made static though.

godtools/Views/TractElements/TractPageContainer.swift
```diff
@@ -0,0 +1,48 @@
...
+    override var properties: TractProperties {
+        get {
+            return (self.parent?.properties)!
+        }
+        set { }
+    }
+    
+    // MARK: - Setup
+    static var marginBottom: CGFloat {
+        return UIDevice.current.iPhoneX() ? CGFloat(44.0) : CGFloat(0.0)
```
ryancarlson: These should be made constants to. They are the same values as the status bar height, but they are use for different things, right? I think it's correct to make another set of constants in this class.

At some point we should try to consolidate our constants somehow, but that could be a separate task for another time.

pablomarti:
Hey @ryancarlson I uploaded the changes requested :D

ryancarlson:
@pablomarti should we do "squash and merge" this time?

PR #180, opened 2017-11-08
GT-434: Fix resolution/aspect ratio on loading screen

pablomarti:


pablomarti:
@ryancarlson would you please check this PR? I tested in the simulator of iPhone 8, iPhone 8+, iPhone X and iPhone SE and it is working well.

pablomarti:
Thanks @ryancarlson !

PR #183, opened 2017-11-16
Add text style properties [bold, italic, underline]

pablomarti:


godtools/GeneralExtensions/UIFont.swift
```diff
@@ -77,4 +85,27 @@ extension UIFont {
         return UIFont(name: fontName, size: fontSize)!
     }
     
+    
+    func setBold() -> UIFont {
+        if isBold {
+            return self
+        } else {
```
ryancarlson: Since you return from the if, there's no need for the `else` statement and block. Just remove the else and there's one less level of indentation.

godtools/GeneralExtensions/UIFont.swift
```diff
@@ -77,4 +85,27 @@ extension UIFont {
...
+            let fontDescriptorVar: UIFontDescriptor = fontDescriptor.withSymbolicTraits(.traitBold)!
```
ryancarlson: Is there any chance the force unwrap could cause an issue here?

We could always use a `guard let` to prevent that, and then return something safe if needed.

godtools/GeneralExtensions/UIFont.swift
```diff
@@ -77,4 +85,27 @@ extension UIFont {
...
+            return UIFont(descriptor: fontDescriptorVar, size: 0)
+        }
+    }
+    
+    func setItalic()-> UIFont {
+        if(isItalic) {
+            return self
+        } else {
```
ryancarlson: Since you return from the if, there's no need for the else statement and block. Just remove the else and there's one less level of indentation.

godtools/GeneralExtensions/UIFont.swift
```diff
@@ -77,4 +85,27 @@ extension UIFont {
...
+            let fontDescriptorVar: UIFontDescriptor = fontDescriptor.withSymbolicTraits(.traitItalic)!
```
ryancarlson: Is there any chance the force unwrap could cause an issue here?

We could always use a guard let to prevent that, and then return something safe if needed.

ryancarlson:
👍 i think this is good!

pablomarti:
Thanks @ryancarlson !

PR #187, opened 2017-12-18
Use a guard let statement instead of force unwrap on translation.language

ryancarlson:
Force unwrapping is probably causing the issue: https://fabric.io/cru/ios/apps/org.cru.godtools/issues/5992cddfbe077a4dcc1c73a4?time=last-thirty-days


ryancarlson:
@pablomarti el primero

PR #188, opened 2017-12-18
Use guard let instead of force unwrap to prevent crash

ryancarlson:
Should fix: https://fabric.io/cru/ios/apps/org.cru.godtools/issues/5a26ca2761b02d480dedfd23?time=last-thirty-days

### DIFF
--- a/godtools/Views/Cells/LanguageTableViewCell.swift
+++ b/godtools/Views/Cells/LanguageTableViewCell.swift
@@ -21,8 +21,11 @@ class LanguageTableViewCell: UITableViewCell {
     var cellDelegate: LanguageTableViewCellDelegate?
     var language: Language? {
         didSet {
-            languageExists(language!.shouldDownload)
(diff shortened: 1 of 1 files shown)


ryancarlson:
@pablomarti el segundo

godtools/Views/Cells/LanguageTableViewCell.swift
```diff
@@ -21,8 +21,11 @@ class LanguageTableViewCell: UITableViewCell {
     var cellDelegate: LanguageTableViewCellDelegate?
     var language: Language? {
         didSet {
-            languageExists(language!.shouldDownload)
-            languageLabel.text = language!.localizedName()
+            guard let language = language else {
+                return
```
pablomarti: Just a question, shouldn't you set language(Exists) to false?

ryancarlson: ah, si!! good catch!

PR #189, opened 2017-12-18
Return cell immediately if section exceeds number of resources in the array

ryancarlson:
I'm not sure how this would happen, probably some sort of race condition. This fix would prevent a crash, and the result would be whatever was in the dequeued cell would be returned. This may produce undefined behavior, but at least the crash is avoided.


ryancarlson:
@pablomarti el tercero

ryancarlson:
Do you prefer it just return an empty UITableViewCell() ? or HomeToolTableViewCell() ?

pablomarti:
I believe the best would be an empty UITableViewCell() @ryancarlson, I believe it is going to be easier to "hide" the problem from the user with that.

ryancarlson:
Cool, i will make that change, then good to merge?

PR #190, opened 2017-12-18
Fabric issue 83

ryancarlson:
also includes a fix for swift4 deprecation. we can now access strings directly w/o using 'characters'


ryancarlson:
@pablomarti el último (which is good, because i don't know how to say "fourth" in spanish :( )

PR #191, opened 2017-12-19
Fix downloads ux

ryancarlson:
Track English download when resource is not available in primary language

This will prevent weird UI/UX b/c w/o this change the AddTools screen will not show any update or pop b/c all notifications are filtered out.

fixes: https://jira.cru.org/browse/GT-440


ryancarlson:
@pablomarti could you review one more small PR please? this is an issue Ann noticed.

ryancarlson:
gracias para revisarlo @pablomarti !

PR #192, opened 2017-12-19
Snapshot configuration

ryancarlson:
Adds support to automatically generate screenshots that are localized in the languages that the App Store content is localized to. The benefit is that users of those languages will now see screenshots that are localized to their language.

There is a known issue with Chinese Traditional/Simplified that's been discussed in HipChat. iOS capitalizes the H in zh-Hans and zh-Hant, but it's lower case in our API. Our code is doing a case sensitive search and therefore can't find the match. We could change the `h` case in the API and solve the problem, but Frett thinks that would create a small problem for Android. We're still investigating.

This PR adds accessibility identifiers and uses a UI test along with Fastlast Snapshot to generate the screenshots.


ryancarlson:
(not ready for review yet)

PR #194, opened 2018-02-12
Gt 443 modify onboarding screens

ryancarlson:
- first commit is the actual change. removes multiple buttons and therefore decision point that was confusing/annoying users
- second commit removes `self` references where they aren't necessary.

### DIFF
--- a/godtools/ViewControllers/Platform/OnboardingViewController.swift
+++ b/godtools/ViewControllers/Platform/OnboardingViewController.swift
@@ -14,9 +14,8 @@ class OnboardingViewController: BaseViewController {
     @IBOutlet weak var page2View: UIView!
     @IBOutlet weak var pageControl: UIPageControl!
     
-    @IBOutlet weak var nowButton: TransparentButton!
-    @IBOutlet weak var laterButton: TransparentButton!
-    @IBOutlet weak var okayButton: TransparentButton!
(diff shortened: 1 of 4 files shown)


ryancarlson:
@wiseguy16 can you review please?

godtools/ViewControllers/Platform/OnboardingViewController.swift
```diff
@@ -14,35 +14,33 @@ class OnboardingViewController: BaseViewController {
     @IBOutlet weak var page2View: UIView!
     @IBOutlet weak var pageControl: UIPageControl!
     
-    @IBOutlet weak var nowButton: TransparentButton!
-    @IBOutlet weak var laterButton: TransparentButton!
-    @IBOutlet weak var okayButton: TransparentButton!
+    @IBOutlet weak var languagesOkayButton: TransparentButton!
+    @IBOutlet weak var toolsOkayButton: TransparentButton!
```
wiseguy16: Don't know if these were being used in Storyboards also...just wondering if there were any "orphaned" outlets when these got switched?

ryancarlson: Yeah, I went through the corresponding .xib file and made sure the appropriate references were where they needed to be.

ryancarlson:
Thanks for the review!

PR #195, opened 2018-02-12
Gt 378 call to action button color

ryancarlson:
Allows the call to action button color to be customized. If the optional control-color attribute is not set, then the fallback will be to page.primary-color and manifest.primary-color in that order.

Tested on stage: w/ control-color set, with page.primary-color set, with manifest.primary-color set. See KGP in stage for different scenarios.


ryancarlson:
hey @pablomarti , could you review this PR for me please?

ryancarlson:
gracias mi amigo!

PR #197, opened 2018-02-13
Fix header background setting

ryancarlson:
In recent commit, it was mistakenly set to "page background" which is always white. It should have been set to "page primary color" which is set at the page or manifest level.

This was a small refactoring mistake.


ryancarlson:
hey @pablomarti i had a small refactoring mistake. could you please review. this mistake made the background of all of the headers at the top white! no one could see the text.

PR #199, opened 2018-02-16
Gt 454 account creation

ryancarlson:


ryancarlson:
@wiseguy16 could you review this sometime next week please? it's lower priority, so if you're able to have a look sometime before I'm back in the office Thursday that'd be fine.

/cc @pablomarti if you're interested to see :)

godtools/ViewControllers/Login/LoginViewController.swift
```diff
@@ -0,0 +1,138 @@
...
+        guard let loginClient = loginClient else {
+            return
+        }
+        
+        loginClient.passwordGrantLogin(for: username!, password: password!) { [unowned self] (result, error) in
+            switch result {
+            case .success:
+                self.showAlert(withTitle: "success".localized, message: "")
+                self.setViewState()
+                debugPrint(loginClient.guid())
```
ryancarlson: note to self, remove these `debugPrint()` lines

godtools/ViewControllers/Login/LoginViewController.swift
```diff
@@ -0,0 +1,138 @@
...
+
+class LoginViewController: BaseViewController {
+    
+    @IBOutlet weak var usernameTextField: UITextField!
+    @IBOutlet weak var passwordTextField: UITextField!
+    
+    @IBOutlet weak var loginButton: BlueButton!
+    @IBOutlet weak var logoutButton: BlueButton!
+    
+    private let serverURL = URL(string: "https://thekey.me/cas/")!
```
ryancarlson: note to self, move these properties to a properties file/class

godtools/ViewControllers/Login/LoginViewController.swift
wiseguy16: I like that your using the .count on Strings here since it's now available instead of someWord.characters.count

ryancarlson:
Thanks for reviewing!

PR #200, opened 2018-02-27
Listeners are space separated, not comma separated

ryancarlson:
per the content spec: https://github.com/CruGlobal/mobile-content-api/blob/master/public/xmlns/content.xsd#L124


ryancarlson:
Thanks!

PR #201, opened 2018-02-27
Gt 444 language setting experience

ryancarlson:
pop back to previous view after selecting language by the download button. also clean up code a bit.


ryancarlson:
Thanks!

PR #202, opened 2018-02-27
Universal Links

ryancarlson:
Adds support for Universal Links.

Both the cloudfront distribution domain and knowgod.com are supported and tracked. If a universal link is transferred to the app, it will open the resource and navigate to the desired page. Some transitions are seen.

If necessary, the specific resource/language combination is downloaded. However other resources in that language, or that resource in other languages will not be downloaded.

There is certainly room for debate on the UX after a user returns from viewing a universal linked resource.


pablomarti:
@ryancarlson just to be sure, if the user did configure 2 langs (English and Spanish for example) and then opens the content in Portuguese for example, the content should be displayed in Portuguese without the segmented control, right?

godtools/AppDelegate.swift
```diff
@@ -16,7 +16,7 @@ import RealmSwift
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
     var window: UIWindow?
-    var flowController: BaseFlowController?
+    var flowController: PlatformFlowController?
```
pablomarti: I understand this because of the function `goToUniversalLinkedResource`, but did you test this @ryancarlson and did you find no problems when using the app? Was concerned about this, but only a test to see if everything is showing up will be enough.

godtools/AppDelegateExtensions/AppDelegate+UniversalLinks.swift
```diff
@@ -0,0 +1,88 @@
...
+    
+    func application(_ application: UIApplication, continue userActivity: NSUserActivity, restorationHandler: @escaping ([Any]?) -> Void) -> Bool {
+        if userActivity.activityType != NSUserActivityTypeBrowsingWeb {
+            return false
+        }
+        
+        guard let url = userActivity.webpageURL else {
+            return false
+        }
+
```
pablomarti: Shouldn't be good to check here `if pathParts.count < 3 ` in order to avoid the conditions in the following methods?

godtools/AppDelegate.swift
ryancarlson: Ah, that should probably still be `BaseFlowController`. In another PR I'm introducing `ArticleFlowController` for the Questions about God articles. I'll look at this some more.

ryancarlson:
@pablomarti To answer your question above... yes the content should open in Portuguese. The segment control would still show the parallel language, if one is selected and the content is available in the parallel language.

This may or may not be the best behavior to show the parallel language, I'm not sure.. but it should work that way.

godtools/AppDelegate.swift
pablomarti: Oh ok, yes so in that way maybe will require a small refactoring.

godtools/AppDelegateExtensions/AppDelegate+UniversalLinks.swift
ryancarlson: Actually, I think it's better w/o that. The individual methods that check for resource, language, page in the URL check to make sure that the part of the path it's looking for exists before it accesses it. I change put in this statement here, but I think it will be more confusing and redundant.

````swift
        guard let url = userActivity.webpageURL, url.pathComponents.count >= 4 else {
            return false
        }
````

godtools/AppDelegate.swift
ryancarlson: Fixed, thanks!

ryancarlson:
hey @pablomarti i fixed the flow controller class issue in the AppDelegate extension, and left a comment on the other thing. Am I good to go now?

PR #204, opened 2018-03-29
Gt 503 adb track actions

wiseguy16:
This tracks all requested actions and sends to ADBMobile.  There is no develop branch, so didn't know if we need one to base against?


godtools/Managers/LanguagesManager.swift
```diff
@@ -75,16 +75,18 @@ class LanguagesManager: GTDataManager {
         if GTSettings.shared.primaryLanguageId == nil {
             return nil
         }
-        
-        return loadFromDisk(id: GTSettings.shared.primaryLanguageId!)
+        let primaryLanguage: Language? = loadFromDisk(id: GTSettings.shared.primaryLanguageId!)
```
ryancarlson: No need to define the type `Language?` here.. Swift will infer it.

godtools/Managers/LanguagesManager.swift
```diff
@@ -75,16 +75,18 @@ class LanguagesManager: GTDataManager {
...
+        UserDefaults.standard.set((primaryLanguage?.code ?? ""), forKey: "kPrimaryLanguageCode")
```
ryancarlson: I think there's a better place to set the primary code, instead of when loading the primary language. Can we set the code when the ID gets set during language selection? That seems more consistent.

godtools/Managers/LanguagesManager.swift
```diff
@@ -75,16 +75,18 @@ class LanguagesManager: GTDataManager {
...
+        return primaryLanguage
     }
     
     func loadParallelLanguageFromDisk() -> Language? {
         if GTSettings.shared.parallelLanguageId == nil {
             return nil
         }
-        
-        return loadFromDisk(id: GTSettings.shared.parallelLanguageId!)
+        let parallelLanguage: Language? = loadFromDisk(id: GTSettings.shared.parallelLanguageId!)
```
ryancarlson: `Language?` doesn't need to be specified here.

godtools/Views/TractElements/TractCard+Actions.swift
```diff
@@ -64,11 +66,23 @@ extension TractCard {
...
+        if !relay.isTimerRunning {
+            relay.timerCounter = 6
+            relay.runTimer()
+        }
+        
+        let cardName = properties.cardIdName
+        let tractPlusCardName = "\(relay.tractName)\(cardName)"
+        relay.tractPlusCardName = tractPlusCardName
+        
+        // TODO - This isn't quite tracking correct values. It's appending letters and not replacing???
```
ryancarlson: Did you get this solved?

godtools/Views/TractElements/TractCardProperties.swift
```diff
@@ -61,5 +61,6 @@ class TractCardProperties: TractProperties {
     
     var cardState = CardState.preview
     var cardNumber = 0
+    var cardIdName = ""
```
ryancarlson: This name is a little bit confusing to me. I'm not sure at face value what value it represents.

godtools/Views/TractElements/TractCards+ChildSetup.swift
```diff
@@ -47,8 +51,12 @@ extension TractCards {
             let element = TractCard(data: dictionary, startOnY: yPosition, parent: self, elementNumber: elementNumber)
             element.yDownPosition = yDownPosition - TractPage.navbarHeight
             element.cardProperties().cardNumber = cardNumber
+            let alphaCode = element.cardProperties().cardNumber.convertToLetter()
+            element.cardProperties().cardIdName = "\(alphaCode)"
```
ryancarlson: Does this need to be tracked as a separate property? I'm not seeing how it's being used.

ryancarlson: It looks like we could just call `cardNumber.convertToLetter()` in TractCards+Actions , then cardIdName isn't necessary.

godtools/AnalyticsRelay.swift
```diff
@@ -0,0 +1,81 @@
...
+import Foundation
+
+class AnalyticsRelay {
+    
+    static let shared = AnalyticsRelay()
+    
+    var tractName: String = ""
+    var tractPlusCardName: String = ""
+    var tractCardCurrentNames: [String] = []
+    var tractCardNextNames: [String] = []
```
ryancarlson: I removed this property and the one above it and didn't encounter any issues, other that where the values are cleared and set. How are they used?

godtools/Managers/LanguagesManager.swift
wiseguy16: I moved setting these values to new spots. We can get the parallel language code when they select from the tableview, but to get primary code, I had to move it to the defaultLanguage getter and setter.  Suggestions for alternatives?

godtools/Views/TractElements/TractCardProperties.swift
wiseguy16: I've changed this to cardLetterName.  I think that communicates what we're trying to get at now

godtools/Views/TractElements/TractCards+ChildSetup.swift
wiseguy16: I'm now using cardLetterName and getting from cardNumber.convertToLetter()

godtools/AnalyticsRelay.swift
wiseguy16: I thought I would need the values here, but I think I can get values without these arrays. They've now been removed

godtools/Managers/LanguagesManager.swift
```diff
@@ -185,6 +184,7 @@ class LanguagesManager: GTDataManager {
 
         } else {
             GTSettings.shared.parallelLanguageId = language.remoteId
+            UserDefaults.standard.set((language.code), forKey: "kParallelLanguageCode")
```
ryancarlson: Why not set the primary language's code in the `if` branch just above?

PR #894, opened 2021-11-02
GT-1308: Video on First Onboarding Screen

reldredge71:
Added YouTube video player to the first screen of onboarding.  It's not perfect right now, there are some limitations that I ran into that's hard to resolve:

- I had to create a new screen to hold the UIView for the YouTube video player.
- The new screen appears on top of the TutorialPagerView like a drawer, with a bit of space at the top.  You can close the screen by swiping it down, and I believe that will stop the video and clean up the player.
- Preferably, I would like to have the video play in full screen, and use the "X" icon in YouTube's player view to close both the player and the drawer screen that contains it together.  So just pressing the "X" once should bring you right back to TutorialPagerView.  However, I can't find a good way to attach a callback to the "X" button that closes full screen, so I don't know how to make that work.  My temporary solution is to disable the full-screen for the video player, and have the user just swipe to close the drawer.  This is not ideal, but it should work.
- The video player and the drawer should both close once the video reaches the end, which is the desired behavior.


levieggertcru:
Hey @reldredge71 what your seeing is the default ```modalPresentationStyle``` for a ```UIViewController```.  I believe there is a ```fullScreen``` you can set that to.

levieggertcru:
Hey @reldredge71 is it possible to hide the control for closing the youtube video and use our own close button in the navigation bar?

reldredge71:
@levieggertcru Okay I can look into making it full screen and putting in my own close button.

levieggertcru:
Hey @reldredge71 go ahead and create a new PR for this.  We are now on GHA :)

PR #897, opened 2021-11-03
GT-1297 GitHub actions

levieggertcru:


levieggertcru:
@gyasistory @frett let me know what you think.  Everything is in the build.yml and there's just one job.  That seemed to be the simplest without repeating steps and everything can run on macos-11 which chooses Xcode 13.

frett:
> @gyasistory @frett let me know what you think. Everything is in the build.yml and there's just one job. That seemed to be the simplest without repeating steps and everything can run on macos-11 which chooses Xcode 13.

Repeating steps between multiple jobs is something that is annoying in GHA. The only solution I've seen is to define a custom step that contains the shared logic. But that seems like a lot of overhead.

On other repos we have duplicated steps when it makes sense to separate a job into multiple jobs

.github/workflows/build.yml
```diff
@@ -28,33 +22,35 @@ jobs:
       - name: Update Fastlane
         run: bundle update fastlane
```
frett, line 23: you shouldn't need to update fastlane here, this is handled automatically by dependabot or renovate

.github/workflows/build.yml
```diff
@@ -28,33 +22,35 @@ jobs:
...
         run: bundle exec fastlane cru_shared_lane_download_and_commit_latest_one_sky_localizations
 
-      - name: Decode AppleAppStoreApi.json.enc For Fastlane App Store Connect Access 
-        if: github.event_name == 'workflow_dispatch'
+      - name: Decrypt AppleAppStoreApi.json.enc For Fastlane App Store Connect Access 
+        if: github.event_name == 'push' && github.ref == 'refs/heads/master'
         env: 
           KEY: ${{ secrets.ENCRYPTED_CF3A694B4ADE_KEY }}
           IV: ${{ secrets.ENCRYPTED_CF3A694B4ADE_IV }}
         run: openssl aes-256-cbc -K ${KEY} -iv ${IV} -in fastlane/AppleAppStoreApi.json.enc -out fastlane/AppleAppStoreApi.json -d
```
frett, line 45: If the API key is an actual json file, we can just store the contents of that file as a secret instead of having to use openssl to decode the file

.github/workflows/build.yml
```diff
@@ -28,33 +22,35 @@ jobs:
...
 
       - name: Increment Xcode Project Build Number
-        if: github.event_name == 'workflow_dispatch'
+        if: github.event_name == 'push' && github.ref == 'refs/heads/master'
         run: bundle exec fastlane cru_shared_lane_increment_xcode_project_build_number
```
frett, line 49: why did you create new fastlane lanes instead of using the ones we had?

frett (changes requested):
several questions, the big one being, why did you create new fastlane lanes instead of using what we already had defined?

frett:
here is GHA for MissionHub using our already existing fastlane lanes: https://github.com/CruGlobal/missionhub-react-native/blob/develop/.github/workflows/ci-cd.yml#L226-L242

.github/workflows/build.yml
levieggertcru: Hey @frett there are a number of reasons.

levieggertcru: @frett  I'll gather some notes and include them here.

levieggertcru: Hey @frett the biggest reason was how code signing and match was done, the cru fastfile is including MPDX's Cal Directory Extension in code signing and match which isn't necessary and would rather opt for a more dynamic approach to this.  That way if we have iOS apps that include other app extensions we don't bloat the cru fastfile.  Also with match you don't need to create a new match for every app extension.  Match supports multiple bundle ids.

levieggertcru: @frett also some of the naming, for example, in localizations download they're using ```CRU_SCHEME``` which actually has nothing to do with the localizations download, that is actually a root directory name.

levieggertcru: @frett I do want to look into including a firebase distribution in the cru fastfile as well which would require different code signing than what is used for TestFlight, which is another reason I wanted to break up some of the lanes.

frett: I definitely agree about not including the MPDX call directory stuff in the shared fastlane file, I hadn't realized a contractor had done that several years ago. I would have definitely pushed more a more configurable option, or at least putting some call directory specific lanes in the MPDX Fastfile.

.github/workflows/build.yml
levieggertcru: Hey @frett I like this idea better, I will look into this.

.github/workflows/build.yml
```diff
@@ -28,33 +22,35 @@ jobs:
       - name: Update Fastlane
         run: bundle update fastlane
 
+        # Required for KotlinMultiplatform
+      - uses: actions/setup-java@v1
+        with:
+          java-version: '11'
+
       - name: Pod Install
         run: bundle exec fastlane cru_shared_lane_pod_install
```
frett, line 31: any reason you are preferring to install the pods via fastlane instead of just running `pod install` here?

.github/workflows/build.yml
frett: ok, I'm going to process the new lanes a bit, it would be nice if there was a PR with the new lanes so I can leave some comments there

.github/workflows/build.yml
levieggertcru: Hey @frett no preferred reason here, I hadn't thought about using pod install, but will try that, one less lane to have.

.github/workflows/build.yml
frett: we've done it for some Android projects, let me get you a link to a sample

frett: https://github.com/CruGlobal/godtools-android/blob/develop/.github/workflows/build.yml#L115-L118

frett: We can generate a new App Store API key from their UI so you don't have to worry about extracting the current one. let me know if you need help with this or can't access it.

.github/workflows/build.yml
frett: Using more direct steps when fastlane isn't actually required (e.g. cocoapods) makes it much more understandable for users who don't know fastlane.

frett:
@levieggertcru it would be beneficial to include this cancel previous jobs job: https://github.com/CruGlobal/godtools-android/blob/develop/.github/workflows/build.yml#L10-L17

that way if there are multiple new commits to a branch/pr it will avoid duplicated work by GHA

fastlane/.env.default
```diff
@@ -1,25 +1,29 @@
...
-CRU_TEST_DEVICES="iPhone 7"
 
-CRU_DEFAULT_BRANCH="master"
 
-## AppStore release related variables
-CRU_API_KEY_PATH="./fastlane/AppleAppStoreAPI.json"
-CRU_IDFA_IS_ENABLED=true
-CRU_IDFA_TRACKS_INSTALL=true
+APPLE_DEVELOPER_TEAM_ID = "DQ48D9BF2V"
+APPLE_ID_USERNAME = "cruapps@cru.org"
```
frett, line 4: as far as I know, this option is no longer required now that we are using an API key

.github/workflows/build.yml
levieggertcru: thanks @frett , I will remove this step

godtools/App/Services/Renderer/BackgroundImageRenderer/MobileContentBackgroundImageRendererType.swift
```diff
@@ -6,7 +6,7 @@
 //  Copyright © 2021 Cru. All rights reserved.
 //
 
-import Foundation
+import UIKit
```
frett, line 9: it looks like there was some code updates that got included with this PR. I'm guessing that wasn't intentional.

Are these updates something that should be handled in a different PR, or are they changes that don't matter being bundled here?

frett: nm, I see the commit. you were fixing the tests

.github/workflows/build.yml
levieggertcru: @frett just created a PR for those new lanes

.github/workflows/build.yml
levieggertcru: @frett I think I can use this instead https://docs.fastlane.tools/actions/app_store_connect_api_key/

How would we go about generating a new App Store API key and using that?

frett: that is what the `AppleAppStoreApi.json` key is ;) it's found on app store connect under "Users and Access" -> "Keys"

.github/workflows/build.yml
```diff
@@ -40,12 +46,11 @@ jobs:
...
-      - name: Decrypt AppleAppStoreApi.json.enc For Fastlane App Store Connect Access 
+      - name: Import App Store Connect API Key
         if: github.event_name == 'push' && github.ref == 'refs/heads/master'
-        env: 
-          KEY: ${{ secrets.ENCRYPTED_CF3A694B4ADE_KEY }}
-          IV: ${{ secrets.ENCRYPTED_CF3A694B4ADE_IV }}
-        run: openssl aes-256-cbc -K ${KEY} -iv ${IV} -in fastlane/AppleAppStoreApi.json.enc -out fastlane/AppleAppStoreApi.json -d
+        env:
+          APP_STORE_CONNECT_API_JSON_PAYLOAD: ${{ secrets.APP_STORE_CONNECT_API_JSON_PAYLOAD }}
+        run: echo $APP_STORE_CONNECT_API_JSON_PAYLOAD > fastlane/AppleAppStoreApi.json
```
frett, line 53: 🎉

frett (approved):

frett (approved):

PR #925, opened 2021-11-16
GT-1308 Fix positioning image to right of button title

levieggertcru:
Hey @reldredge71 this should do it.  One I had to force the layout on the button by calling layoutIfNeeded, and also added an override for when the view is actually added to the superview and updating the button layout when that happens because when we configure that view it's not even part of the view hierarchy at that point.

Also fixed the shared extension by including the imageWidth when positioning the image to the right of the button title. Including that imageWidth places the image left edge right on the title right edge.


reldredge71 (approved):
Great, thank you for looking into it!

PR #933, opened 2021-11-18
GT-1309: Onboarding Quick Start Screen

reldredge71:
Add a screen to the end of the onboarding flow with three "quick start" links leading to different parts of the app.


reldredge71:
@levieggertcru I was having trouble spacing out the cells of the table view.  One thing that I read said to render each cell in its own section, and set the header and footer heights.... now they look a little too far apart, even when header and footer heights are zero.  Any thoughts on how to control this?  Should I use something other than UITableView?

levieggertcru:
Hey @reldredge71 simplest way I found to space cells is add the spacing within the cell.

reldredge71:
That worked great, thank you!

godtools/App/Features/Onboarding/OnboardingQuickStartView.swift
```diff
@@ -0,0 +1,149 @@
...
+        
+        quickStartTableView.delegate = self
+        quickStartTableView.dataSource = self
+    }
+    
+    private func setupLayout() {
+        
+        titleLabel.text = viewModel.title
+        
+        endTutorialButton.setTitle(viewModel.endTutorialButtonTitle, for: .normal)
```
levieggertcru, line 62: Hey @reldredge71 can you move any configuring / observing of the viewModel into setupBinding() just to keep consistent with other views.

godtools/App/Features/Onboarding/OnboardingQuickStartView.swift
```diff
@@ -0,0 +1,149 @@
...
+                
+        let quickStartItem = viewModel.quickStartCellWillAppear(index: indexPath.row)
+        cell.configure(item: quickStartItem)
+        
+        return cell
+    }
+    
+    func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
+        
+        let headerView = UIView()
```
levieggertcru, line 144: Hey @reldredge71 is this doing anything?  I see in Figma this table list doesn't have any type of header UI so I am curious the intent of this.

godtools/App/Features/Onboarding/OnboardingQuickStartView.swift
```diff
@@ -0,0 +1,149 @@
...
+        
+        quickStartTableView.separatorStyle = .none
+        quickStartTableView.rowHeight = UITableView.automaticDimension
+    }
+    
+    private func setupBinding() {
+        
+    }
+    
+    private func reloadData() {
```
levieggertcru, line 79: Hey @reldredge71 this method isn't doing anything, is it needed?

godtools/App/Features/Onboarding/Views/OnboardingQuickStartCell.swift
```diff
@@ -0,0 +1,41 @@
...
+//
+
+import UIKit
+
+class OnboardingQuickStartCell: UITableViewCell {
+    
+    static let nibName: String = "OnboardingQuickStartCell"
+    static let reuseIdentifier: String = "OnboardingQuickStartCellReuseIdentifier"
+        
+    private var item: OnboardingQuickStartItem?
```
levieggertcru, line 16: Hey @reldredge71 I see you added a ViewModel for this cell, but it hasn't been added to the Xcode project, I do see it in Finder however.  Can you use a ViewModel here instead of accessing the model directly to keep consistent with the View/ViewModel pattern.

godtools/App/Features/Onboarding/Views/OnboardingQuickStartCell.swift
```diff
@@ -0,0 +1,41 @@
...
+        super.prepareForReuse()
+        
+        titleLabel.text = nil
+        subtitleLabel.text = nil
+    }
+    
+    func configure(item: OnboardingQuickStartItem) {
+        
+        self.item = item
+
```
levieggertcru, line 37: Can you set the ```selectionStyle``` on this cell to ```none```.  By default a color is used which usually doesn't work well with our custom cells.
https://developer.apple.com/documentation/uikit/uitableviewcell/1623221-selectionstyle

godtools/App/Features/Onboarding/OnboardingQuickStartView.swift
```diff
@@ -0,0 +1,149 @@
...
+//
+//  Created by Robert Eldredge on 11/8/21.
+//  Copyright © 2021 Cru. All rights reserved.
+//
+
+import UIKit
+
+class OnboardingQuickStartView: UIViewController {
+    
+    enum DismissOnboardingTutorialType {
```
levieggertcru, line 13: This type doesn't look right to me here because DismissOnboardingTutorialType is related to steps to take in the Flow.  If this type were used by this View that would make more sense to have it declared here.  See my comment in FlowStep ```dismissOnboardingTutorial```.

godtools/App/Flows/FlowStep.swift
```diff
@@ -18,7 +18,8 @@ enum FlowStep {
     case showMenu
     case showLanguageSettings
     case showOnboardingTutorial(animated: Bool)
-    case dismissOnboardingTutorial
+    case dismissOnboardingTutorial(dismissOnboardingTutorialType: OnboardingQuickStartView.DismissOnboardingTutorialType?)
```
levieggertcru, line 21: I like the idea here of providing this dismiss type.  It fits well with some places in the app where I've been starting to implement a flow completed state.  See ```LessonFlowCompletedState``` and ```TractFlowCompletedState```.  I'd like to do the same here with our OnboardingFlow and have an ```OnboardingFlowCompletedState``` which will contain those cases you created in ```DismissOnboardingTutorialType```.

godtools/App/Features/Onboarding/OnboardingQuickStartView.xib
```diff
@@ -0,0 +1,103 @@
...
+                        <constraint firstAttribute="trailing" secondItem="l4Q-zP-eGG" secondAttribute="trailing" constant="30" id="Ca8-FY-dOM"/>
+                        <constraint firstItem="NBq-0r-pAH" firstAttribute="leading" secondItem="WFX-ff-YxE" secondAttribute="leading" constant="30" id="PU4-xU-xiE"/>
+                        <constraint firstAttribute="trailing" secondItem="NBq-0r-pAH" secondAttribute="trailing" constant="30" id="TB2-lM-0zY"/>
+                        <constraint firstAttribute="bottom" secondItem="l4Q-zP-eGG" secondAttribute="bottom" constant="24" id="dAa-O5-HTa"/>
+                        <constraint firstItem="l4Q-zP-eGG" firstAttribute="top" secondItem="NBq-0r-pAH" secondAttribute="bottom" constant="20" id="kHf-Hq-wIh"/>
+                        <constraint firstItem="l4Q-zP-eGG" firstAttribute="leading" secondItem="WFX-ff-YxE" secondAttribute="leading" constant="30" id="vFo-B7-pe3"/>
+                    </constraints>
+                </view>
+                <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="Pke-Cm-Fv3" userLabel="Footer View">
```
levieggertcru, line 55: Hey @reldredge71 I'm thinking we need to add something here for shorter devices where not all of the quick actions will be visible.  Can you add a white fade those items would fade into, I think that will help see that there are more items.


iPhone_5s

 
Something like this.



fade

levieggertcru (changes requested):

godtools/App/Features/Onboarding/OnboardingQuickStartView.swift
reldredge71: Ah, that was how I was trying to space out the cells before.  One StackOverflow thread said the only way they were able to add space was to put each cell in its own section and use the section headers for spacing.  I did what you suggested instead, to add the spacing in the .xib for the cell, but I forgot to remove this headerView.

godtools/App/Features/Onboarding/OnboardingQuickStartView.xib
reldredge71: I've been trying to get a gradient to work on this.  I wrapped the UITableView in a UIView and applied the gradient to that.  the "locations" of the points of the gradient don't make sense to me.   If the screen is so small that the table view starts scrolling, then I can't see the bottom fade-out at all.  It almost seems to me like the gradient's frame is taking its dimensions from the full height of the table view.... but the whole point of wrapping it in a UIView was to avoid that and have the gradient only fill the visible part of the table.  IDK what's up, you can play around with it if you have any ideas what the issue might be.

levieggertcru: Hey @reldredge71 that gradient does fill the entire view, I think in your case you'd have to set those locations with a combination of using clear color to get the desired effect.  Wrapping the table in a gradient view makes it more complex to calculate across device sizes too.
How I've done something similar in the past is have the tableview fill the entire screen and then setting the bottom contentInset of the tableview to match the height of the gradient area which overlaid the bottom of the tableview, that gave the effect of those items scrolling behind something. 
https://developer.apple.com/documentation/uikit/uiscrollview/1619406-contentinset

levieggertcru (approved):

PR #936, opened 2021-11-22
GT add git push to build workflow

levieggertcru:
Hey @frett adding that ```git push``` step to push any commits during the workflow.  I'm not sure if I will need to use any GITHUB_TOKEN or personal access token?


frett:
you should be able to use the `GITHUB_TOKEN`

frett:
we may need to make it a writable token, not totally sure until we try it

frett (approved):

PR #937, opened 2021-11-24
GT 1309 add fade

levieggertcru:
Hey @reldredge71 added a fade to the quick actions list.  Let me know what you think.
I also updated and flattened the view hierarchy.  A good tip I learned when reading about autolayout is if possible keep the hierarchy as flat as possible, makes it easier to work with and update constraints.

### DIFF
--- a/godtools/App/Features/Onboarding/OnboardingQuickStartView.xib
+++ b/godtools/App/Features/Onboarding/OnboardingQuickStartView.xib
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="19455" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="19162" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES">
     <device id="retina6_1" orientation="portrait" appearance="light"/>
     <dependencies>
         <deployment identifier="iOS"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="19454"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="19144"/>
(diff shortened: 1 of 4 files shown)


godtools/App/Features/Onboarding/OnboardingQuickStartView.xib
```diff
@@ -109,8 +80,8 @@
         </designable>
     </designables>
     <resources>
-        <systemColor name="systemBackgroundColor">
-            <color white="1" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
+        <systemColor name="systemPurpleColor">
+            <color red="0.68627450980392157" green="0.32156862745098042" blue="0.87058823529411766" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
```
reldredge71, line 84: What is this color change for?

godtools/App/Features/Onboarding/Views/OnboardingQuickStartCell.swift
```diff
@@ -13,8 +13,10 @@ class OnboardingQuickStartCell: UITableViewCell {
     static let nibName: String = "OnboardingQuickStartCell"
     static let reuseIdentifier: String = "OnboardingQuickStartCellReuseIdentifier"
     
+    @IBOutlet weak private var contentBackgroundView: UIView!
     @IBOutlet weak private var titleLabel: UILabel!
-    @IBOutlet weak private var subtitleLabel: UILabel!
+    @IBOutlet weak private var actionLabel: UILabel!
+    @IBOutlet weak private var actionIcon: UIImageView!
```
reldredge71, line 19: Do we need this outlet, if we're not using it in this view class?

godtools/App/Features/Onboarding/OnboardingQuickStartView.xib
levieggertcru: Hey @reldredge71 that systemPurpleColor is set on the bottomGradientView in the .xib.  It's okay to have it set there because it is going to be set programmatically when applying the gradient.  Sometimes I do that in the .xibs just to highlight elements and if they are going to have to be programmatically set.

godtools/App/Features/Onboarding/Views/OnboardingQuickStartCell.swift
levieggertcru: Hey @reldredge71 you're right that is not being used and we really don't have to connect it, it's not necessary or required.  It's more of a personal preference on my part, it helps me quickly see what elements make up a view without opening the .xib.  In the future migrating to SwiftUI will be nice because then there won't be anymore .xib and swift file.  All UI will be in one place.

reldredge71 (approved):

PR #951, opened 2021-12-03
GT-894 Integrate Okta

levieggertcru:
Ensure you can login and logout using Okta.  There is a known issue where the logout alert says Sign In instead of Sign Out.  This issue has been reported to Okta.

- Added  CRU Swift Package OktaAuthentication which depends on OktaOidc
- Added OktaUserAuthentication which conforms to UserAuthenticationType
- Integrated UserAuthenticationType with FirebaseAnalytics, SnowplowAnalytics, AccountViewModel, and MenuViewModel
- Removed TheKey from Cocoapods
- Removed TheKey related code


godtools/App/Services/UserAuthentication/Okta/OktaUserAuthentication.swift
```diff
@@ -0,0 +1,161 @@
...
+//  godtools
+//
+//  Created by Levi Eggert on 12/1/21.
+//  Copyright © 2021 Cru. All rights reserved.
+//
+
+import UIKit
+import OktaAuthentication
+
+class OktaUserAuthentication: UserAuthenticationType {
```
wrandall22, line 12: I wonder, why doesn't this live in the library? Nothing here seems GodTools specific. Is it because of where it lives in the flow of the app?

levieggertcru: Hey @wrandall22 it possibly could and I have been considering that.  The only conditions are 

1. OktaUserAuthentication assumes that Okta is configured with the offline_access in order for renewing authentication.  Although, I guess regardless if that scope was there are not this implementation would still function.

2. It assumes that auth user attributes are consistent between apps.  I'm not sure where those user attributes are defined and if they would match up between GodTools and MPDX.  That's a concern.

It would be nice to share this logic and that is something I'm considering.  It would also move the protocol UserAuthenticationType into the shared library which isn't that big of a deal.  I'd just want to make that more clear in the shared library.

wrandall22: > OktaUserAuthentication assumes that Okta is configured with the offline_access in order for renewing authentication. Although, I guess regardless if that scope was there are not this implementation would still function.
I would think all mobile apps would want offline_access, right?

> It assumes that auth user attributes are consistent between apps. I'm not sure where those user attributes are defined and if they would match up between GodTools and MPDX. That's a concern.
What are some examples of these attributes? Maybe I could help you know where they are defined.

levieggertcru: Hey @wrandall22 yeah I think all apps would want the offline_access.
At the moment GodTools is using "email", "given_name", "grMasterPersonId", "family_name", "ssoguid".
MPDX is using: "email", "name", "grMasterPersonId", "ssoguid" which does match up with GodTools.  I just want to ensure they would stay consistent.

levieggertcru: I also would want to remove the ObservableValue and Signal dependencies from UserAuthenticationType and use closures instead.

wrandall22: I see, so it sounds like those are attributes within Okta itself. Is `name` and `family_name`/`given_name` the same thing or are you using different fields for them? I suppose the easy way would be to include all the ones all the apps use and decide within each app whether or not to care about said field.

levieggertcru: Yeah those are defined in Okta.  I'm not sure about the name.  I need to run MPDX and see what fields it gives back to me.

wrandall22: Another option besides including all the properties would be to have a class in the library with the common properties and use class extensions in the apps where necessary.

levieggertcru: Yeah that's another option which the shared library does provide through a CruOktaUser.swift as well as methods for fetching straight user json and even a generic Decodable type if clients want to have the shared library decode a type provided by the client.

It would be good to move the refresh authorization logic to the shared library and possibly the UserAuthenticationType protocol since both MPDX and GodTools support that.

frett: to add a little bit of Okta context to this discussion, For each app configured in Okta you can choose to map user profile attributes however you want. so `ssoguid` could be `sso_guid` for a different app. it's really down to conventions within the app config to get any consistency with profile attribute names.

frett: on android I did create a centralized definition of common profile attributes and accessors in gto-support: https://github.com/CruGlobal/android-gto-support/blob/master/gto-support-okta/src/main/kotlin/org/ccci/gto/android/common/okta/oidc/net/response/UserInfo.kt#L6-L19

wrandall22: Our common spot for Java web is [here](https://github.com/CruGlobal/okta-client/blob/master/src/main/kotlin/org/cru/okta/client/model/user/OktaUser.kt#L6)

gyasistory (changes requested):
`Create Account` and `Login` send you to the same page.  So is this really needed?



image

levieggertcru:
Hey @gyasistory TheKey allowed us to choose if we wanted to land on create-account vs sign-in.  Okta does not.  Ann is aware of this and we decided to land on sign-in regardless which option is tapped.  Does GodTools Android do something differently for the menu here?

gyasistory (approved):

PR #964, opened 2021-12-14
GT-1310: Setup Parallel Language

reldredge71:
Create a modal for selecting a parallel language when landing on the tools list for the first time.


godtools/App/Features/Tools/ToolsMenuView.swift
```diff
@@ -53,6 +53,8 @@ class ToolsMenuView: UIViewController {
             target: self,
             action: #selector(handleMenu(barButtonItem:))
         )
+        
+        viewModel.openSetupParallelLanguageifAvailable()
```
levieggertcru, line 57: Hey @reldredge71 I don't think we want to present the setup parallel language anytime this view loads.  I noticed each time I relaunch the app the setup parallel language is presented over the tools.
Can this be part of the OnboardingFlow.  Similar to how quick actions is part of the onboarding flow.  It would look smoother if this were pushed onto the onboarding flow navigation stack when leaving quick actions.

godtools/App/Features/SetupParallelLanguage/Views/ParallelLanguageModal.swift
```diff
@@ -0,0 +1,138 @@
...
+//  ParallelLanguageModal.swift
+//  godtools
+//
+//  Created by Robert Eldredge on 12/10/21.
+//  Copyright © 2021 Cru. All rights reserved.
+//
+
+import UIKit
+
+class ParallelLanguageModal: UIViewController {
```
levieggertcru, line 11: Hey @reldredge71 I'm wondering if we can do something different for this view.  To me it just doesn't look that great, too small and hard to navigate.

I was thinking, it would be nice to have a language picker view that can be shared because in the future there will be a capability for picking a language once inside of a tract.  It would be nice to present this view from a flow and have a callback with the picked language and the flow can do whatever it needs to from there with the picked language.  My thoughts for this view is having a full screen list of languages with search bar at the top for filtering.  Pretty much similar to the choose language view.  What are your thoughts on that?

language-picker

levieggertcru (changes requested):

godtools/App/Features/SetupParallelLanguage/Views/ParallelLanguageModal.swift
reldredge71: That would be nice, and I could re-use the screens we already have rather than making something new.  But that is up to Ann and Justin.  The Figma design shows a different interface.

levieggertcru (changes requested):

godtools/App/Features/SetupParallelLanguage/SetupParallelLanguageView.xib
```diff
@@ -0,0 +1,159 @@
...
+                <outlet property="view" destination="3AB-3x-2v2" id="Ont-NA-60H"/>
+                <outlet property="yesButton" destination="WW4-C7-5Sn" id="2d6-ZV-k2B"/>
+            </connections>
+        </placeholder>
+        <placeholder placeholderIdentifier="IBFirstResponder" id="-2" customClass="UIResponder"/>
+        <view contentMode="scaleToFill" id="3AB-3x-2v2" userLabel="Setup Parallel Language View">
+            <rect key="frame" x="0.0" y="0.0" width="414" height="896"/>
+            <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMaxY="YES"/>
+            <subviews>
+                <stackView opaque="NO" contentMode="scaleToFill" axis="vertical" distribution="fillEqually" translatesAutoresizingMaskIntoConstraints="NO" id="2Iz-0h-kZx">
```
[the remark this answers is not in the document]
reldredge71: Hmm those are good thoughts.  I made sure the bottom buttons ("yes", "no", "get started") matched the rest of onboarding, but I didn't think to do that for the animation and title.  I can make those changes now, and find some other way to position the Select a Language button.

godtools/App/Features/SetupParallelLanguage/SetupParallelLanguageView.swift
```diff
@@ -0,0 +1,148 @@
...
+        
+        noButton.setTitle(viewModel.noButtonText, for: .normal)
+        
+        getStartedButton.setTitle(viewModel.getStartedButtonText, for: .normal)
+        
+        setupSelectLanguageButton()
+        setupBottomButtons()
+    }
+    
+    func setupSelectLanguageButton() {
```
levieggertcru, line 103: Hey @reldredge71 this method can be private.

godtools/App/Features/SetupParallelLanguage/SetupParallelLanguageView.swift
```diff
@@ -0,0 +1,148 @@
...
+        
+        selectLanguageButtonView.layer.shadowColor = UIColor(red: 0, green: 0, blue: 0, alpha: 1).cgColor
+        selectLanguageButtonView.layer.shadowOffset = CGSize(width: 2, height: 2)
+        selectLanguageButtonView.layer.shadowOpacity =  0.25
+        selectLanguageButtonView.layer.shadowRadius = 4
+        selectLanguageButtonView.layer.masksToBounds = false
+        selectLanguageButtonView.layer.cornerRadius = 6
+    }
+    
+    func setupBottomButtons() {
```
levieggertcru, line 113: This method can be private too.

godtools/App/Features/SetupParallelLanguage/SetupParallelLanguageView.swift
```diff
@@ -0,0 +1,148 @@
...
+        yesButton.layer.cornerRadius = 6
+        
+        noButton.layer.cornerRadius = 6
+        noButton.drawBorder(color: ColorPalette.gtBlue.color)
+        noButton.layer.borderWidth = 1
+        
+        getStartedButton.layer.cornerRadius = 6
+    }
+    
+    @objc func handleSelectLanguageTapped() {
```
levieggertcru, line 124: The @objc methods can be marked private as well.  Typically make everything private unless it needs outside access.

godtools/App/Features/SetupParallelLanguage/Views/ParallelLanguageModal.swift
```diff
@@ -0,0 +1,141 @@
...
+        viewModel.numberOfLanguages.addObserver(self) { [weak self] (numberOfLanguages: Int) in
+            self?.languagesTableView.reloadData()
+        }
+        
+        viewModel.selectedLanguageIndex.addObserver(self) { [weak self] (index: Int?) in
+            self?.languagesTableView.reloadData()
+        }
+    }
+    
+    @objc func handleBackgroundTapped() {
```
levieggertcru, line 81: Make these @objc methods private.

levieggertcru (changes requested):

levieggertcru (approved):

PR #967, opened 2021-12-15
GT-1373 and GT-1342 Fix: Handle Flow Completion Steps inside Flows

reldredge71:


godtools/App/Flows/TractFlow.swift
```diff
@@ -180,7 +173,7 @@ class TractFlow: NSObject, ToolNavigationFlow, Flow {
         
         case .homeTappedFromTool(let isScreenSharing):
             
-            if isScreenSharing && isShowingInitialTract {
+            if isScreenSharing && parentFlowIsHomeFlow {
```
levieggertcru, line 176: Hey Robert, I'm kind of curious what would happen if there was a screen share from a tract that was deep linked into from another tract.  If we'd want this check here or not, we still might want to show that alert.  ```parentFlowIsHomeFlow``` What does Ann suggest for that scenario?

levieggertcru (approved):

levieggertcru (approved):

PR #973, opened 2021-12-29
Gt add completion fix for dismiss animated false

levieggertcru:
Adding helper extension for dismissing a viewController and using the completion block.  I believe if animated is false, the UIKit dismiss won't fire the completion block.


godtools/App/Share/Extensions/UIViewController+DismissPresented.swift
```diff
@@ -0,0 +1,32 @@
...
+//  Copyright © 2021 Cru. All rights reserved.
+//
+
+import UIKit
+
+extension UIViewController {
+    
+    // Helper ensures completion will always be called.  I believe if animated is false when calling UIKit's UIViewController.dismiss(animated:) the completion closure will never be called.
+    
+    func dismissPresented(animated: Bool, completion: @escaping (() -> Void)) {
```
reldredge71, line 15: `dismissPresented` won't get called automatically, right?  We'll have to trigger this ourselves?

levieggertcru: Hey Robert yeah that's right, we'd have to call this extension.

reldredge71: Good to know, thanks!

reldredge71 (approved):

PR #976, opened 2022-01-03
GT-1367: Apply textScale to MultiplatformContentButton

reldredge71:


godtools/App/Services/Renderer/Views/MobileContent/Views/ContentButton/MobileContentButtonView.swift
```diff
@@ -137,6 +136,10 @@ extension MobileContentButtonView {
             relatedBy: .greaterThanOrEqual,
             priority: 1000
         )
+        
+        let scaledFont: UIFont = viewModel.getScaledFont()
```
levieggertcru, line 140: Hey @reldredge71 I noticed the viewModel has ```font``` and ```getScaledFont()``` and the view is accessing both of these.  We only need one of these and the view doesn't care about the details or specifics, just that a font is provided by the viewModel for it to consume.  Those details are hidden in the viewModel.  So just do any manipulation you need to the ```font``` computed property.

levieggertcru (changes requested):

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentButton/MobileContentButtonView.swift
reldredge71: I agree with that, I'll simplify it that way.  I was getting inspiration from `MobileContentTextViewModel`, which has `font` as well as `getScaledFont()`.   This might be because `MobileContentTextView` gets implemented inside of other classes (like `MobileContentHeaderView` for example), with `font` and `getScaledFont()` used in different places.  Maybe having two properties was necessary in that case, but since it's not necessary here I'll make the change.

levieggertcru: Hey @reldredge71 hmm, let me check that, I wonder if we need to update that.

levieggertcru: Okay yeah there are different fonts depending on the view.  I may need to update that.  This PR looks good to me.

levieggertcru (approved):

PR #980, opened 2022-01-04
GT-1386 add shared request operation package

levieggertcru:
Removed all the RequestOperation classes and generic json decoding logic into the shared request-operation-ios swift package.


godtools/App/Services/Articles/Repository/Downloader/ArticleAemDownloadOperation.swift
```diff
@@ -7,6 +7,7 @@
 //
 
 import Foundation
+import RequestOperation
```
gyasistory, line 10: @levieggertcru why do you need the import if now code is added?

godtools/App/Services/Articles/Repository/Downloader/ArticleAemDownloader.swift
```diff
@@ -8,6 +8,7 @@
 
 import Foundation
 import RealmSwift
+import RequestOperation
```
gyasistory, line 11: @levieggertcru same question here?

godtools/App/Services/EmailSignUp/EmailSignUpService.swift
```diff
@@ -7,6 +7,7 @@
 //
 
 import Foundation
+import RequestOperation
```
gyasistory, line 10: @levieggertcru again here

reldredge71 (approved):

godtools/App/Share/ViewModels/TranslationDownloaderErrorViewModel.swift
```diff
@@ -7,6 +7,7 @@
 //
 
 import Foundation
+import RequestOperation
```
gyasistory, line 10: here also

gyasistory (changes requested):
Not sure if the imports with no changes are needed.

godtools/App/Services/Articles/Repository/Downloader/ArticleAemDownloadOperation.swift
levieggertcru: Hey @gyasistory that code is now in a shared framework (pulled in via swift package manager) here -> https://github.com/CruGlobal/request-operation-ios.

PR #982, opened 2022-01-11
GT-1251 Removed deprecated xml node renderer

levieggertcru:
- Removed all swift files related to the old xml node renderer because we replaced it with the shared Multiplatform :)


wrandall22:
Looks like some other things got intermingled with this commit (e.g. changing types). Maybe it should have been 2 commits? Or is another branch's work sneaking in here?

levieggertcru:
Could have been broken into a couple commits in that case.

reldredge71 (approved):

PR #983, opened 2022-01-13
GT-1381: Fix right-to-left text in Tools

reldredge71:
Ensure that all text elements render as right-to-left if the primary language is right-to-left.  Also flip the number and the title in the ToolPageHeader for right-to-left languages.


wrandall22:
I like the clear, concise commits.

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentText/MobileContentTextViewModel.swift
```diff
@@ -77,16 +79,34 @@ class MobileContentTextViewModel: MobileContentTextViewModelType {
...
-        let modelTextAlignment: MobileContentTextAlignment? = textModel.textAlignment ?? containerModel?.textAlignment
+        
+        let textModelAlignment = textModel.textAlignment
+        let containerModelAlignment = containerModel?.textAlignment
+        
+        //first determine the desired alignment for left-to-right languages
+        var modelTextAlignment = textModelAlignment ?? containerModelAlignment
+        
+        //if primary language is right-to-left, flip the alignment
+        if languageTextAlignment == .right {
```
levieggertcru, line 90: Hey @reldredge71 this ```languageTextAlignment``` is actually the alignment based on the language direction for the current rendered language.  So if we are viewing the parallel language, then this alignment is based off of that parallel language direction.  It's not related to the primary language in anyway making the above comment not accurate and maybe not what is intended here.

godtools/App/Services/Renderer/Views/Tool/Views/Header/ToolPageHeaderView.swift
```diff
@@ -45,8 +45,10 @@ class ToolPageHeaderView: MobileContentView {
         
         backgroundColor = .clear
         backgroundView.backgroundColor = viewModel.backgroundColor
+        
+        backgroundView.semanticContentAttribute = viewModel.languageDirectionSemanticContentAttribute
```
levieggertcru, line 49: That fixed it!  For some reason it doesn't seem to work anymore when setting directly on the ToolPageHeaderView.  I'm not sure why that is.  Better to just apply to both of those views.

levieggertcru (changes requested):

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentText/MobileContentTextViewModel.swift
reldredge71: @levieggertcru I think I'm following.  I've been forgetting about rendering the parallel language.  So my desired effect would be to first assess what the XML objects state the alignment should be, and then take into account the language currently being rendered and flip the alignment if it is a right-to-left language.  So if the XML says left-aligned and the primary language is left-to-right, then all primary language text should stay left-aligned; if the parallel language is right-to-left, then all parallel language text should be right-aligned.  Is my logic still good, and I need to update the comment?  Or is it my logic that needs to be changed?

godtools/App/Services/Renderer/Views/Tool/Views/Header/ToolPageHeaderView.swift
reldredge71: Yeah it must be treating the backgroundView as the parent of the title and number, not the ToolPageHeaderView.

godtools/App/Services/Renderer/Renderer/Multiplatform/MobileContentMultiplatformRenderer.swift
```diff
@@ -89,7 +89,7 @@ class MobileContentMultiplatformRenderer: MobileContentRendererType {
                 
         for childModel in childModels {
             
-            let childMobileContentView: MobileContentView? = recurseAndRender(renderableModel: childModel, renderableModelParent: renderableModel, rendererPageModel: rendererPageModel, containerModel: containerModel)
+            let childMobileContentView: MobileContentView? = recurseAndRender(renderableModel: childModel, renderableModelParent: renderableModel, rendererPageModel: rendererPageModel, containerModel: containerModel, primaryLanguage: primaryLanguage)
```
levieggertcru, line 92: Don't need primaryLanguage here either.

godtools/App/Services/Renderer/Renderer/Multiplatform/MobileContentMultiplatformRenderer.swift
```diff
@@ -67,15 +67,15 @@ class MobileContentMultiplatformRenderer: MobileContentRendererType {
...
+        if let renderableView = recurseAndRender(renderableModel: pageModel, renderableModelParent: nil, rendererPageModel: rendererPageModel, containerModel: nil, primaryLanguage: primaryRendererLanguage) {
             return .success(renderableView)
         }
                 
         let failedToRenderPageError: Error = NSError(domain: "", code: -1, userInfo: [NSLocalizedDescriptionKey: "Failed to render page."])
         return .failure(failedToRenderPageError)
     }
     
-    private func recurseAndRender(renderableModel: MobileContentRenderableModel, renderableModelParent: MobileContentRenderableModel?, rendererPageModel: MobileContentRendererPageModel, containerModel: MobileContentRenderableModelContainer?) -> MobileContentView? {
+    private func recurseAndRender(renderableModel: MobileContentRenderableModel, renderableModelParent: MobileContentRenderableModel?, rendererPageModel: MobileContentRendererPageModel, containerModel: MobileContentRenderableModelContainer?, primaryLanguage: LanguageModel) -> MobileContentView? {
```
levieggertcru, line 78: Don't need to pass primary language here as it is available in MobileContentRenderableModel.

levieggertcru (changes requested):

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentText/MobileContentTextViewModel.swift
reldredge71: I can ask, but I think it should flip based on how someone would expect to read the text.  I imagine that a reader of a right-to-left language would expect to read things right-aligned in most cases.  I can still ask though.

reldredge71: https://jira.cru.org/browse/GT-1381

It looks like they would like the alignment to be based on the currently rendered langauge

levieggertcru: Thanks @reldredge71, yeah that makes more sense like you said.

levieggertcru (approved):

PR #986, opened 2022-01-19
GT-1407 renderer download image fallback

levieggertcru:
Adding some additional logic to the renderer that will download image attachments if for some reason we are unable to fetch them from the cache.  This is could happen if a user opens a tool from a deep link because we sync all the attachments on first launch, so it is possible they go into a tool while that sync is happening and then land on a page with at image that hasn't been downloaded and cached yet.
This update will introduce some additional logic to download any non cached images from the remote database and display those in the renderer.


levieggertcru:
This PR is not valid.

PR #991, opened 2022-01-24
GT-1310 Fix: Present SelectParallelLanguage as Modal

reldredge71:
Move SelectParallelLanguage view from OnboardingFlow to ToolsFlow, triggered when landing on the ToolsFlow for the first time.  It is presented as a modal now instead of a full screen.  Also add functionality for the darkened background behind the modal to be pressed in order to exit the modal.


godtools/App/Features/SetupParallel/SetupParallelLanguageView.swift
```diff
@@ -6,19 +6,19 @@
 //  Copyright © 2021 Cru. All rights reserved.
 //
 
-import Foundation
 import UIKit
 
-class SetupParallelLanguageView: UIViewController {
+class SetupParallelLanguageView: UIViewController, UIGestureRecognizerDelegate {
```
levieggertcru, line 11: Hey @reldredge71 I have a functionality question on this one.  If I tap a language from the list, and then tap the overlays to close out the setup parallel language modal, my selection will be saved to settings.  I'm wondering if we only want to save that selection if the Get Started button is tapped.  In my head I was thinking tapping the overlay would cancel out, but maybe some people would want the option to stick.  Could you check with Ann and Daniel on this one and see when that setting should be saved.  If at the moment of tapping it in the list, or when Get Started is tapped.

godtools/App/Features/SetupParallel/SetupParallelLanguageView.swift
```diff
@@ -45,63 +45,66 @@ class SetupParallelLanguageView: UIViewController {
         
         super.viewDidLoad()
         
+        tapRecognizer = UITapGestureRecognizer(target: self, action: #selector(handleSelectLanguageTapped))
+        tapRecognizer.cancelsTouchesInView = false
+        tapRecognizer.delegate = self
+        self.view.addGestureRecognizer(tapRecognizer)
```
levieggertcru, line 51: I noticed when I tapped No, the language list appeared and then the entire modal was dismissed.  It looks like because this tap gesture is being applied to the entire view.  Instead it should be applied to the selectLanguageButtonView.

godtools/App/Features/SetupParallel/SetupParallelLanguageView.swift
levieggertcru, line 11: I don't see any UIGestureRecognizerDelegate methods being used in this class. https://developer.apple.com/documentation/uikit/uigesturerecognizerdelegate

No need to add the delegate for UITapGestureRecognizer.

godtools/App/Features/SetupParallel/SetupParallelLanguageView.swift
```diff
@@ -45,63 +45,66 @@ class SetupParallelLanguageView: UIViewController {
         
         super.viewDidLoad()
         
+        tapRecognizer = UITapGestureRecognizer(target: self, action: #selector(handleSelectLanguageTapped))
+        tapRecognizer.cancelsTouchesInView = false
+        tapRecognizer.delegate = self
```
levieggertcru, line 50: No need to add the delegate, not being used.

godtools/App/Features/SetupParallel/SetupParallelLanguageView.swift
```diff
@@ -45,63 +45,66 @@ class SetupParallelLanguageView: UIViewController {
...
+        self.view.addGestureRecognizer(tapRecognizer)
+        
         setupLayout()
         setupBinding()
     }
     
+    override func viewWillDisappear(_ animated: Bool) {
+        
+        self.view.removeGestureRecognizer(self.tapRecognizer)
+        self.tapRecognizer.delegate = nil
```
levieggertcru, line 60: Delegate not being used.

godtools/App/Features/SetupParallel/SetupParallelLanguageView.swift
```diff
@@ -45,63 +45,66 @@ class SetupParallelLanguageView: UIViewController {
...
+        tapRecognizer.delegate = self
+        self.view.addGestureRecognizer(tapRecognizer)
+        
         setupLayout()
         setupBinding()
     }
     
+    override func viewWillDisappear(_ animated: Bool) {
+        
+        self.view.removeGestureRecognizer(self.tapRecognizer)
```
levieggertcru, line 59: Is removing the gesture doing anything?  I think if it's removed it won't get added back because it is added on viewDidLoad.  If you do need to enable / disable a gesture that can also be done using the isEnabled property. https://developer.apple.com/documentation/uikit/uigesturerecognizer/1624220-isenabled

godtools/App/Features/SetupParallel/SetupParallelLanguageView.swift
```diff
@@ -6,19 +6,19 @@
...
+    
+    private var tapRecognizer: UITapGestureRecognizer!
+    
     @IBOutlet weak private var animatedView: AnimatedView!
     @IBOutlet weak private var promptLabel: UILabel!
     @IBOutlet weak private var selectLanguageButtonView: UIView!
-    @IBOutlet weak private var selectLanguageButtonGestureRecognizer: UITapGestureRecognizer!
     @IBOutlet weak private var selectLanguageButtonTitle: UILabel!
     @IBOutlet weak private var yesButton: UIButton!
     @IBOutlet weak private var noButton: UIButton!
```
levieggertcru, line 24: Can you make sure all of these button types are custom type in the .xib.  I think this one is System.  Want to ensure we keep those as custom types otherwise iOS versions could change the appearance of them.

godtools/App/Features/SetupParallel/SetupParallelLanguageView.swift
```diff
@@ -45,63 +45,66 @@ class SetupParallelLanguageView: UIViewController {
...
+        setupBottomButtons()
+    }
+    
     private func setupBinding() {
         
         animatedView.configure(viewModel: viewModel.animatedViewModel)
         
-        selectLanguageButtonGestureRecognizer.addTarget(self, action: #selector(handleSelectLanguageTapped))
-        
         yesButton.addTarget(self, action: #selector(handleYesTapped), for: .touchUpInside)
```
levieggertcru, line 81: Move addTargets into the viewDidLoad() just because we keep the setupBinding() relative to the viewModel.

godtools/App/Share/Views/TransparentModal/TransparentModalView.swift
```diff
@@ -68,6 +75,16 @@ class TransparentModalView: UIViewController {
         // modalView
         addModalView(modalView: modalView)
     }
+    
+    private func setupBinding() {
+        
+        overlayButton.addTarget(self, action: #selector(handleBackgroundTapped), for: .touchUpInside)
```
levieggertcru, line 81: Move into the viewDidLoad and keep the setupBinding() for any viewModel binding.

levieggertcru (changes requested):

godtools/App/Features/SetupParallel/SetupParallelLanguageView.swift
reldredge71: I can ask, but I think what I have is what's expected.

godtools/App/Features/SetupParallel/SetupParallelLanguageView.swift
reldredge71: I noticed that at one point, but I thought it went away.  I can try to make that change.

godtools/App/Features/SetupParallel/SetupParallelLanguageView.swift
reldredge71: I was just doing this for cleanup when the view gets closed.  I don't think I need to enable or disable it, I just want to properly remove the gesture recognizer when the screen gets closed.

levieggertcru: In that case you'd want to use the deinit to clean that up.  viewWillDisappear will get called if something is presented over this view, or if navigating forward to a view from a push etc. which would cause the tap gesture to get removed and then never added back.

godtools/App/Features/SetupParallel/SetupParallelLanguageView.swift
levieggertcru: Thanks Robert.

reldredge71: Justin said that he would like the parallel language to be set even if the user doesn't press "Get Started".  Doing it the way I have it could also give the app a couple more seconds to pre-load the parallel languages for the first few tools.  I know that the Favorited tools will download the primary language automatically when landing on the Tools Flow, is the same true for parallel language?

godtools/App/Features/SetupParallel/SetupParallelLanguageView.swift
```diff
@@ -6,19 +6,19 @@
...
-    private var closeButton: UIBarButtonItem?
-
+    private let buttonCornerRadius: CGFloat = 6
+    
+    private var tapRecognizer: UITapGestureRecognizer!
+    
     @IBOutlet weak private var animatedView: AnimatedView!
     @IBOutlet weak private var promptLabel: UILabel!
     @IBOutlet weak private var selectLanguageButtonView: UIView!
-    @IBOutlet weak private var selectLanguageButtonGestureRecognizer: UITapGestureRecognizer!
```
levieggertcru, line 21: Hey @reldredge71 instead of having this declared with the forced unwrap ```!``` can you instead make it immutable and allocate this when the view is initialized. I don't think you will be able to set the target and action on initialization of UITapGestureRecognizer, but take a look at the UITapGestureRecognizer addTarget method to do that.

https://developer.apple.com/documentation/uikit/uigesturerecognizer/1624230-addtarget

godtools/App/Features/SetupParallel/SetupParallelLanguageViewModel.swift
```diff
@@ -48,6 +48,8 @@ class SetupParallelLanguageViewModel: NSObject, SetupParallelLanguageViewModelTy
         
         super.init()
         
+        setupParallelLanguageAvailability.markSetupParallelLanguageViewed()
+        
         reloadData()
         
         setupBinding()
```
levieggertcru, line 55: I didn't catch this before, but when this viewModel is deallocated can you remove this observer. languageSettingsService.parallelLanguage

levieggertcru (changes requested):

godtools/App/Features/SetupParallel/SetupParallelLanguageView.swift
reldredge71: @levieggertcru I'm a little confused.  Are you talking about `selectLanguageButtonView`?  Should it not be an IBOutlet?  Or are you talking about the tapRecognizer?  If it's the tapRecognizer, then I could probably implement it similar to here: https://github.com/CruGlobal/godtools-swift/blob/develop/godtools/App/Services/Renderer/Views/MobileContent/Views/ContentImage/MobileContentImageView.swift#L26

levieggertcru: Hey @reldredge71 I was talking about the tapGesture.  IBOutlets you have to declare unwrapped ```!```, but anything else it's good practice not to declare with the ```!```.

levieggertcru (approved):

PR #996, opened 2022-01-31
GT-1394: Fix Bottom Gradient in Tool Cards

reldredge71:
There were two parts to the issue I found:
1. The gradient is initially being set before the content insets between the CardsView and the Card are set.  So the width of the gradient is the full with of the CardsView, not the new shrunken width of the Card.
2. Multiple gradient layers are getting layered on top of each other each time `setupLayout()` is getting stepped through.  So, even when the content insets are applied to the card and I call `layoutSubviews()`, the original gradient still exists.

My solution is to:
1. Make sure the gradient is getting set after the content insets are in place, and
2. If there is a gradient layer already in place, replace it instead of adding a new one.

### DIFF
--- a/godtools/App/Services/Renderer/Views/Tool/Views/Card/ToolPageCardView.swift
+++ b/godtools/App/Services/Renderer/Views/Tool/Views/Card/ToolPageCardView.swift
@@ -143,6 +143,7 @@ class ToolPageCardView: MobileContentView {
         let rootView: UIView? = subviews.first
         rootView?.layer.cornerRadius = cardCornerRadius
         
+        // bottom gradient
         bottomGradientView.isUserInteractionEnabled = false
         bottomGradientView.backgroundColor = .clear
         bottomGradientLayer.frame = bottomGradientView.bounds
@@ -185,7 +186,6 @@ class ToolPageCardView: MobileContentView {
     
     func setDelegate(delegate: ToolPageCardViewDelegate?) {
         self.delegate = delegate
-        
     }
     
     func onCardVisible() {
--- a/godtools/App/Services/Renderer/Views/Tool/Views/Card/ToolPageCardView.swift
+++ b/godtools/App/Services/Renderer/Views/Tool/Views/Card/ToolPageCardView.swift
@@ -93,6 +93,11 @@ class ToolPageCardView: MobileContentView {
         }
     }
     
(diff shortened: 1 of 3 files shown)


godtools/App/Services/Renderer/Views/Tool/Views/Card/ToolPageCardView.swift
```diff
@@ -177,8 +169,29 @@ class ToolPageCardView: MobileContentView {
         nextButton.isHidden = viewModel.hidesNextButton
     }
     
+    private func setupBottomGradientView() {
+        
+        bottomGradientView.isUserInteractionEnabled = false
+        bottomGradientView.backgroundColor = .clear
+        let bottomGradient = CAGradientLayer()
```
levieggertcru, line 176: Hey @reldredge71 ahh what's happening is when the bottomGradientView's bounds change (due to autolayout constraints) the added ```bottomGradient: CAGradientLayer``` doesn't update because constraints aren't applied to that CALayer.

What you have works, but it would be more optimized to just keep a reference to the added CAGradientLayer and then in layoutSubviews() method update the frame of the CAGradientLayer to match bounds of the bottomGradientView.

Also don't forget in the override layoutSubviews() to call super.

levieggertcru (changes requested):

godtools/App/Services/Renderer/Views/Tool/Views/Cards/ToolPageCardsView.swift
```diff
@@ -353,6 +353,8 @@ extension ToolPageCardsView {
             cardTopConstraints.append(top)
             
             cardView.setDelegate(delegate: self)
+            
+            cardView.layoutSubviews()
```
levieggertcru, line 357: Hey @reldredge71 one last thing, I'm pretty sure you don't need this call here. layoutSubviews() will get called by the system when the constraints layout. 
You also don't want to call layoutSubviews() directly according to Apple's documentation.  Instead you'd want to call either setNeedsLayout() or layoutIfNeeded() depending on your needs.  https://developer.apple.com/documentation/uikit/uiview/1622482-layoutsubviews

levieggertcru (changes requested):

levieggertcru (approved):

PR #1000, opened 2022-02-03
GT-1310 Fixes: New Animation & Trigger for SelectParallelLanguage

reldredge71:
Some QA fixes for GT-1310:
- The animation on the SetupParallelLanguage modal was wrong.  I forgot that I put a placeholder there until I had the real one.
- The flow from the Articles quick start link was not working correctly.  For one thing, the download screen seemed to reach 99% but would not close on its own.  For another thing, the expected behavior was that the SetupParallelLanguage modal would pop up would appear when the user leaves the Articles tool and lands on the Favorite Tools list for the first time.  In reality, it would not appear at all.  The reason for both of these issues, I think, was that the ToolsFlow was attempting to present the SetupParallelLanguage modal on `init`, but the Articles Quick Start link attempts to open the Questions About God tool immediately after.  To fix this, I removed the `showSetupParallelLanguage` flow step from `init` and triggered it separately from the `showTools` flow step.  In the case of the Articles flow, I delay triggering `showSetupParallelLanguage` until `articleFlowCompleted` is triggered.

Let me know if you have a different idea of how I should handle this.


levieggertcru:
Hey @reldredge71 yeah something strange is happening there.  I know too since that download article modal will get presented and the setup parallel is also presented one will override the other since there can only be 1 presentation at a time.  This is kind of tricky.  I'm going to approve this PR and do some more thinking on this one.

levieggertcru (approved):

reldredge71:
Yeah it is odd.  It sounds like Ann and Jill don't want the Setup Parallel Language modal to be presented until the user closes the Questions about God tool anyway.  So it needs to be delayed in that flow.  Thank you for approving; maybe there's a more robust solution, but what I have seems to work fine.

PR #1003, opened 2022-02-04
GT-1421 Don't include the private use subtag when generating a language display name

frett:


levieggertcru (approved):

frett:
I'm assuming all display of language names goes through this ViewModel. If not then a similar patch should be made in the other locations, or even better the other locations should be updated to reference the LanguageViewModel

PR #1005, opened 2022-02-07
GT-1281 conversation starter tool

levieggertcru:
This PR adds the new conversation starter tool which can be accessed from the All Tools list.  In this PR the following was added:
- Flow Layout (MobileContentFlowView, MobileContentFlowItemView, and MobileContentFlowRow)
- Card Collection Page (MobileContentCardCollectionPageView and MobileContentCardCollectionPageCardView)
- Content Page (MobileContentPageView)
- Content Card (MobileContentCardView)
- Content Buttons and Content Images support both fixed widths and percentage widths of container through the new Dimension type.  Added MobileContentViewWidth enum to wrap the Dimension type.
- Added shared NavBarSelectorView which will replace the language toggle in the navigation bar.
- Added HorizontallyCenteredCollectionViewLayout for rendering card collection pages so they are centered in the device screen with some visibility on next and previous pages.


levieggertcru:
Just a heads up, I created a few additional tasks that will need to be completed for the conversation starter tool (choose your own adventure).  These tasks here.
https://jira.cru.org/browse/GT-1450
https://jira.cru.org/browse/GT-1451
https://jira.cru.org/browse/GT-1452

godtools/App/Services/Renderer/Views/MobileContent/Views/CardCollectionPageCard/MobileContentCardCollectionPageCardView.swift
```diff
@@ -14,26 +14,62 @@ class MobileContentCardCollectionPageCardView: MobileContentView {
...
     override func renderChild(childView: MobileContentView) {
         super.renderChild(childView: childView)
                 
         if let contentStackView = childView as? MobileContentStackView {
             addContentStackView(contentStackView: contentStackView)
         }
     }
+    
+    override func finishedRenderingChildren() {
+        super.finishedRenderingChildren()
```
reldredge71, line 67: Is this override needed, if all it does is call `super.finishedRenderingChildren()`?

reldredge71 (commented):
This is a lot of work!  And a lot of commits to go through.  Honestly not much seems out of place or questionable to me; it's hard to follow all of the resizing calculations across all the pages, but in general I trust that you've tested it all.  Is there anything on staging that I could look at that tests these render objects?

frett:
@reldredge71 you should be able to interact with the Openers tool on production with this branch. That covers probably 90% of the use cases. It just won't cover some corner cases and parallel language toggling.

There is also a CYOA test tool and preliminary version of conversation starters on stage that test some of those corner cases.

some of the new content elements are captured in the renderer test tool as well

godtools/App/Services/Renderer/Views/MobileContent/Views/CardCollectionPageCard/MobileContentCardCollectionPageCardView.swift
levieggertcru: Hey @reldredge71 the override isn't needed here.  I'll remove that.

reldredge71:
@frett @levieggertcru I explored those three tools -- Openers on prod, CYOA test tool on stage, and Conversation Starter on stage.  I noticed a couple of things.

1. On the initial screen, there seems to be no padding for the text, it just expands to the edge of the screens.
2. On the CYOA test, the "play" and "stop" buttons don't seem to do anything, also I'm not seeing any way to progress further than the screen with the Apple/Android animation (if there is anything).
3. On the Conversation Start preliminary tool, pressing the "Let's do it" button brings up the following error:

Screen Shot 2022-02-09 at 10 39 43 AM

levieggertcru:
Hey @reldredge71 thank you for testing these.  I'll take a look into 2 and 3.  1 will get resolved in GT-1450.

levieggertcru:
Hey @frett is the play and stop buttons for the animation?

levieggertcru:
Hey @reldredge71 I'm going to create a separate Jira ticket for item 2 as well.  I just fixed that error in item 3.  When the flow view was getting laid out again due to a bounds change, I wasn't clearing out the existing array of flow item views.  So any layout change the existing array would increase by whatever was currently in there.

reldredge71 (approved):
Thank you for making that fix!  Looks good now!

PR #1010, opened 2022-02-08
GT-1371 Add download OneSky localizations workflow

levieggertcru:
Added workflow that will run once per day, download onesky localizations, and create a pull request.

### DIFF
--- a/.github/workflows/download_onesky_translations.yml
+++ b/.github/workflows/download_onesky_translations.yml
@@ -40,5 +40,6 @@ jobs:
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
         with:
+          branch: 'latest-onesky-translations'
           token: ${{ secrets.GITHUB_TOKEN }}
           title: 'Latest OneSky Translations'


.github/workflows/download_onesky_translations.yml
```diff
@@ -1,36 +1,44 @@
...
+        env:
+          ONESKY_PUBLIC_KEY: ${{ secrets.ONESKY_PUBLIC_KEY }}
+          ONESKY_SECRET_KEY: ${{ secrets.ONESKY_SECRET_KEY }}
+        run: bundle exec fastlane cru_shared_lane_download_and_commit_latest_one_sky_localizations
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          title: 'Latest OneSky Translations'
```
frett, line 44: I would include the `branch` option for this step as well. I'd like a more descriptive branch name instead of the generic `create-pull-request/patch`

.github/workflows/build.yml
```diff
@@ -45,13 +45,6 @@ jobs:
       - name: Run Tests
         run: bundle exec fastlane cru_shared_lane_run_tests
 
-      - name: Download And Commit Latest OneSky Localizations
-        if: github.event_name == 'push' && github.ref == 'refs/heads/master'
-        env:
-          ONESKY_PUBLIC_KEY: ${{ secrets.ONESKY_PUBLIC_KEY }}
-          ONESKY_SECRET_KEY: ${{ secrets.ONESKY_SECRET_KEY }}
-        run: bundle exec fastlane cru_shared_lane_download_and_commit_latest_one_sky_localizations
-
```
frett, line 54: I would probably leave this in place to ensure test flight builds do have the latest translations.

.github/workflows/download_onesky_translations.yml
```diff
@@ -1,36 +1,44 @@
...
 
-# Controls when the workflow will run
 on:
-  # Triggers the workflow on push or pull request events but only for the develop branch
-  push:
-    branches: [ develop ]
-  pull_request:
-    branches: [ develop ]
+  schedule:
+    - cron: "0 23 * * *"
```
frett, line 5: there is also a trigger that allows you to manually trigger it from the GHA interface. We would probably want to include that with this workflow

.github/workflows/download_onesky_translations.yml
levieggertcru: Hey @frett changed the branch name to be more descriptive.

.github/workflows/download_onesky_translations.yml
```diff
@@ -1,36 +1,44 @@
...
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
```
levieggertcru, line 8: Hey @frett I do have the workflow_dispatch here which allows to manually trigger.

.github/workflows/build.yml
levieggertcru: Good thinking, added that back in.

.github/workflows/download_onesky_translations.yml
frett: ah, didn't read far enough in the workflow :)

frett (approved):

PR #1011, opened 2022-02-08
GT-1444 add swedish and finnish language localizations

levieggertcru:
- Added localization files to Xcode project from project settings
- Added language codes to list of OneSky localization files that are downloaded into the Xcode project

### DIFF
--- a/godtools/fi.lproj/Localizable.strings
+++ b/godtools/fi.lproj/Localizable.strings
@@ -1,0 +1,277 @@
+"language_onboard" = "80+ Languages";
+
+/* This is to indicate to the user that there are no additional tools or resources available for download because they have already downloaded everything. */
+"no_more_tools" = "You have downloaded all available tools.";
+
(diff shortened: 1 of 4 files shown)


godtools/fi.lproj/Localizable.strings
```diff
@@ -0,0 +1,277 @@
+"language_onboard" = "80+ Languages";
+
+/* This is to indicate to the user that there are no additional tools or resources available for download because they have already downloaded everything. */
+"no_more_tools" = "You have downloaded all available tools.";
```
reldredge71, line 4: I assume the translations will get loaded to these files when the Onesky download command is run in GHA?

levieggertcru: Hey @reldredge71 yeah I'm thinking these will get overwritten during the OneSky download.  Xcode initially creates them based off of the Base Localizable.strings.

reldredge71 (approved):

PR #1022, opened 2022-02-15
GT-1422 remove parser protocol and wrapper

levieggertcru:


godtools/App/Features/Tool/Views/ToolNavBar/ToolNavBarViewModel.swift
```diff
@@ -40,8 +40,8 @@ class ToolNavBarViewModel: NSObject, ToolNavBarViewModelType {
         self.hidesShareButton = hidesShareButton
         
         backButtonImage = backButtonImageType.getImage()
-        navBarColor = manifestAttributes.navbarColor?.uiColor ?? manifestAttributes.primaryColor.uiColor
-        navBarControlColor = manifestAttributes.navbarControlColor?.uiColor ?? manifestAttributes.primaryTextColor.uiColor
+        navBarColor = manifestAttributes.navbarColor?.uiColor ?? .darkGray
+        navBarControlColor = manifestAttributes.navbarControlColor?.uiColor ?? .lightGray
```
frett, line 44: I have constants defined for default colors, and most things in kotlin are setup to return those default colors if the manifest doesn't exist. I'm not 100% sure how all of that translates to iOS, but we can investigate to see if that prevents you from needing to define fallback colors like this

frett: we can also expose those fallback colors in other manners if necessary

frett: cool, for the most part I've tried to make all color/dimensions non-nullable, if you encounter one that isn't let me know and we can try and address it

levieggertcru:
@reldredge71 @gyasistory @frett this is ready for review now.  I was able to remove all of the protocols and most of the wrappers.  There are some areas where I still need a concrete class for a MobileContentRenderableModel.  This is due to the way rendering happens.  But, all of the viewModels now reference a GodToolsToolParser model instead of a protocol / wrapper.

godtools.xcodeproj/project.pbxproj.orig
```diff
@@ -47,14 +46,12 @@
 		450E44552756B6E000D4311E /* MenuItemViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 450E44522756B6E000D4311E /* MenuItemViewModel.swift */; };
 		450E44572756B6FC00D4311E /* MenuItemViewModelType.swift in Sources */ = {isa = PBXBuildFile; fileRef = 450E44562756B6FC00D4311E /* MenuItemViewModelType.swift */; };
 		450EF834247EB3D300978926 /* SF-Pro-Text-Bold.otf in Resources */ = {isa = PBXBuildFile; fileRef = 450EF833247EB3D300978926 /* SF-Pro-Text-Bold.otf */; };
-		450F94CB27A1EEF7009DEA43 /* CardCollectionPageCard+MobileContentRenderer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 450F94CA27A1EEF7009DEA43 /* CardCollectionPageCard+MobileContentRenderer.swift */; };
+		450F94CB27A1EEF7009DEA43 /* CardCollectionPageCard+MobileContentRenderableModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 450F94CA27A1EEF7009DEA43 /* CardCollectionPageCard+MobileContentRenderableModel.swift */; };
```
frett, line 49: I don't think think this file should actually be tracked in git, it is produced when there is a merge conflict on `project.pbxproj`. in fact it could be beneficial to define a `.gitignore` rule to ignore `*.orig`

levieggertcru: Good catch.  Just removed that.  Yeah I think that's a good idea to add that to the .gitignore.

godtools/App/Features/ChooseYourOwnAdventure/ChooseYourOwnAdventureViewModel.swift
```diff
@@ -28,16 +29,14 @@ class ChooseYourOwnAdventureViewModel: MobileContentPagesViewModel, ChooseYourOw
...
-            navBarColor = barColor
-            navBarControlColor = controlColor
+        if let manifest = manifest {
+            navBarColor = manifest.navBarColor
+            navBarControlColor = manifest.navBarControlColor
         }
         else {
-            
             navBarColor = ColorPalette.gtBlue.color
             navBarControlColor = .white
```
frett, line 41: hmm, we should talk about how you are able use the default colors that are present in the parser.

within Kotlin the `navBarColor` and `navBarControlColor` properties use a nullable receiver, so if my manifest is null they will resolve to the default colors. But I don't know if that concept is translated to Swift or if we will need to expose those defaults in another manner.

```kotlin
val manifest: Manifest? = null
manifest.navBarColor // this would return the default color because manifest is null
```

godtools/App/Features/ToolTraining/ToolTrainingViewModel.swift
```diff
@@ -30,12 +31,12 @@ class ToolTrainingViewModel: NSObject, ToolTrainingViewModelType {
...
     
-    required init(flowDelegate: FlowDelegate, renderer: MobileContentRendererType, trainingTipId: String, tipModel: TipModelType, analytics: AnalyticsContainer, localizationServices: LocalizationServices, viewedTrainingTips: ViewedTrainingTipsService) {
+    required init(flowDelegate: FlowDelegate, pageRenderer: MobileContentPageRenderer, renderedPageContext: MobileContentRenderedPageContext, tipPages: [TipPage], trainingTipId: String, tipModel: Tip, analytics: AnalyticsContainer, localizationServices: LocalizationServices, viewedTrainingTips: ViewedTrainingTipsService) {
         
         self.flowDelegate = flowDelegate
-        self.renderer = renderer
-        self.resource = renderer.resource
-        self.language = renderer.language
+        self.renderedPageContext = renderedPageContext
+        self.tipPages = tipPages
```
frett, line 38: any reason you are passing in the `tipPages` array instead of just accessing it out of the `Tip` model which you already have?

levieggertcru: Hey @frett good catch, I missed that one.

godtools/App/Services/Renderer/ManifestResourcesCache/ManifestResourcesCache.swift
```diff
@@ -7,25 +7,26 @@
 //
 
 import UIKit
+import GodToolsToolParser
 
-class ManifestResourcesCache: ManifestResourcesCacheType {
```
reldredge71, line 11: @levieggertcru could you remind me why you're removing the protocols?  I'm assuming it is because it is easier/cleaner to just reference the classes directly.  But it makes me curious, when is it preferable to use protocols and when is it preferable to not use them?  Should you only use them when you expect to have multiple classes that have the same properties?  Or when you anticipate switching out one service for another?  I'm curious what the best practices are when it comes to protocols.

levieggertcru: Hey @reldredge71 that's a good question.  I initially created those protocols when switching from the XML parser to GodToolsToolParser.  It was a big transition and I needed to be able to switch between both the XML parser and GodToolsTool parser.  I decided to remove the protocols for a couple of reasons. 
1. I don't see us needing to make GodToolsToolParser interchangeable just because we maintain it and Android, iOS, and Web are dependent on it.
2. It created a lot of extra files and overhead.  For every type in the GodToolsToolParser we'd need to create 2 additional files (the protocol and either an extension or concrete wrapper class to implement the protocol on GodToolsToolParser models).  This created additional overhead when parsing models to view / viewModel, not sure how much it impacted rendering performance, but it could impact it some.

Protocols are suited for abstraction.  See the Interface Segregation Principle ( https://reflectoring.io/interface-segregation-principle/#:~:text=The%20Interface%20Segregation%20Principle%20(ISP,%E2%80%9D%20or%20%E2%80%9Cfat%E2%80%9D%20interface )
They're also good to make things interchangeable.  In Clean Architecture they talk about using them to make the data layer interchangeable with the domain layer. 
They also act similarly as a header file in Objective-C.  We've been using them for our ViewModels which make it easy to see what our ViewModel exposes to the View.

levieggertcru: Also see Protocol Oriented Programming -> https://scotteg.github.io/protocol-oriented-programming

levieggertcru: Hey @reldredge71 also a good read https://docs.swift.org/swift-book/LanguageGuide/Protocols.html

godtools/App/Features/ChooseYourOwnAdventure/ChooseYourOwnAdventureViewModel.swift
levieggertcru: Hey @frett I haven't heard of the nullable receiver concept before.  I actually need to update that because I changed the behavior for opening a tool.  If the manifest is null an error is presented instead of opening to a tool.

frett: alright, let me know if there are any other places that you could benefit from having the defaults values from the shared parser

frett: the Kotlin nullable receiver concept may not be something that translates into Swift, in which case we will probably want to figure out an alternate way to expose those defaults

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentButton/MobileContentButtonViewModel.swift
```diff
@@ -29,34 +29,40 @@ class MobileContentButtonViewModel: NSObject, MobileContentButtonViewModelType {
...
             borderColor = buttonColor
+        case .unknown:
+            backgroundColor = buttonColor
+            titleColor = buttonTitleColor ?? renderedPageContext.pageModel.primaryTextColor
+            borderColor = UIColor.clear
+        default:
+            backgroundColor = buttonColor
+            titleColor = buttonTitleColor ?? renderedPageContext.pageModel.primaryTextColor
+            borderColor = UIColor.clear
         }
```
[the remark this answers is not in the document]
levieggertcru: Hey @reldredge71 I like that idea.

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentButton/MobileContentButtonViewModel.swift
```diff
@@ -38,29 +38,33 @@ class MobileContentButtonViewModel: NSObject, MobileContentButtonViewModelType {
         
         buttonWidth = MobileContentViewWidth(dimension: buttonModel.width)
         
-        let buttonColor: UIColor = buttonModel.buttonColor
-        let buttonTitleColor: UIColor? = buttonModel.text?.textColor
-                
+        let defaultBackgroundColor: UIColor = buttonModel.buttonColor
+        let defaultTitleColor: UIColor = buttonModel.text?.textColor ?? renderedPageContext.pageModel.primaryTextColor
```
frett, line 42: hopefully my work on making `text` non-null will make this fallback behavior unnecessary

levieggertcru: Hey @frett I like the sound of that.

frett: https://github.com/CruGlobal/kotlin-mpp-godtools-tool-parser/pull/228

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentButton/MobileContentButtonViewModel.swift
```diff
@@ -113,23 +123,23 @@ class MobileContentButtonViewModel: NSObject, MobileContentButtonViewModelType {
...
     
-    var rendererState: MobileContentMultiplatformState {
-        return rendererPageModel.rendererState
+    var rendererState: State {
+        return renderedPageContext.rendererState
     }
     
     func buttonTapped() {
-        mobileContentAnalytics.trackEvents(events: buttonModel.getAnalyticsEvents(), rendererPageModel: rendererPageModel)
+        mobileContentAnalytics.trackEvents(events: buttonModel.analyticsEvents, renderedPageContext: renderedPageContext)
```
frett, line 143: you should use `buttonModel.getAnalyticsEvents(Trigger.CLICKED)` for fetching the analytics events to trigger here

godtools/App/Services/Renderer/Analytics/MobileContentAnalytics.swift
```diff
@@ -23,27 +24,28 @@ class MobileContentAnalytics {
...
                 
-        let data: [String: String] = event.getAttributes()
+        let data: [String: String] = event.attributes
+        let systems: [AnalyticsEvent.System] = Array(event.systems)
         
-        for system in event.systems {
+        for system in systems {
              
-            if let analyticsSystem = analyticsSystems[system.lowercased()] {
+            if let analyticsSystem = analyticsSystems[system.name.lowercased()] {
```
frett, line 45: maybe an enhancement here is to define the `analyticsSystems` dictionary to use the System enum values as the keys instead of having to convert them to strings.

godtools/App/Services/Renderer/BackgroundImageRenderer/MobileContentBackgroundImageRenderer.swift
```diff
@@ -42,41 +43,50 @@ class MobileContentBackgroundImageRenderer: MobileContentBackgroundImageRenderer
...
+                    
+        case .fillX:
             return scaleRectToHorizontallyFitContainer(container: container, rect: rect)
         
-        case .fillVertically:
+        case .fillY:
             return scaleRectToVerticallyFitContainer(container: container, rect: rect)
+            
+        default:
+            return scaleRectToFitContainer(container: container, rect: rect)
```
frett, line 54: since your default case is the same as the `.fit` case, can you combine them like this:

```swift
case .fit:
default:
    return scaleRectToFitContainer(container: container, rect: rect)
```

that will make it more clear you are defaulting to the fit scale type if the provided scaleType isn't found

godtools/App/Services/Renderer/BackgroundImageRenderer/MobileContentBackgroundImageRenderer.swift
```diff
@@ -42,41 +43,50 @@ class MobileContentBackgroundImageRenderer: MobileContentBackgroundImageRenderer
...
+        
+        case .end:
             positionedRect = positionRectInContainer(container: container, rect: positionedRect, align: .end, languageDirection: languageDirection)
-        }
-        else if alignment.isCenterX {
+       
+        case .center:
+            positionedRect = centerRectHorizontallyInContainer(container: container, rect: positionedRect)
+            
+        default:
```
frett, line 73: similar comment as the default case of the `ScaleType`

godtools/App/Services/Renderer/Views/Tool/Views/Modal/ToolPageModalView.swift
```diff
@@ -98,9 +99,9 @@ class ToolPageModalView: MobileContentView {
         centerContentStackVerticallyIfNeeded()
     }
     
-    override func didReceiveEvents(eventIds: [MultiplatformEventId]) {
+    override func didReceiveEvents(eventIds: [EventId]) {
             
-        let eventsContainFollowUp: Bool = eventIds.contains(MultiplatformEventId.followUp)
+        let eventsContainFollowUp: Bool = eventIds.contains(EventId.Companion().FOLLOWUP)
```
frett, line 104: are you triggering the modal if the FOLLOWUP event is triggered? There is actually listeners defined on the Modal model that contain which events should trigger a specific modal.

frett: hmm, I see you're using the listeners on line 108, but I'm confused why it's filtering based on the presence of the followup event.

godtools/App/Features/ToolDetails/ToolDetailViewModel.swift
```diff
@@ -208,20 +211,29 @@ class ToolDetailViewModel: NSObject, ToolDetailViewModelType {
...
         case .success(let translationManifest):
             
-            let manifest: MobileContentXmlManifest = MobileContentXmlManifest(translationManifestData: translationManifest)
-            
-            hidesLearnToShareButton = manifest.tips.isEmpty
+            let result: Result<Manifest, Error> = mobileContentParser.parse(translationManifestData: translationManifest)
             
+            switch result {
+            case .success(let manifest):
+                tips = manifest.tips
```
frett, line 226: not sure how you are using `tips` elsewhere, but there is a `manifest.findTip(name)` for finding a specific tip model. I'm not sure if this would help with anything in your logic

godtools/App/Services/Renderer/BackgroundImageRenderer/MobileContentBackgroundImageRenderer.swift
levieggertcru: @frett unfortunately it appears I can't do that with the default case.

godtools/App/Services/Renderer/Analytics/MobileContentAnalytics.swift
frett: I'm not sure if this is actually possible, the documentation says the dictionary key needs to be hashable. I'm not sure if a Kotlin enum implements that protocol or not

godtools/App/Services/Renderer/BackgroundImageRenderer/MobileContentBackgroundImageRenderer.swift
frett: k, wasn't sure language limitations in Swift. I actually can't do that in Kotlin, but it works fine in Java. it depends on how the language handles switches

godtools/App/Services/Renderer/BackgroundImageRenderer/MobileContentBackgroundImageRenderer.swift
frett: resolving due to discussion on other comment

godtools/App/Features/ToolDetails/ToolDetailViewModel.swift
frett: nm, this is tool details where you just need to know if tips exist or not.

frett: can't you just do:
```
hidesLearnToShareButton = manifest.tips.isEmpty
```

godtools/App/Services/Renderer/Views/Tool/Views/Modal/ToolPageModalView.swift
frett: Levi is going to investigate this in a followup ticket: https://jira.cru.org/browse/GT-1480

godtools/App/Services/Renderer/BackgroundImageRenderer/MobileContentBackgroundImageRenderer.swift
levieggertcru: I can chain cases by comma separating them, but it won't allow me to do it with the default case.  I wish I didn't have to include the default case, but the compiler is forcing me too.  I'm pretty sure if new cases are added here I won't get a compiler error because the default is there.

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentButton/MobileContentButtonViewModel.swift
levieggertcru: Hey @frett I'm going to create a separate ticket for this one too.

levieggertcru: https://jira.cru.org/browse/GT-1481

godtools/App/Services/Renderer/BackgroundImageRenderer/MobileContentBackgroundImageRenderer.swift
frett: yeah, it would be nice if the compiler recognized it as an enumeration and that you were handling all the cases already

godtools/App/Features/ToolDetails/ToolDetailViewModel.swift
frett: so, I was thinking you didn't need to track intermediary objects at all, your logic could be something as simple as:

```swift
let hidesLearnToShareButton: Bool

let result = parseManifest()
switch result {
    case .success(let manifest):
        hidesLearnToShareButton = manifest.tips.isEmpty
    case .failure(let error):
        hidesLearnToShareButton = true
}
```

godtools/App/Services/Renderer/BackgroundImageRenderer/MobileContentBackgroundImageRenderer.swift
frett: this looks like a feature request for this functionality:
https://youtrack.jetbrains.com/issue/KT-48068

I starred and liked it so that I get notifications if the status ever changes

godtools/App/Features/ToolDetails/ToolDetailViewModel.swift
levieggertcru: Hey @frett I like that approach.  I'll include that.

godtools/App/Services/Renderer/ManifestResourcesCache/ManifestResourcesCache.swift
```diff
@@ -7,25 +7,26 @@
...
+    required init(manifest: Manifest, translationsFileCache: TranslationsFileCache) {
         
         self.manifest = manifest
         self.translationsFileCache = translationsFileCache
     }
     
     private func getSHA256FileLocation(fileName: String) -> SHA256FileLocation? {
         
-        guard let resource = manifest.getResource(fileName: fileName) else {
+        guard let localeName = manifest.resources[fileName]?.localName else {
```
frett, line 25: there is still a `manifest.getResource(fileName)` method that will give you the appropriate resource

frett: either way works.

I would have wanted to eventually make the actual `resources` Dictionary internal to the library, but with the upcoming tool distribution changes, we will need to be able to access all the resources that are included with a tool when downloading it

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentLink/MobileContentLinkViewModel.swift
```diff
@@ -7,25 +7,26 @@
...
-    required init(linkModel: ContentLinkModelType, rendererPageModel: MobileContentRendererPageModel, mobileContentAnalytics: MobileContentAnalytics, fontService: FontService) {
+    required init(linkModel: Link, renderedPageContext: MobileContentRenderedPageContext, mobileContentAnalytics: MobileContentAnalytics, fontService: FontService) {
         
         self.linkModel = linkModel
-        self.rendererPageModel = rendererPageModel
+        self.renderedPageContext = renderedPageContext
         self.mobileContentAnalytics = mobileContentAnalytics
         self.fontService = fontService
-        self.titleColor = linkModel.getTextColor()?.uiColor ?? rendererPageModel.pageColors.primaryColor.uiColor
+        self.titleColor = linkModel.textColor
```
frett, line 29: I think this suffers from the same text styles issues that buttons suffered from. `Link` probably currently implements `Styles` which is what is giving you the `textColor` property, but that doesn't take a `text-color` set on the text node into account.

frett: I'm going to work up another PR to do the same thing I did to Buttons to Links

godtools/App/Services/Renderer/ManifestResourcesCache/ManifestResourcesCache.swift
frett: scratch this, that method is internal only. I guess the normal expected usage is to access the resource object directly from the Image/Animation/etc that is referencing it.

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentButton/MobileContentButtonViewModel.swift
```diff
@@ -29,34 +29,44 @@ class MobileContentButtonViewModel: NSObject, MobileContentButtonViewModelType {
...
+        default:
+            backgroundColor = defaultBackgroundColor
+            titleColor = defaultTitleColor
+            borderColor = defaultBorderColor
         }
-        
-        if let name = buttonModel.iconName, let image = rendererPageModel.resourcesCache.getImageFromManifestResources(fileName: name)  {
+                
+        if let name = buttonModel.icon?.name,
+            let image = renderedPageContext.resourcesCache.getImageFromManifestResources(fileName: name)  {
```
frett, line 69: an optimization you could make is make `getImageFromManifestResources()` take the `Resource` object directly, that way you don't have to re-lookup the resource object inside that logic.

godtools/App/Services/Renderer/Views/MobileContent/Views/Page/MobileContentPageViewModel.swift
```diff
@@ -55,29 +52,40 @@ class MobileContentPageViewModel: MobileContentPageViewModelType {
...
-        else if manifestAttributes.backgroundImageExists {
-            backgroundImageModel = manifestAttributes
+        else if let manifestBackgroundImageFileName = manifest.backgroundImage?.name {
+            
+            backgroundImageModel = BackgroundImageModel(
+                backgroundImageFileName: manifestBackgroundImageFileName,
+                backgroundImageAlignment: manifest.backgroundImageGravity,
+                backgroundImageScale: manifest.backgroundImageScaleType
+            )
         }
```
frett, line 74: I actually think you aren't fully handling background images correctly here. It's possible to define both a manifest and a page background and have them layered overtop of each other.

There isn't any actual usages in production that I'm aware of, but the manifest & page background logic is meant to be layered and not one or the other.

frett: this is a bigger change than the scope of this PR

godtools/App/Services/Renderer/Views/Tool/Views/Header/ToolPageHeaderViewModel.swift
```diff
@@ -7,24 +7,75 @@
...
     }
     
     var languageDirectionSemanticContentAttribute: UISemanticContentAttribute {
-        return rendererPageModel.language.languageDirection.semanticContentAttribute
+        return renderedPageContext.language.languageDirection.semanticContentAttribute
     }
     
     var backgroundColor: UIColor {
-        return rendererPageModel.pageColors.primaryColor.uiColor
+        return renderedPageContext.pageModel.primaryColor
```
frett, line 36: you should be able to use `headerModel.backgroundColor` for this

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentLink/MobileContentLinkViewModel.swift
frett: https://github.com/CruGlobal/kotlin-mpp-godtools-tool-parser/pull/230

frett (approved):
just the last couple comments, feel free to address them or merge as is

godtools/App/Services/Renderer/Views/MobileContent/Views/Page/MobileContentPageViewModel.swift
levieggertcru: Hey @frett created this ticket here -> https://jira.cru.org/browse/GT-1485

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentButton/MobileContentButtonViewModel.swift
levieggertcru: @frett I like this idea.

levieggertcru: Hey @frett just committed this change.

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentLink/MobileContentLinkViewModel.swift
levieggertcru: Hey @frett will I need to update this to use the ```linkModel.text.textColor```?  Or does your change make ```linkMode.textColor``` equivalent to that?

godtools/App/Services/Renderer/ManifestResourcesCache/ManifestResourcesCache.swift
```diff
@@ -39,9 +40,13 @@ class ManifestResourcesCache: ManifestResourcesCacheType {
...
     
-    func getImageFromManifestResources(fileName: String) -> UIImage? {
+    func getImageFromManifestResources(resource: Resource) -> UIImage? {
         
-        guard let location = getSHA256FileLocation(fileName: fileName) else {
+        guard let resourceName = resource.name, !resourceName.isEmpty else {
+            return nil
+        }
+        
+        guard let location = getSHA256FileLocation(fileName: resourceName) else {
```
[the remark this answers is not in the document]
frett: it looks like you will have to move the current fileName -> resource lookup logic into `getFile()` since that also calls `getSHA256FileLocation()`. That code path might be able to be improved to pass in a `resource` object as well?

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentLink/MobileContentLinkViewModel.swift
frett: you will need to change to `linkModel.text.textColor`.

`linkModel.textColor` will actually become invalid when you upgrade past 0.5.0 since that property is no longer present on the `Link` model itself

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentLink/MobileContentLinkViewModel.swift
```diff
@@ -37,18 +38,19 @@ class MobileContentLinkViewModel: MobileContentLinkViewModelType {
...
-    var rendererState: MobileContentMultiplatformState {
-        return rendererPageModel.rendererState
+    var rendererState: State {
+        return renderedPageContext.rendererState
     }
     
     func linkTapped() {
-        mobileContentAnalytics.trackEvents(events: linkModel.getAnalyticsEvents(), rendererPageModel: rendererPageModel)
+                        
+        mobileContentAnalytics.trackEvents(events: linkModel.analyticsEvents, renderedPageContext: renderedPageContext)
```
frett, line 54: you should use `linkModel.getAnalyticsEvents(.clicked)`

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentText/MobileContentTextViewModel.swift
```diff
@@ -58,7 +49,7 @@ class MobileContentTextViewModel: MobileContentTextViewModelType {
     }
     
     var hidesStartImage: Bool {
-        guard let resource = textModel.startImage else {
+        guard let resource = startImageResourceName else {
```
frett, line 52:
```suggestion
        guard let resource = textModel.startImage else {
```

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentText/MobileContentTextViewModel.swift
```diff
@@ -58,7 +49,7 @@ class MobileContentTextViewModel: MobileContentTextViewModelType {
...
             return true
         }
         return resource.isEmpty
```
frett, line 55: this could be simplified to just be based off of `textModel.startImage == null`

frett: similar can be done for the end image (if that logic exists)

godtools/App/Services/Renderer/ManifestResourcesCache/ManifestResourcesCache.swift
```diff
@@ -7,41 +7,46 @@
...
-    private func getSHA256FileLocation(fileName: String) -> SHA256FileLocation? {
+    private func getSHA256FileLocation(resource: Resource) -> SHA256FileLocation? {
         
-        guard let resource = manifest.getResource(fileName: fileName) else {
+        guard let resourceName = resource.name else {
             return nil
         }
         
-        let location: SHA256FileLocation = SHA256FileLocation(sha256WithPathExtension: resource.src)
+        guard let localeName = manifest.resources[resourceName]?.localName else {
```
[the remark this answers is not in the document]
levieggertcru: Hey @frett updated that.  Yeah no need for me to fetch from manifest.resources like I was doing.

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentText/MobileContentTextViewModel.swift
levieggertcru: @frett  I like that better.

frett (approved):

PR #1037, opened 2022-02-28
GT-1367 Fix: Use Text-Scaling on Button Text, not Button

reldredge71:
This applies the proper value of the `textScaling` property to button text.

### DIFF
--- a/godtools/App/Services/Renderer/Parser/Multiplatform/Models/MultiplatformContentButton.swift
+++ b/godtools/App/Services/Renderer/Parser/Multiplatform/Models/MultiplatformContentButton.swift
@@ -80,7 +80,7 @@ class MultiplatformContentButton: ContentButtonModelType {
     }
     
     var textScale: MobileContentTextScale {
-        return MobileContentTextScale(doubleValue: button.textScale)
+        return MobileContentTextScale(doubleValue: button.text?.textScale ?? button.textScale)
     }
     
     var buttonWidth: GodToolsToolParser.Dimension {


godtools/App/Services/Renderer/Parser/Multiplatform/Models/MultiplatformContentButton.swift
```diff
@@ -80,7 +80,7 @@ class MultiplatformContentButton: ContentButtonModelType {
     }
     
     var textScale: MobileContentTextScale {
-        return MobileContentTextScale(doubleValue: button.textScale)
+        return MobileContentTextScale(doubleValue: button.text?.textScale ?? 1.0)
```
frett, line 83:
```suggestion
        return MobileContentTextScale(doubleValue: button.text?.textScale ?? button.textScale)
```
it's probably safer to just default back to the text scale that the button inherits instead of to a constant of 1.0

levieggertcru: Hey @reldredge71 can you hold on this PR until  GT-1422 is merged?  MobileContentTextScale is going to be non existent once that is merged in.  And GT-1422 is ready for review now :)

reldredge71: @levieggertcru Sounds good.  I'll look through GT-1422 now.

PR #1044, opened 2022-03-04
GT-1398 -- Fix for disappearing Screen Share Icon

reldredge71:
Screen Share Icon should remain visible at the top of the screen while screen sharing is happening.


levieggertcru:
Hey @reldredge71 I was just testing and looks like we also need to set the bar button item to null.  ```remoteShareActiveNavItem = nil```. Maybe do that when all bar button items are removed.

levieggertcru (approved):

PR #1046, opened 2022-03-07
GT-1477: Add Analytics Tracking for Pressing Quick Start Links

reldredge71:
Created a couple of new analytics events to be triggered when pressing the Quick Start links, or pressing the "Get Started" button.

### DIFF
--- a/godtools/App/Features/Onboarding/Models/OnboardingQuickStartItem.swift
+++ b/godtools/App/Features/Onboarding/Models/OnboardingQuickStartItem.swift
@@ -10,8 +10,8 @@ import Foundation
 
 struct OnboardingQuickStartItem {
     
-    var title: String
-    var linkButtonTitle: String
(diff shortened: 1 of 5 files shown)


godtools/App/Features/Onboarding/Models/OnboardingQuickStartItem.swift
```diff
@@ -13,4 +13,5 @@ struct OnboardingQuickStartItem {
     var title: String
     var linkButtonTitle: String
     var linkButtonFlowStep: FlowStep
+    var linkButtonAnalyticsAction: String
```
levieggertcru, line 16: Hey @reldredge71 I didn't catch this before, but can you change these to ```let``` so they're immutable.

levieggertcru (changes requested):

levieggertcru (approved):

PR #1055, opened 2022-03-14
GT create folders for different flows

levieggertcru:
This PR cleans up the Flows directory.  Before, the Flows directory contained all of the Flow class swift files and FlowCompletedState enumeration swift files and it was getting difficult to visually see the different Flow classes.  This change creates a new directory each Flow class to better clean up the Flows directory.  Some Flow classes have an associated FlowCompletedState enumeration swift file.  Each Flow directory will contain both the Flow swift file and FlowCompletedState enumeration swift file.


godtools/App/Flows/ToolsFlow.swift
```diff
@@ -1,384 +0,0 @@
...
-//  ToolsFlow.swift
-//  godtools
-//
-//  Created by Levi Eggert on 5/26/20.
-//  Copyright © 2020 Cru. All rights reserved.
-//
-
-import UIKit
-
-class ToolsFlow: ToolNavigationFlow, Flow {
```
frett, line 11: did you mean to delete this and not move it?

levieggertcru: Hey Daniel I should have mentioned that in the PR description, but yes I did mean to remove that file.  That logic there was merged some time ago into the App Flow and is no longer being used.

PR #1077, opened 2022-03-30
GT-1406 Implement flow item state visibility

levieggertcru:
This PR includes work that will update flowItems visibility (gone, hidden, visible) within the content flow.  Whenever a flowItem's visibility changes and the previous visibility or new visibility is equal to gone, it will trigger a rebuilding of the entire content flow and flow items.

### DIFF
--- a/godtools/App/Services/Renderer/Views/MobileContent/Views/ContentFlowItem/MobileContentFlowItemViewModelType.swift
+++ b/godtools/App/Services/Renderer/Views/MobileContent/Views/ContentFlowItem/MobileContentFlowItemViewModelType.swift
@@ -11,5 +11,6 @@ import GodToolsToolParser
 
 protocol MobileContentFlowItemViewModelType {
     
+    var visibilityState: ObservableValue<MobileContentViewVisibilityState> { get }
     var width: MobileContentViewWidth { get }
 }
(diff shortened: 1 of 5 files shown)


godtools/App/Services/Renderer/Views/MobileContent/Views/ContentFlowItem/MobileContentFlowItemViewModelType.swift
```diff
@@ -11,5 +11,6 @@ import GodToolsToolParser
 
 protocol MobileContentFlowItemViewModelType {
     
+    var visibilityState: ObservableValue<MobileContentViewVisibilityState> { get }
```
frett, line 14: hmm, would it be worthwhile to update parts of the shared parser to produce `ObservableValue` wrappers around the various coroutines flows to make it easier to digest than a watcher on your end?

frett: I'm assuming `ObservableValue` is a core API and not from a 3rd party library

levieggertcru: Hey @frett let's hold off on that.  ObservableValue is a simple Class I created for observing a stored value and when it changes.  With our moving to iOS 13 the plan is to eventually phase out ObservableValue in place of Combine's CurrentValueSubject https://developer.apple.com/documentation/combine/currentvaluesubject

frett (approved):

PR #1075, opened 2022-03-28
GT-1493 Upgrade Min Deployment Target to iOS 13

rachaelblue:
https://jira.cru.org/browse/GT-1493

Bumping the deployment target here in preparation for SwiftUI work.

~~ Shouldn't be merged into develop yet since this isn't going in the next release.  It may be worth making a new feature branch for all iOS 13 work until we're ready for it-- thoughts?~~

**Update:** Created the iOS-13 feature branch and merging into that here.

PR #1085, opened 2022-04-11
Merge latest from develop

levieggertcru:


rachaelblue:
So this is going into an actual PR branch, not the iOS-13 feature branch.  Can we merge this PR (https://github.com/CruGlobal/godtools-swift/pull/1075), and then keep iOS-13 up to date with dev?

levieggertcru:
Hey @rachaelblue yeah lets merge this in.  I meant to be pushing dev to the feature branch not here.  Thank you for catching that.

PR #1089, opened 2022-04-11
GT-1431 Add support for flat multi select option

levieggertcru:
Added support for flat MultiSelectOption style which will not render a shadow on the view.


frett (approved):
that's pretty simple

PR #1094, opened 2022-04-14
GT-1509 Add SwiftUIPreviewDiContainer

levieggertcru:


rachaelblue (approved):
Wow, this was a lot simpler than I thought 😅 thank you!!  Not ideal to have network calls in previews, but it's gonna be a huge effort to work around that.

PR #1096, opened 2022-04-18
GT-1512 vertically center text in tract header

levieggertcru:


levieggertcru:
In this PR:
- Vertically center text in tract header
- Added content insets to mobile content cards

frett (approved):

PR #1100, opened 2022-04-20
GT-1474 refactoring to deep linking configuration

levieggertcru:


levieggertcru:
In this PR some refactoring was done to how DeepLinkingService.swift is instantiated.  The intent of this PR is to give a broad overview of what urls the DeepLinkingService supports and which parsers are parsing those urls.

Originally the DeepLinkingService was allocated with concrete parser classes that it would run through and attempt to get a valid ParsedDeepLinkType from.  

Now DeepLinkingService is instantiated with the GodToolsDeepLinkingManifest which pre-defines the parsers and the rules which allow a parser to be instantiated.  Currently there are only rules setup for DeepLinkingParserManifestUrl types.  The url must match the scheme, host, and root path component in the manifest in order for the parser to be instantiated and parsed against.

In addition, this PR supports parsing the new AppsFlyer DeepLinkValue's and old AppsFlyer DeepLinkValue parsing is moved to the LegacyAppsFlyerDeepLinkValueParser.

rachaelblue:
I like the manifest, nice way to organize the schemes!  Is there a way you recommend to generate deep links (there's only deep link support, not universal link, correct?)

levieggertcru:
Hey @rachaelblue we have a mix of both.  The URL scheme we're using is ```godtools```.  We also have some universal links setup on ```knowgod.com``` and ```godtoolsapp.com```.  I've been using a little test site here for testing deep links into the app https://deeplinking-7dce4.web.app/

frett:
long term I would like to see us migrate away from any `godtoolsapp.com` links because those don't actually exist on the web.

`knowgod.com` links do work on the web, so we should continue to prefer those whenever possible

PR #1104, opened 2022-04-22
GT-1283 fix content image fixed point sizing and sizing based on percentage of container

levieggertcru:


levieggertcru:
This PR should fix the content image width logic when sizing images to a fixed width or percentage width of their container when rendered in the content stackview. 
I had to manually do these calculations when the content stack bounds changed to ensure the image would size properly to its containing view and not go any larger than the width defined from the model.

One thing to note is that when those image constraints are updated due to a bounds change, there will be warnings in Xcode about those constraints breaking.  This is specifically true when a row of content images are rendered because the row of content is expecting all of those images to be the same height.

gyasistory (approved):

PR #1105, opened 2022-04-27
GT-1505 upgrade google tag manager

levieggertcru:
The initial work for this PR was upgrading google tag manager which resulted in a few other needed changes.
- Upgraded GoogleTagManager version (Now pulling from Swift Package Manager)
- GoogleTagManager is dependent on the FirebaseSDK (Now pulling Firebase dependencies from Swift Package Manager)
- This import is no longer needed in the bridging header ```#import <GoogleAnalytics/GAI.h>```. I believe removing this resulted in a lot of compiler errors where UIKit wasn't getting imported.  Updated those files to import UIKit.
- Also updated the crashlytics build phase for swift package manager

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,9 +21,9 @@ jobs:
             access_token: ${{ github.token }}
   
   build:
-    runs-on: macos-11
+    runs-on: macos-12
     env:
-      DEVELOPER_DIR: /Applications/Xcode_13.2.app/Contents/Developer
+      DEVELOPER_DIR: /Applications/Xcode_13.3.1.app/Contents/Developer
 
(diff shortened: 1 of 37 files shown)


rachaelblue:
Hey @levieggertcru, I can't seem to make the new packages ever resolve.  Any ideas?  Could be an M1 issue, I think I had trouble the first time with the other packages too 🤔
Screen Shot 2022-04-28 at 10 20 57 PM

levieggertcru:
Hey @rachaelblue try resetting the package caches and see if that resolves it.  If that doesn't resolve the packages then what Xcode version are you running?  You may need to upgrade to the latest Xcode 13.3.1.

PR #1111, opened 2022-05-02
GT-1541 all tools remove scroll to top

rachaelblue:
https://jira.cru.org/browse/GT-1541


godtools/App/Features/Tools/ToolsMenuView.swift
```diff
@@ -123,8 +123,8 @@ class ToolsMenuView: UIViewController {
         lessonsView?.scrollToTopOfLessonsList(animated: false)
         favoritedToolsView?.scrollToTopOfToolsList(animated: false)
         
-        // TODO: - GT-1541 Remove the ability to programmatically scroll to the top and just reset the whole view instead
-        allToolsView?.rootView.viewModel.scrollToTop(animated: false)
+        // Programmatically scrolling to the top in SwiftUI is only available on iOS 14+, so we'll just reset the whole view instead
+        allToolsView?.rootView = AllToolsContentView(viewModel: viewModel.allToolsWillAppear())
```
levieggertcru, line 127: Hey @rachaelblue there are some additional steps that would be needed to reset the allToolsView since the ToolsMenuView is built programmatically by attaching each tools list view (lessons, favorites, allTools) to a scrollView.  This would involve some extra steps such as removing the old view from the scrollView and attaching the new view, which could be done by getting the frame position of the original view and copying it to the newly attached view.

However, I think we should just leave this PR as is, and before merging the iOS 13 feature branch I'd like to complete this task here (https://jira.cru.org/browse/GT-1545) which is removing this reset method all together and just re-instantiate the ToolsMenuView from the flow rather than calling reset on it.  That will makes things a lot simpler.

levieggertcru (approved):

godtools/App/Features/Tools/ToolsMenuView.swift
rachaelblue: Gotcha.  What's the goal then with removing the VC from the scroll view and reattaching a new one?  If all we need is the All Tools to be scrolled to the top, this should be sufficient as is?

levieggertcru: This current change ```allToolsView?.rootView = AllToolsContentView(viewModel: viewModel.allToolsWillAppear())``` is part of the steps needed to reset AllTools, but the old view is still in memory and attached to the ScrollView.  So there would be extra steps in removing the old rootView from memory and attaching the newly created rootView to the ScrollView.

PR #1113, opened 2022-05-03
GT-1058 Add build script phase to download languages and resources json files

levieggertcru:
Added a build script phase that will download and include the languages and resources json files to the project to keep it up to date when releasing builds to the app store.


godtools.xcodeproj/project.pbxproj
```diff
@@ -6290,6 +6291,24 @@
...
+			inputPaths = (
+			);
+			name = "Download And Add Initial Resources and Languages JSON Files";
+			outputFileListPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "curl -o godtools/Assets/InitialResources/languages.json https://mobile-content-api.cru.org/languages.json\ncurl -o godtools/Assets/InitialResources/resources.json https://mobile-content-api.cru.org/resources?include=latest-translations,attachments.json\n";
```
frett, line 6310: does curl require an extension on the URL? I would say that you shouldn't add the `.json` to the URLs you are downloading.

so use these 2 urls instead:
```
https://mobile-content-api.cru.org/languages
https://mobile-content-api.cru.org/resources?include=latest-translations,attachments
```

frett: also, does this phase run for builds targeting the stage environment as well? should there be a way to use the languages/resources from the stage environment instead in that case?

levieggertcru: Hey Daniel, the extension wasn't required.  Just updated.

The phase will execute against both staging and production schemes.  Let me see if I can differentiate between them.  That would be nice to add.

levieggertcru: So the build phase executes against any running scheme (GodTools-Staging, GodTools-Production, godtools, GodTools-Release, etc.) I wonder if I can check which scheme it is from build phase...

levieggertcru: Hey @frett there isn't a way I can differentiate which scheme the build phase is in.  I think it will be fine as long as we run staging with an active internet connection.

frett (dismissed):

frett (commented):
Now that we download languages.json and resources.json, should we no longer track those in git?

levieggertcru:
Hey @frett we probably don't need to, let me see what happens if I don't.  I'm curious how that will be treated with the Xcode project when loading those from the bundle.

levieggertcru:
Hey @frett looks like I will have to leave those tracked in git.  Just because they need to be part of the Xcode project when loading those resources from the main app bundle.

frett (approved):

PR #1118, opened 2022-05-05
Gt 1514 fix Vietnamese prayer pathway registration bug

levieggertcru:
This PR fixes an issue where the ```followup:send``` event was getting sent to all forms in a page when it should only be sent to the form in the view hierarchy that the event was generated from.  
This was fixed by changing how events are sent to views in a page.  Originally, whenever an event is triggered, it will start at the rootView (page) of the view hierarchy and recurse all children passing the eventIds to each child for processing.  This means that the ```followup:send``` event would get passed to each form in the page hierarchy which is what we don't want.  

To fix this, a couple things were done:

1. Changed the didReceiveEvents method to process a single eventId at a time and added a ```ProcessedEventResult?``` return type.  This return type contains a flag ```shouldRemoveEventId``` which gives more granular control on which eventIds should be removed if successfully processed.

2. Changed how events are passed to views when an event is triggered.  Now when an event is triggered from a view, we walk up the view hierarchy starting from that view and process the event on the children of each parent as we walk up.  This gives direct ascendants a chance to process events before reaching the root view and sending events to all children in the view hierarchy.

### DIFF
--- a/godtools/App/Services/Renderer/Views/MobileContent/Views/MobileContentView.swift
+++ b/godtools/App/Services/Renderer/Views/MobileContent/Views/MobileContentView.swift
@@ -151,7 +151,7 @@ class MobileContentView: UIView {
                 
         walkUpHierarchyAndSendEventsToChildren(view: self, eventIds: &eventIds)
                 
-        recurseChildrenAndSendEvents(view: getRootView(), eventIds: &eventIds)
+        recurseChildrenAndSendEventsFromView(view: getRootView(), eventIds: &eventIds)
     }
     
     private func walkUpHierarchyAndSendEventsToChildren(view: MobileContentView, eventIds: inout [EventId]) {
@@ -172,14 +172,14 @@ class MobileContentView: UIView {
         }
     }
     
-    private func recurseChildrenAndSendEvents(view: MobileContentView, eventIds: inout [EventId]) {
+    private func recurseChildrenAndSendEventsFromView(view: MobileContentView, eventIds: inout [EventId]) {
         
         guard !eventIds.isEmpty else {
             return
         }
         
         for childView in view.children {
-            recurseChildrenAndSendEvents(view: childView, eventIds: &eventIds)
+            recurseChildrenAndSendEventsFromView(view: childView, eventIds: &eventIds)
         }
         
         sendEventsToView(view: view, eventIds: &eventIds)
--- a/godtools/App/Services/Renderer/Views/MobileContent/Views/MobileContentView.swift
+++ b/godtools/App/Services/Renderer/Views/MobileContent/Views/MobileContentView.swift
@@ -146,26 +146,60 @@ class MobileContentView: UIView {
     // MARK: - Events
     
     func sendEventsToAllViews(eventIds: [EventId], rendererState: State) {
-        
-        let eventIds: [EventId] = eventIds.flatMap({$0.resolve(state: rendererState)})
-        
-        recurseChildrenAndSendEvents(view: getRootView(), eventIds: eventIds)
+            
+        var eventIds: [EventId] = eventIds.flatMap({$0.resolve(state: rendererState)})
+                
+        walkUpHierarchyAndSendEventsToChildren(view: self, eventIds: &eventIds)
+                
+        recurseChildrenAndSendEvents(view: getRootView(), eventIds: &eventIds)
     }
     
-    private func recurseChildrenAndSendEvents(view: MobileContentView, eventIds: [EventId]) {
+    private func walkUpHierarchyAndSendEventsToChildren(view: MobileContentView, eventIds: inout [EventId]) {
         
-        for childView in view.children {
-            recurseChildrenAndSendEvents(view: childView, eventIds: eventIds)
+        var nextParentView: MobileContentView? = view.parent
+        
+        while nextParentView != nil && !eventIds.isEmpty {
+            
+            if let children = nextParentView?.children {
+                
+                for childView in children {
+                    
+                    sendEventsToView(view: childView, eventIds: &eventIds)
+                }
+            }
+            
+            nextParentView = nextParentView?.parent
+        }
+    }
+    
+    private func recurseChildrenAndSendEvents(view: MobileContentView, eventIds: inout [EventId]) {
+        
+        guard !eventIds.isEmpty else {
+            return
         }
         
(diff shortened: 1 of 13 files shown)


levieggertcru:
Hey @frett do you see any issues with this implementation?  I will remove an eventId if it is successfully processed.  The only drawback is if there could be multiple listeners triggered by a single eventId.

frett:
@levieggertcru We might depend on multiple things listening to the same eventId, nothing in the spec/design is built to try to restrict that, and there aren't well written processing rules in that case to define what gets the event if multiple things process it

godtools/App/Services/Renderer/Views/Tool/Views/Modal/ToolPageModalView.swift
```diff
@@ -87,18 +87,17 @@ class ToolPageModalView: MobileContentView, NibBased {
...
         
-        for eventId in eventIds {
-            
-            if viewModel.listeners.contains(eventId) && !eventsContainFollowUp {
-                delegate?.toolPageModalListenerActivated(modalView: self)
-            }
-            else if viewModel.dismissListeners.contains(eventId) {
-                delegate?.toolPageModalDismissListenerActivated(modalView: self)
-            }
+        let isFollowUpEvent: Bool = eventId == EventId.Companion().FOLLOWUP
```
frett, line 92: :( I'm sad that this code requires `Companion()`. I'm not sure of a way around that in kotlin native currently

levieggertcru:
Hey @frett alright fixed this by adding a more granular control mechanism for removing processed event ids.  Any eventId being processed by a view can return an option type of ```ProcessedEventResult``` and set a flag ```shouldRemoveEventId```.  This way when a FormView processes an event it can tell the event system to remove that eventId.

godtools/App/Services/Renderer/Views/Tool/Views/Modal/ToolPageModalView.swift
```diff
@@ -87,18 +87,17 @@ class ToolPageModalView: MobileContentView, NibBased {
...
+                
+        if viewModel.listeners.contains(eventId) && !isFollowUpEvent {
+            delegate?.toolPageModalListenerActivated(modalView: self)
+        }
+        else if viewModel.dismissListeners.contains(eventId) {
```
frett, line 97: `dismissListeners` should also never contain a followup eventid

godtools/App/Services/Renderer/Views/Tool/Views/Modal/ToolPageModalView.swift
levieggertcru: Yeah I know, it's a bit clunky

godtools/App/Services/Renderer/Views/Tool/Views/Form/ToolPageFormView.swift
```diff
@@ -59,12 +59,15 @@ class ToolPageFormView: MobileContentFormView {
...
-                      
-        if eventIds.contains(EventId.Companion().FOLLOWUP) {
-            pageFormViewModel.sendFollowUp(inputModels: super.getInputModels(), eventIds: eventIds)
+    override func didReceiveEvent(eventId: EventId, eventIdsGroup: [EventId]) -> ProcessedEventResult? {
+        
+        let isFollowUpEvent: Bool = eventId == EventId.Companion().FOLLOWUP
+                
+        if isFollowUpEvent {
+            pageFormViewModel.sendFollowUp(inputModels: super.getInputModels(), eventIds: eventIdsGroup)
+            return ProcessedEventResult(shouldRemoveEventId: true, successfullyProcessed: true)
```
frett, line 68: 👍

godtools/App/Services/Renderer/Views/MobileContent/Views/MobileContentView.swift
```diff
@@ -144,23 +144,74 @@ class MobileContentView: UIView {
...
+        var nextParentView: MobileContentView? = view.parent
         
-        recurseChildrenAndSendEvents(view: getRootView(), eventIds: eventIds)
+        while nextParentView != nil && !eventIds.isEmpty {
+            
+            if let children = nextParentView?.children {
+                
+                for childView in children {
+                    
+                    sendEventsToView(view: childView, eventIds: &eventIds)
```
frett, line 165: hmm, events may be triggered twice because you trigger events on the walk up, but also from the root view

frett: imagine this xml:
```xml
<page>
  <content:paragraph>
    <content:animation toggleListener="eventA" />
    <content:button event="eventA" />
  </content:paragraph>
</page>
```

I believe walking up you will trigger the animation `toggleListener` because you process children of all ancestor nodes. Plus you will trigger `toggleListener` again when you start recursively iterating from the root node.

frett: you probably want a method signature like this for the walk up:

```
func handleSpecialEvents(eventId: EventId) -> Boolean {
     guard parent != null else {
        return false
     }
     
     return parent.handleSpecialEvents(eventId)
}
```

with the form view overriding it like this:

```
func handleSpecialEvents(eventId: EventId) -> Boolean {
   guard eventId == EventId.Companion().FOLLOW_UP else {
     return super(eventId)
   }
   
   // submit followup event for this form
   
   return true
}
```

then when triggering events you could do something like:
```
for eventId in eventIds {
  if (!view.handleSpecialEvents(eventId)) {
    recurseChildrenAndSendEvents(eventId)
  }
}
```

levieggertcru: Hey @frett I like that approach.

levieggertcru: Hey @frett I think this should work now.  I created a separate method signature ```didReceiveAncestryEventFromWalkingUpViewHierarchy``` that views can listen for instead of the plain ```didReceiveEvent```.  The ```ToolPageFormView``` will now look for the new ancestry event signature when processing those forms.  I decided to leave the remove eventId logic for added measure of removing that follow up eventId when processed.

frett (approved):
I think everything looks good now

PR #1132, opened 2022-05-12
GT-1528 consolidate lists for simplification

levieggertcru:


godtools/App/Features/Tools/SwiftUI/AllTools/ItemsList.swift
```diff
@@ -0,0 +1,115 @@
...
+//  godtools
+//
+//  Created by Levi Eggert on 5/12/22.
+//  Copyright © 2022 Cru. All rights reserved.
+//
+
+import SwiftUI
+import UIKit
+
+struct ItemsList<ListItem: Identifiable, ListItemView: View>: View {
```
rachaelblue, line 12: I'm wondering how this is going to work in adding the Spotlight and Categories sections.  I was going to put them all in a List/LazyVStack together, but this only allows a single type of view?

godtools/App/Features/Tools/SwiftUI/AllTools/AllToolsContentView.swift
```diff
@@ -53,25 +29,29 @@ struct AllToolsContentView: View {
...
                         
-                        // Pull to refresh is supported only in iOS 15+
-                        AllToolsList(viewModel: viewModel, width: width)
-                            .refreshable {
-                                viewModel.refreshTools()
-                            }
+                        let cardViewModel = viewModel.cardViewModel(for: item)
                         
-                    } else if #available(iOS 14.0, *) {
+                        return ToolCardView(viewModel: cardViewModel, cardWidth: cardWidth)
```
rachaelblue, line 48: Would it be possible to add the tap gesture here, before returning the view, rather than doing it inside `ItemsList`?

rachaelblue (commented):
This works great testing on iOS 13-15, but the code makes me a little nervous 😬  It seems like this works for a specific use case, but I'm concerned it may be difficult to customize as we add on to the page.  I realize this allows us to use the same view across all iOS versions, but I'd probably vote to keep List for iOS 13 only, and then just use LazyVStack for 14+.  Then as soon as we drop iOS 13, we're back down to a single view with basic SwiftUI components, rather than a custom solution we'd have to continue to upkeep (until we decide to rewrite the view without it).  Thoughts?

godtools/App/Features/Tools/SwiftUI/AllTools/AllToolsContentView.swift
levieggertcru: Hey @rachaelblue I see what you're saying with it being difficult to build on this component.  There are ways that can be achieved.   The main thing I'm looking for is that code isn't being duplicated.  I'd love to have just the AllToolsContentView referencing the viewModel and any logic working with the viewModel is done in the AllToolsContentView and is done once, not repeated in multiple places.  Same with adding categories and spotlight, that code should be done once and not duplicated.  However that can be achieved is what is preferable. 
I'm definitely in favor of dropping List as well and sticking with the LazyVStack going forward (I haven't looked into pull to refresh on LazyVStack, but that option would be nice to have).  It would be nice if we could just go forward with iOS 14.

PR #1137, opened 2022-05-17
GT-1498 All Tools Spotlight

rachaelblue:
https://jira.cru.org/browse/GT-1498

### Things I Did
- refactored the ToolCardView-- broke it out into several subviews to make it easier to read
- adjusted ToolCardView to resize when used in the Spotlight section
- added a Realm attribute for the spotlight on Resource
- added strings to the base localization files (and sent the strings/ids to Ann to add translations)
- refactored AllToolsList so that we don't have separate versions for iOS 13/14/15.  Now, the only version check happens in AllToolsContentView, and all versions use the same AllToolsList (@levieggertcru, I realized this would work better than what we previously discussed 😄)


levieggertcru (changes requested):
Hey @rachaelblue this is looking great.  Here's a recap of those changes we discussed:
- Ensure AppDiContainer dependencies are injected into Mock viewModels.
Also, created this task here for creating the Mock swift local package for all of our Mock files https://jira.cru.org/browse/GT-1577.

godtools/App/Features/Tools/SwiftUI/ToolCardView/ViewModels/MockToolCardViewModel.swift
```diff
@@ -21,8 +20,8 @@ class MockToolCardViewModel: BaseToolCardViewModel {
             parallelLanguageName = "Arabic (Bahrain) ✓"
         }
         
-        if showBannerImage, let deviceImage = appDiContainer.deviceAttachmentBanners.getDeviceBanner(resourceId: "2") {
-            bannerImage = Image(uiImage: deviceImage)
+        if showBannerImage {
+            bannerImage = Image.from(uiImage: DeviceAttachmentBanners().getDeviceBanner(resourceId: "2"))
```
levieggertcru, line 24: Hey @rachaelblue can you actually inject the DeviceAttachmentBanners into the ViewModel and fetch it from the SwiftUIPreviews AppDiContainer.  I should have been more clear on that.  Eventually the MockDiContainer will provide an instance that can be injected.

godtools/App/Features/Tools/SwiftUI/ToolCardView/SubViews/ToolCardBannerImageView.swift
```diff
@@ -0,0 +1,61 @@
...
+            .cornerRadius(cornerRadius, corners: [.topLeft, .topRight])
+    }
+}
+
+// MARK: - Preview
+
+struct ToolCardBannerImageView_Previews: PreviewProvider {
+    static var previews: some View {
+        let isSpotlight = true
+        let bannerImage = DeviceAttachmentBanners().getDeviceBanner(resourceId: "2")
```
levieggertcru, line 50: Same here can you fetch the DeviceAttachmentBanners from the SwiftUIPreviews AppDiContainer.  I wasn't clear on that.

levieggertcru (changes requested):

godtools/App/Features/Tools/SwiftUI/ToolCardView/ViewModels/MockToolCardViewModel.swift
rachaelblue: That makes sense.  Fixed!

levieggertcru (approved):

PR #1160, opened 2022-06-07
GT-1265 all tools categories

rachaelblue:
https://jira.cru.org/browse/GT-1265

Added the category filter buttons.  The buttons are determined programmatically based on what tools are available, and then ordered based on which tools are available in the user's primary language.


godtools/App/Features/Tools/SwiftUI/ToolCategoriesView/ToolCategoryButtonView/ViewModels/BaseToolCategoryButtonViewModel.swift
```diff
@@ -0,0 +1,52 @@
...
+        self.state = buttonState
+        
+        setPublishedValues()
+    }
+}
+
+// MARK: - Public
+
+extension BaseToolCategoryButtonViewModel {
+    func setPublishedValues() {
```
levieggertcru, line 36: Hey @rachaelblue it seems this is supposed to be invoked whenever ```state: ToolCategoryButtonState``` is mutated.  However, I would be able to mutate that property without this method ever getting called.  A safer way would be to force a state argument here, maybe instead of ```setPublishedValues()``` method it could be ```func setButtonState(state: ToolCategoryButtonState)``` that way state is forced here and those published values are updated when the state is intended to be changed.

godtools/App/Services/Localization/LocalizationServices.swift
```diff
@@ -109,4 +109,9 @@ class LocalizationServices {
         
         return key
     }
+    
+    func toolCategoryStringForBundle(bundle: Bundle, attrCategory: String) -> String {
```
levieggertcru, line 113: Hey @rachaelblue at first I wasn't sure about this change with it being viewModel (domain layer) logic, but then I realized most services have a mix of both domain layer and data layer logic already.
I'm going to think about that more and the possibility of creating more distinction between data layer logic and domain layer logic.  It would involve us either creating more use cases or viewModel extensions for sharing domain logic between viewModels.

levieggertcru (changes requested):
Hey @rachaelblue looking good.  Just a few little things.

godtools/App/Features/Tools/SwiftUI/AllTools/AllToolsContentViewModel.swift
```diff
@@ -21,10 +21,13 @@ class AllToolsContentViewModel: NSObject, ObservableObject {
     private let favoritedResourcesCache: FavoritedResourcesCache
     private let favoritingToolMessageCache: FavoritingToolMessageCache
     private let analytics: AnalyticsContainer
+        
+    private var spotlightViewModel: ToolSpotlightViewModel?
```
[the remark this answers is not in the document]
rachaelblue: Good call! For some reason I was thinking it would error because `dataDownloader` and all the other params needed to be initialized first.  But that makes sense that if it's `lazy`, the compiler knows all the instance variables will exist already.

godtools/App/Share/SwiftUI Views/TwoRowHGrid.swift
```diff
@@ -0,0 +1,54 @@
...
+import SwiftUI
+
+// TODO: (DEPRECATED) Once we stop supporting iOS 13, we can use a LazyHGrid instead
+@available(*, deprecated)
+struct TwoRowHGrid<Content: View>: View {
+    let itemCount: Int
+    let spacing: CGFloat
+    let content: (Int) -> Content
+
+    init(itemCount: Int, spacing: CGFloat, @ViewBuilder content:@escaping (Int) -> Content) {
```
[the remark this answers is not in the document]
rachaelblue: Yeah!!  Can make views so much more reusable if you don't have to know what the actual content is, just the layout.

godtools/App/Features/Tools/SwiftUI/ToolCategoriesView/ToolCategoryButtonView/ViewModels/BaseToolCategoryButtonViewModel.swift
rachaelblue: That makes sense, good catch!

godtools/App/Services/Localization/LocalizationServices.swift
rachaelblue: Yeah, I wasn't sure either 😄  The issue I saw was that we have a fair amount of logic that is repeated between viewModels.  You're right, it _is_ domain layer logic, so it is correct that it's in the viewModel, but it seems like there should be a separate place to store domain layer logic that multiple viewModels use.  Some of the logic is probably so ubiquitous that putting it in an extension or useCase seems like overkill to me-- maybe just have some static methods stashed somewhere central?

levieggertcru: For now I think the best thing is to create that Share/Domain folder and organize domain logic in there, that way it's easy to locate since there isn't a standard way to share domain logic.  It's a bit mixed at the moment.  So it will either live there or in Services at the moment. https://jira.cru.org/browse/GT-1624

levieggertcru (approved):

PR #1169, opened 2022-06-13
GT-1612 recreate tool details in swift UI for meta

levieggertcru:
This PR lays some ground work for new meta versions UI in tool details which I would like to do in SwiftUI.  This PR rebuilds the tool details view in SwiftUI.  The original UIKit ToolDetailView is still in the project and will removed in a separate Jira task.  This PR adds ToolDetailsView in SwiftUI.


godtools/App/Features/ToolDetails/Presentation/ToolDetails/ToolDetailsViewModel.swift
```diff
@@ -0,0 +1,252 @@
...
+            aboutDetailsValue = primaryTranslation.translatedDescription
+            languageBundle = localizationServices.bundleLoader.bundleForResource(resourceName: primaryLanguage.code) ?? Bundle.main
+        }
+        else if let englishTranslation = resourcesCache.getResourceLanguageTranslation(resourceId: resource.id, languageCode: "en") {
+            
+            nameValue = englishTranslation.translatedName
+            aboutDetailsValue = englishTranslation.translatedDescription
+            languageBundle = localizationServices.bundleLoader.englishBundle ?? Bundle.main
+        }
+        else {
```
rachaelblue, line 196: I see this pattern a lot:
```
if let primary language = ..., let primaryTranslation = ... {

} else if let englishTranslation = ... {

} else {
   (use resource itself)
}
```

I had been trying to figure out if there was a stop we could put this logic so we didn't have it in so many places, but I'm not sure where it would make sense.  I tried to do this for the language bundle logic in #1160, but maybe it should all go in some other class that holds helper logic like this?

godtools/App/Features/ToolDetails/Presentation/ToolDetails/ToolDetailsViewModel.swift
```diff
@@ -0,0 +1,252 @@
...
+        }
+                
+        name = nameValue
+        totalViews = String.localizedStringWithFormat(localizationServices.stringForBundle(bundle: languageBundle, key: "total_views"), resource.totalViews)
+        openToolButtonTitle = localizationServices.stringForBundle(bundle: languageBundle, key: "toolinfo_opentool")
+        learnToShareToolButtonTitle = localizationServices.stringForBundle(bundle: languageBundle, key: "toolDetails.learnToShareToolButton.title")
+        addToFavoritesButtonTitle = localizationServices.stringForBundle(bundle: languageBundle, key: "add_to_favorites")
+        removeFromFavoritesButtonTitle = localizationServices.stringForBundle(bundle: languageBundle, key: "remove_from_favorites")
+        aboutTitle = localizationServices.stringForBundle(bundle: languageBundle, key: "about")
+        versionsTitle = localizationServices.stringForBundle(bundle: languageBundle, key: "toolDetails.versions.title")
```
rachaelblue, line 210: What do you think about having a container to store all the string keys?  I tend to think having the literals is error-prone, and difficult to change if something needs to be adjusted in the future.  Something like
```
enum LocalizedStringKey {
  static let totalViews = "total_views"
  static let openToolButtonTitle = "toolinfo_opentool"
 ...etc...
}
```

godtools/App/Share/ColorPalette.swift
```diff
@@ -61,6 +67,14 @@ enum ColorPalette {
         return UIColor(red: 128 / 255, green: 130 / 255, blue: 132 / 255, alpha: 1.0)
     }
     
+    private func getMediaColor() -> UIColor {
+        return UIColor(red: 230 / 255, green: 230 / 255, blue: 230 / 255, alpha: 1.0)
+    }
+    
+    private func getPrimaryTextColor() -> UIColor {
+        return UIColor(red: 90 / 255, green: 90 / 255, blue: 90 / 255, alpha: 1.0)
```
rachaelblue, line 75: `.primaryTextColor` is the same as `.gtGrey`.  We could just return gtGrey in the primaryTextColor definition, instead of defining the same color twice.

godtools/App/Share/SwiftUI Preferences/ViewGeometry.swift
```diff
@@ -0,0 +1,19 @@
...
+
+import SwiftUI
+
+struct ViewGeometry: View {
+    
+    var body: some View {
+        
+        GeometryReader { geometry in
+            Color.clear
+                .preference(key: ViewSizePreferenceKey.self, value: geometry.size)
```
rachaelblue, line 16: I'm a little unclear what this is for.  It just adds a clear background to fill the space basically?

godtools/App/Share/SwiftUI Views/AnimatedSwiftUIView/AnimatedSwiftUIView.swift
```diff
@@ -0,0 +1,42 @@
...
+//
+//  Created by Levi Eggert on 6/10/22.
+//  Copyright © 2022 Cru. All rights reserved.
+//
+
+import Foundation
+import Lottie
+import SwiftUI
+
+struct AnimatedSwiftUIView: UIViewRepresentable {
```
rachaelblue, line 13: Was this working for you?  I wasn't seeing it animate-- Honor Restored should be one that does, right?

godtools/App/Share/SwiftUI Views/TextWithLinks/TextWithLinks.swift
```diff
@@ -0,0 +1,88 @@
...
+//  godtools
+//
+//  Created by Levi Eggert on 6/10/22.
+//  Copyright © 2022 Cru. All rights reserved.
+//
+
+import UIKit
+import SwiftUI
+
+struct TextWithLinks: UIViewRepresentable {
```
rachaelblue, line 12: Too bad that SwiftUI can't do this yet!

rachaelblue (commented):
Looks good!  I've been thinking about a few things as I've been working on my tickets and brought some of them up here, but I'm happy to make separate tickets to discuss further if it makes sense.

godtools/App/Features/ToolDetails/Presentation/ToolDetails/ToolDetailsViewModel.swift
```diff
@@ -0,0 +1,252 @@
...
+        reloadLearnToShareToolButtonState()
+    }
+    
+    deinit {
+        print("x deinit: \(type(of: self))")
+        favoritedResourcesCache.resourceFavorited.removeObserver(self)
+        favoritedResourcesCache.resourceUnfavorited.removeObserver(self)
+    }
+    
+    private var analyticsScreenName: String {
```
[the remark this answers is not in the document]
levieggertcru: Hey @rachaelblue I think that would be a great idea. 
Maybe too we have a protocol that all of our concrete ViewModel's must implement, something like ViewModelType, (this would be noted in programming guide in the ReadMe), and then anything we want to standardize across all viewModels it's just a matter of adopting that to our ViewModelType protocol.  So if we want all ViewModels to implement PageViewed protocol, just drop that on the ViewModelType. Thoughts?

godtools/App/Features/ToolDetails/Presentation/ToolDetails/ToolDetailsViewModel.swift
```diff
@@ -0,0 +1,252 @@
...
+    }
+    
+    private static func getMediaType(resource: ResourceModel, dataDownloader: InitialDataDownloader) -> ToolDetailsMediaType {
+        
+        let mediaType: ToolDetailsMediaType
+        
+        if !resource.attrAboutOverviewVideoYoutube.isEmpty {
+            
+            let playsInFullScreen: Int = 0
+            let playerParameters: [String: Any] = ["playsinline": playsInFullScreen]
```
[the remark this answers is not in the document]
levieggertcru: Hey @rachaelblue that's a good idea, we should have a central place for any strings.

godtools/App/Features/ToolDetails/Presentation/ToolDetails/ToolDetailsViewModel.swift
```diff
@@ -0,0 +1,252 @@
...
+        }
+        
+        let determineToolTranslationsToDownload = DetermineToolTranslationsToDownload(
+            resourceId: resource.id,
+            languageIds: [primaryLanguage.id],
+            resourcesCache: dataDownloader.resourcesCache,
+            languagesRepository: languagesRepository
+        )
+        
+        getToolTranslationsUseCase.getToolTranslations(determineToolTranslationsToDownload: determineToolTranslationsToDownload, downloadStarted: { [weak self] in
```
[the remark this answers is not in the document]
levieggertcru: Hey @rachaelblue I'll silence those warnings.  I have a Jira task as well for fixing warnings in the godtools target which I'd like to be done sometime in the near future.
https://jira.cru.org/browse/GT-1583

godtools/App/Features/ToolDetails/Presentation/ToolDetails/ToolDetailsViewModel.swift
```diff
@@ -0,0 +1,252 @@
...
+                    if let primaryLanguageTranslationManifest = toolTranslations.languageTranslationManifests.first {
+                        
+                        hidesLearnToShareToolButtonValue = primaryLanguageTranslationManifest.manifest.tips.isEmpty
+                    }
+                    else {
+                        
+                        hidesLearnToShareToolButtonValue = true
+                    }
+                    
+                case .failure(let error):
```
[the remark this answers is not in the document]
levieggertcru: Hey @rachaelblue I'll go ahead and silence that warning since error is not being used.  Might be best to keep warnings like those silenced if not being used for anything.

godtools/App/Features/ToolDetails/Presentation/ToolDetails/ToolDetailsViewModel.swift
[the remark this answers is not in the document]
rachaelblue: We could run into issues with ViewModelType including analytics, since some of the subview viewModels don't do their own analytics (or are just part of a page and wouldn't have a pageViewed event).  But I do like the idea of having a ViewModelType, I'm just not sure if there's anything that stands out as being true for _all_ viewModels.  Worth thinking about though!

godtools/App/Features/ToolDetails/Presentation/ToolDetails/ToolDetailsViewModel.swift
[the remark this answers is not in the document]
rachaelblue: Sounds good!

godtools/App/Features/ToolDetails/Presentation/ToolDetails/ToolDetailsViewModel.swift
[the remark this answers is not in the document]
levieggertcru: That's a good point about the sub viewModels.  Let's scratch the ViewModelType and not make things too complicated.

godtools/App/Share/SwiftUI Views/TextWithLinks/TextWithLinks.swift
levieggertcru: I know I wish there was something provided in SwiftUI.  In iOS 15 url links are supported automatically from Text elements, however, I don't think you can detect the URL tap itself to perform any custom logic.  It just automatically links out to Safari.

godtools/App/Features/ToolDetails/Presentation/ToolDetails/ToolDetailsViewModel.swift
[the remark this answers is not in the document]
levieggertcru: With use case extensions we can create a SharedUseCaseType+PageViewed extension which performs the analytics tracking call.
```func pageViewed(pageViewedAnalytics: PageViewedAnalytics, properties: TrackScreenModel)```.

File: godtools/App/Features/ToolDetails/Presentation/ToolDetails/ToolDetailsViewModel.swift (hunk 2 above)
Comment (levieggertcru) [reply godtools/App/Features/ToolDetails/Presentation/ToolDetails/ToolDetailsViewModel.swift:210]: Hey @rachaelblue created a JIRA task for this one here: https://jira.cru.org/browse/GT-1621

File: godtools/App/Share/ColorPalette.swift (hunk 3 above)
Comment (levieggertcru) [reply godtools/App/Share/ColorPalette.swift:75]: Hey @rachaelblue the color naming gets a little tricky to me. Because maybe some UI elements share a color, however, we might want to create a specific color tied to different types of UI elements.  That way changing a color only effects those specific UI elements, not all elements that share that color.  Does that make sense?  Does ```gtGrey``` get applied to anything other than text and anything that might not be primary text?  I'm not sure if ```primaryTextColor``` is even a good name because what elements would be considered as primary text?  Maybe this is where it's best to have a designer standardize the app color theme that we just mimic?  What are your thoughts here and in other apps you've worked on how have you solved those questions?

File: godtools/App/Features/ToolDetails/Presentation/ToolDetails/ToolDetailsViewModel.swift [hunk 11]
```diff
@@ -0,0 +1,252 @@
...
+                self?.hidesLearnToShareToolButton = hidesLearnToShareToolButtonValue
+            }
+        })
+    }
+    
+    private func reloadFavorited() {
+                
+        if favoritedResourcesCache.isFavorited(resourceId: resource.id) {
+            hidesAddToFavoritesButton = true
+            hidesRemoveFromFavoritesButton = false
```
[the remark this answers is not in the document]
levieggertcru: Hey @rachaelblue I think this case would be better to have an enum rather than two Bool values.  I'm going to look into cleaning this one up, because either one or the other state will always be present.

godtools/App/Features/ToolDetails/Presentation/ToolDetails/ToolDetailsViewModel.swift
levieggertcru: Hey @rachaelblue this looks like another place where we could use shared use case extensions.  I did add a getPrimaryLanguageBundle to this PR here (https://github.com/CruGlobal/godtools-swift/pull/1176).   I can create separate use cases for the translated name and description.

godtools/App/Share/SwiftUI Views/AnimatedSwiftUIView/AnimatedSwiftUIView.swift
levieggertcru: Hey @rachaelblue just checked this in Staging and animations are playing, but the frame isn't resizing to fit.  Going to fix this in this PR.

godtools/App/Share/ColorPalette.swift
levieggertcru: Other than those questions, I think returning ```gtGrey``` from ```primaryTextColor``` is a good idea.

rachaelblue: Totally get ya, this is always a conundrum.  I tend to steer away from naming colors for places they're used, like `primaryTextColor`, because so often the color ends up changing in some areas but not others, and then the name is no longer accurate, but it rarely gets corrected because the color itself is still in use.  Also, it's harder to figure out what the color is _supposed_ to be when it's named after the area where it's used.  And then, more often than not, there's just a graveyard of colors that may or may not be used, but no one touches them because no one knows which we still need.

I think, ideally, a designer would standardize the app colors, and we'd name them things like gtGrey, gtBlue, etc. There shouldn't be _that_ many colors (and there really aren't now, either), and they shouldn't really change unless the entire app is being rebranded.  And if there's only a few core colors to choose from, I think the `primaryTextColor` naming scheme becomes unnecessary.  If you need a little more variation, names like `gtPrimaryGrey` and `gtSecondaryGrey` help distinguish without getting too specific.

levieggertcru: Hey @rachaelblue this is good input because I would have created a graveyard of colors.  Makes more sense to keep naming less specific.

godtools/App/Features/ToolDetails/Presentation/ToolDetails/ToolDetailsViewModel.swift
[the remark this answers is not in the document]
levieggertcru: Okay updated the viewModel here to have ```@Published var isFavorited: Bool```.  Also, since the button design was similar I moved button creation logic into ```ToolDetailsToggleFavoriteButton.swift```.

godtools/App/Features/ToolDetails/Presentation/ToolDetails/ToolDetailsViewModel.swift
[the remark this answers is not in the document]
levieggertcru: I'm going to think about this some more, I know I would like to have a use case for this logic that way it isn't duplicated across viewModels.

godtools/App/Features/ToolDetails/Presentation/ToolDetails/ToolDetailsViewModel.swift
levieggertcru: Definitely don't want to be duplicating code.  I'll create a separate task for this and look into consolidating that.

rachaelblue (approved):
Looks great!

PR #1175, opened 2022-06-16
GT-1509 all tools fixes

rachaelblue:
https://jira.cru.org/browse/GT-1509?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&showAll=true

Addressing Jill's comments about changes to Figma:
- Could we change the margin on the left and right of the "All Tools" tools to have less white space? This would match the spacing in Favorites and make the amount of white space the same between tools vertically and on between the tool and edge of screen. I talked with Justin and this was an error on the Figma design.
- For the parallel language the Figma design has the checkmark after the language name however it actually needs to be flipped to remain consistent with the favorites page.

@levieggertcru - this is another spot where there's reused domain layer logic (placing the checkmark in front of the parallel language).  I went the extension route, but not sure if there's a better place for it.


levieggertcru:
Hey @rachaelblue , for now we'll have to leave shared domain logic as is (mixed in services, extensions, etc.).  I'm planning to create a Share/Domain directory and move some of the domain logic that is currently scattered through the Share directory into there just to have it in 1 place. (https://jira.cru.org/browse/GT-1624)

levieggertcru (approved):

rachaelblue:
@levieggertcru ok that sounds good!  I'll make sure to put domain logic in there in the future.

PR #1176, opened 2022-06-16
GT - Create SharedUseCaseType protocol

levieggertcru:
Hey @rachaelblue what are your thoughts on something like this for sharing domain layer logic between viewModels?
- Added a protocol SharedUseCaseType
- Any domain layer logic we want to share, but don't exactly want to create a class for we can extend off of SharedUseCaseType and have a single method that acts as a use case and inject data dependencies into that method. 

With this we still get the benefit of having a screaming architecture since each extension will perform a single use case and be named accordingly to reflect that use case.

SharedUseCaseType extensions used across features would go in the Shared/Domain folder.  SharedUseCaseType extensions specific to a feature should go in the Feature/Domain folder.

Thoughts?


levieggertcru:
Hey @rachaelblue thinking about this more and the only downside is if we add any additional dependencies to a method argument or edit dependencies then any viewModels using the extension would have to be updated as well.  I don't see using extensions as a viable option here.

PR #1180, opened 2022-06-17
GT-1627 Use language translatedLanguageName in languages list

levieggertcru:
This PR uses the language's translated language name in the languages list.

### DIFF
--- a/godtools/App/AppDiContainer.swift
+++ b/godtools/App/AppDiContainer.swift
@@ -337,6 +337,13 @@ class AppDiContainer {
         return ShareToolScreenTutorialNumberOfViewsCache(sharedUserDefaultsCache: sharedUserDefaultsCache)
     }
     
+    func getToolLanguagesUseCase() -> GetToolLanguagesUseCase {
+        return GetToolLanguagesUseCase(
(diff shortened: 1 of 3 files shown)


godtools/App/AppDiContainer.swift
```diff
@@ -337,6 +337,13 @@ class AppDiContainer {
         return ShareToolScreenTutorialNumberOfViewsCache(sharedUserDefaultsCache: sharedUserDefaultsCache)
     }
     
+    func getToolLanguagesUseCase() -> GetToolLanguagesUseCase {
```
rachaelblue, line 340: I like having the use cases in here, but it does make me wonder how it'll work if we start using use cases more frequently, since most classes right now don't have a reference to AppDiContainer.

rachaelblue (commented):
Should I be seeing the languages here translated? 
 Screen Shot 2022-06-20 at 11 11 17 AM

levieggertcru:
Hey @rachaelblue the languages here are translated based on the device system language.

rachaelblue (approved):
Ok cool, that works!

PR #1186, opened 2022-06-22
GT-1636 fix content sections to render content stack instead of single text element

levieggertcru:
This PR makes a change to the accordion content section view.  Before, content sections would only render a single text element in the expanded view.  The single text element has been replaced with a content stack which makes this view more dynamic in that it can render a stack of content.
This change can be seen in Staging > The Four (USA) tool in the Did you pray this prayer? accordion content section view.

accordion_1

accordion_2


gyasistory (approved):
Everything looks good but there is an issue but I don't think it is with this PR.  When you click the version in the image below it doesn't change.

image

levieggertcru:
Hey @gyasistory thanks!  And yep that versions tab will be completed in GT-1418.

PR #1213, opened 2022-07-01
GT-1661 Add versions message text to tool details

levieggertcru:
This PR adds the message text in the versions section of tool details.

### DIFF
--- a/godtools/App/Features/ToolDetails/Subviews/ToolDetailsVersionsView.swift
+++ b/godtools/App/Features/ToolDetails/Subviews/ToolDetailsVersionsView.swift
@@ -16,6 +16,12 @@ struct ToolDetailsVersionsView: View {
         
         VStack(alignment: .leading, spacing: 0) {
             
+            Text(viewModel.versionsMessage)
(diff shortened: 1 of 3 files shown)


godtools/App/Features/ToolDetails/Subviews/ToolDetailsVersionsView.swift
```diff
@@ -16,6 +16,12 @@ struct ToolDetailsVersionsView: View {
         
         VStack(alignment: .leading, spacing: 0) {
             
+            Text(viewModel.versionsMessage)
+                .foregroundColor(ColorPalette.gtGrey.color)
+                .font(FontLibrary.sfProTextRegular.font(size: 16))
+                .multilineTextAlignment(.center)
```
rachaelblue, line 22: I think the text needs to be aligned left?

levieggertcru: Yes, that is supposed to be left aligned.

PR #1221, opened 2022-07-05
GT-1657 Fix navigation to tools from quick links

levieggertcru:
This PR fixes navigation to tools (lessons, favorites, all tools) from a quick link.  I moved the logic for navigating to the tools menu to occur before dismissing the onboarding flow.  This way when onboarding is dismissed a user is seeing the correct tools list.
Previously there was a navigation to tools menu happening before dismissing onboarding and another happening when dismissing onboarding completed.


rachaelblue (approved):
Quick links navigate correctly!

PR #1222, opened 2022-07-05
GT-1599 disable tutorial for non english

rachaelblue:
https://jira.cru.org/browse/GT-1599
Note: Latvian isn't a system language option on iOS, so the tutorial only shows for English.


gyasistory (approved):

godtools/App/AppDiContainer.swift
```diff
@@ -176,8 +176,7 @@ class AppDiContainer {
             snowplowAnalytics: SnowplowAnalytics(config: config, userAuthentication: userAuthentication, loggingEnabled: analyticsLoggingEnabled)
         )
                                                           
-        openTutorialCalloutCache = OpenTutorialCalloutUserDefaultsCache()
-                           
+        openTutorialCalloutCache = OpenTutorialCalloutUserDefaultsCache(getTutorialIsAvailableUseCase: GetTutorialIsAvailableUseCase(deviceLanguage: deviceLanguage))
```
levieggertcru, line 179: Hey @rachaelblue for this we don't want a useCase (domain layer) getting injected into a data layer class.  It's probably not very clear to you with the services folder containing mixed data and domain logic.  That's something I would like to eventually make more distinct in GodTools as we refactor those services. 
What's your schedule looking like today? It would be good to quickly connect on this one.

rachaelblue:
Closing this PR and will open a new one on a separate branch.

PR #1223, opened 2022-07-05
GT-1638 Changes the ToolSettings ReviewShareShareable Screen to a fullscreen with [x] close button

levieggertcru:
For this PR I was going to add an overlay to the preview share shareable screen, but decided against that and opted for a full screen with white background and close button.  Similar to the choose primary and parallel language lists from tool settings.  Having an overlay might not work with smaller devices or there wouldn't be much of an overlay at least.


levieggertcru:
Definitely would like to fix the blue background.  Maybe something with the safe area.  I'll take a look.

levieggertcru:
Hey @rachaelblue I reached out to Justin and he said to go with the rounded corner as well.  So 2 against 1.  I'm going to also update the languages list that is presented from ToolSettings to behave the same.

Simulator Screen Shot - iPhone 8 - 2022-07-06 at 10 33 24

PR #1244, opened 2022-07-08
GT-1683-explore-using-combine-with-optin-banner-use-case-enabled

levieggertcru:


levieggertcru:
Hey @rachaelblue I was able to do a little bit of research into using Combine and in this PR testing it out on GetOptInOnboardingBannerEnabledUseCase.  Here are the changes made:
- Add enabled publisher to OptInOnboardingBannerEnabledRepository.
- Add publisher to GetOptInOnboardingTutorialAvailableUseCase.
- Update GetOptInOnboardingBannerEnabledUseCase to combine those publishers and return a publisher of the result to ViewModel.
- Ended up adding a OptInOnboardingBannerEnabledCache to encapsulate the cache logic.  Also decided by default it should return null for enabled and have the UseCase determine the default starting value if null comes from the data layer.

godtools/App/Features/Tools/SwiftUI/Favorites/FavoritesContentViewModel.swift
```diff
@@ -70,7 +70,13 @@ class FavoritesContentViewModel: NSObject, ObservableObject {
         self.disableOptInOnboardingBannerUseCase = disableOptInOnboardingBannerUseCase
         
         super.init()
-        
+                
+        disableOptInOnboardingBannerSubscription = getOptInOnboardingBannerEnabledUseCase.getBannerIsEnabled()
+            .receive(on: DispatchQueue.main)
```
rachaelblue, line 75: So saying `receive(on: .main)` means you don't have to put `self?.hideTutorialBanner = !isEnabled` in an async block on main?  That's nice!

godtools/App/Share/Data/OptInOnboardingBannerEnabledRepository/OptInOnboardingBannerEnabledRepository.swift
```diff
@@ -7,16 +7,32 @@
...
 
 import Foundation
+import Combine
 
 class OptInOnboardingBannerEnabledRepository {
+        
+    private static let sharedPublishers: OptInOnboardingBannerEnabledRepositoryPublishers = OptInOnboardingBannerEnabledRepositoryPublishers()
     
-    let userDefaultKey = UserDefaultKeys.optInOnboardingBannerDisabled
+    private let cache: OptInOnboardingBannerEnabledCache
```
rachaelblue, line 16: It seems kinda overkill to me to have a whole cache for 1 value, and then a repository on top of that.  But I guess, in general, repositories would be pulling the data from a separate class anyway, not doing it themselves.  So maybe it fits the pattern better to have it separated out.  Just feels a little contrived in this case 😄

godtools/App/Share/Data/OptInOnboardingBannerEnabledRepository/OptInOnboardingBannerEnabledRepository.swift
```diff
@@ -7,16 +7,32 @@
...
+    func getEnabled() -> AnyPublisher<Bool?, Never> {
+                
+        return publishers.enabled.eraseToAnyPublisher()
+    }
+    
+    func storeEnabled(enabled: Bool) {
+        
+        cache.storeEnabled(enabled: enabled)
+        
+        publishers.enabled.send(enabled)
```
rachaelblue, line 36: My initial hope was that userDefaults would allow us to have a publisher on it directly, so that we didn't have to manually publish changes like this, they would just happen automatically when we store a new value in UserDefaults.  But it seems like UserDefaults itself doesn't have a way to support publishers.

godtools/App/Share/Domain/UseCases/GetOptInOnboardingBannerEnabledUseCase/GetOptInOnboardingBannerEnabledUseCase.swift
```diff
@@ -19,9 +20,16 @@ class GetOptInOnboardingBannerEnabledUseCase {
...
     }
         
-    func getBannerIsEnabled() -> Bool {
-        
-        return getOptInOnboardingTutorialAvailableUseCase.getOptInOnboardingTutorialIsAvailable() &&
-                optInOnboardingBannerEnabledRepository.getBannerIsEnabled()
+    func getBannerIsEnabled() -> AnyPublisher<Bool, Never> {
+                
+        return Publishers
+            .CombineLatest(getOptInOnboardingTutorialAvailableUseCase.getOptInOnboardingTutorialIsAvailable(), optInOnboardingBannerEnabledRepository.getEnabled())
```
rachaelblue, line 26: 🤯 sudden realization about why Combine is called Combine lol

godtools/App/Share/Domain/UseCases/GetOptInOnboardingTutorialAvailableUseCase/GetOptInOnboardingTutorialAvailableUseCase.swift
```diff
@@ -21,4 +22,10 @@ class GetOptInOnboardingTutorialAvailableUseCase {
         
         return deviceLanguage.isEnglish
     }
+    
+    func getOptInOnboardingTutorialIsAvailable() -> AnyPublisher<Bool, Never> {
+        
+        return Just(getOptInOnboardingTutorialIsAvailable())
```
rachaelblue, line 28: Oh cool!  I was wondering how you convert a single value into a publisher when there's not a stream of data.  That's handy.

rachaelblue (approved):
This is super cool!  I like how the use case pattern and combine work together.  My initial reaction was that it seemed really complicated for something that should be simple, but the pattern itself seems like something we can continue to use with more complicated logic as well.  And as I'm trying to learn how all the pieces relate, it probably helps that this was a simple scenario!

godtools/App/Features/Tools/SwiftUI/Favorites/FavoritesContentViewModel.swift
levieggertcru: Yeah much cleaner!

godtools/App/Share/Data/OptInOnboardingBannerEnabledRepository/OptInOnboardingBannerEnabledRepository.swift
levieggertcru: Yeah it does seem that way to me too, a lot of work for cacheing a single attribute.  I'm hoping that sticking to a consistent pattern will help when it comes to additional developers working in the code-base.

godtools/App/Share/Data/OptInOnboardingBannerEnabledRepository/OptInOnboardingBannerEnabledRepository.swift
levieggertcru: I would prefer this as well.  Maybe it's possible.  I was trying to get it to work originally and thought it wasn't working and then I switched to this setup, then I realized I wasn't doing Combine correctly in the UseCase.  I never got back to testing UserDefaults again.  I'm going to create a ticket to explore publishing those changes from UserDefaults.

godtools/App/Features/Tools/SwiftUI/Favorites/FavoritesContentViewModel.swift
levieggertcru: My hope is too if we have multiple network requests happening in a UseCase being able to use Combine to merge those results will be much cleaner.

PR #1247, opened 2022-07-11
GT-1693 add delete account screen

levieggertcru:
This PR adds an option to the Menu to Delete Account.  Tapping Delete Account will navigate to the DeleteAccountView with instructions on how to make a request to HelpDesk to delete your Okta account.  Tapping the help@cru.org will launch the native mail app (if available) with a prefilled message to send to the Help Desk.

menu-delete-account

delete-account-screen

delete-account-email

### DIFF
--- a/godtools/App/Flows/Menu/MenuFlow.swift
+++ b/godtools/App/Flows/Menu/MenuFlow.swift
@@ -221,7 +221,7 @@ class MenuFlow: Flow {
             let viewModel = MailViewModel(
                 toRecipients: ["help@cru.org"],
                 subject: "Please delete my account",
-                message: "I have created an account on the GodToolsapp and I would like to request that you delete my Okta account.",
+                message: "I have created an account on the GodTools app and I would like to request that you delete my Okta account.",
                 isHtml: false,
                 finishedSendingMailHandler: finishedSendingMail
             )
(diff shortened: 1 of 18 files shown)


godtools/App/Flows/Menu/MenuFlow.swift
```diff
@@ -197,6 +194,39 @@ class MenuFlow: Flow {
...
+        case .emailHelpDeskToDeleteOktaAccountTappedFromDeleteAccount:
+            
+            let finishedSendingMail = CallbackHandler { [weak self] in
+                self?.navigationController.dismiss(animated: true, completion: nil)
+            }
+            
+            let viewModel = MailViewModel(
+                toRecipients: ["help@cru.org"],
+                subject: "Please delete my account",
+                message: "I have created an account on the GodToolsapp and I would like to request that you delete my Okta account.",
```
rachaelblue, line 224: I'm assuming it's ok that this isn't translated.  Should there be a space in "GodToolsapp"?

rachaelblue (approved):
One question but otherwise looks good!

godtools/App/Flows/Menu/MenuFlow.swift
levieggertcru: Added a space and for now we don't need to since account creation is English only.

PR #1255, opened 2022-07-13
GT-1694 lessons in swift UI

rachaelblue:
https://jira.cru.org/browse/GT-1527
https://jira.cru.org/browse/GT-1694
https://jira.cru.org/browse/GT-1686


godtools/App/Share/StandardLeadingTrailingMargin.swift
```diff
@@ -0,0 +1,18 @@
...
+//  StandardLeadingTrailingMargin.swift
+//  godtools
+//
+//  Created by Rachael Skeath on 7/13/22.
+//  Copyright © 2022 Cru. All rights reserved.
+//
+
+import SwiftUI
+
+enum StandardLeadingTrailingMargin {
```
levieggertcru, line 11: Hey @rachaelblue I'm curious what your idea is for the StandardLeadingTrailingMargin?  Would this apply to just the lists in the ToolsMenu?
Or would this be a standard leading and trialing across the app?

rachaelblue: Good question-- that may be something to ask design, but my intention was more specific to Lists.  I can rename to reflect that.

levieggertcru: If this is logic pertaining directly to the lists in the home screen then I would suggest including this file within the Tools feature and maybe name something less generic.
However, if we do want some type of Shared place for styling Views (margins, padding, etc.) then maybe we can have a dedicated folder in the Share directory for that?  This is getting into an area I'm not too familiar with.  I can see the benefit though if there is a standard button height on buttons or standard margins to the edge of a screen across multiple screens of the app.  What are your thoughts and experience here?

rachaelblue: I was thinking it might make sense to go in the folder with `BackwardCompatibleList`, because that's really what it's used with, regardless of feature.  And rename it `BackwardCompatibleListLeadingTrailingMargin`.  

I think it's hard to have a shared place for things like margins and sizing just because it's so dependent on the designer to be consistent, and they often aren't 😄  I've seen that type of thing work with companies that have very strong branding guidelines-- specific colors, font styles and sizes, button looks, etc, and the designers knew they had to follow those exactly.  So unless we get to that point sometime in the future, I think it makes more sense to keep things like margin constants coupled with the view that they apply to.

levieggertcru (approved):

PR #1264, opened 2022-07-14
GT Begin adding programming guide to ReadMe to help with consistency in GodTools

levieggertcru:


README.md
```diff
@@ -1,35 +1,148 @@
...
+
+#### Presentation Layer
+
+The presentation layer makes up the Views and ViewModels.  In the GodTools app Views and ViewModels are organized by Feature in the Features folder.  Below will explain how presentation files are named and organized and conventions for Views and ViewModels.
+
+- [File Naming and Organization](#file-naming-and-organization)
+- [Views](#views)
+- [ViewModels](#viewmodels)
+
+##### File Naming and Organization
```
rachaelblue, line 53: This section is a little hard to follow.  It might be easier to follow bullet points that mimic the structure with indentions?  Or a screenshot of some of the directories in Xcode for some of the examples you mention?  (Or both?).  Also would be good to mention if you want things sorted alphabetically (that's not something I've seen on other projects!).

📁**Features**
  - 📁 Lessons
   --   📁 LessonList
    -----  📄 LessonList.swift
    -----  📄 LessonListViewModel.swift
    -----  📁 Subviews
     --------   📄 LessonListTitleView.swift

^ not formatted well but just an idea to help cut down on some of the text here.

README.md
```diff
@@ -1,35 +1,148 @@
...
-- Sending user inputs to the viewModel (button tap, entering text input, etc.).
+- Owns a ViewModel.
+- Observes ViewModel output.  In SwiftUI this is a combination of @ObservedObject and @Published properties.
+- Send inputs to the ViewModel (button tap, entering text input, etc.).
+
+File Naming and Organization:
+- All newly created views should be created in SwiftUI.
+- Views should have only 1 ViewModel.
+- Subviews that are static and help make up a screen (View) can point to the parent screen (View) ViewModel or they can have their own ViewModel.
+- Subviews that are dynamic such as views in collections (lists, stacks, etc.) should have there own ViewModel and only 1 ViewModel.
```
rachaelblue, line 95: This is a good distinction, I was going back and forth in a couple places on when to use the parent's viewModel and when to create a separate one.  Also some static subviews are so small that I didn't make a viewModel, just pass in the title (or whatever simple data) directly.

rachaelblue (commented):
This is awesome, thanks for writing this up!  One other thing that might be worth adding is a section about comments-- making sure comments are clean, and MARKs.  I use marks out of habit and tend to like seeing them on the minimap, but as viewModels become really small they may be unneeded.

README.md
```diff
@@ -1,35 +1,148 @@
...
+
+- [Summary](#summary)
+- [Presentation Layer](#presentation-layer)
+- [Domain Layer](#domain-layer)
+- [Data Layer](#data-layer)
+- [Coordinator](#coordinator)
+
+
+#### Summary
+The architectural structure for the GodTools app can be summarized as following a Clean Architecture Pattern (Presentation Layer, Domain Layer, an Data Layer), along with a Coordinator Pattern (Navigation decisions and logic), and SOLID principles(https://www.geeksforgeeks.org/solid-principle-in-programming-understand-with-real-life-examples/) .
```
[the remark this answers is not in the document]
levieggertcru: That's a good idea.  I'll see if I can dig up some of the older resources I had read on these and include them here as well.

README.md
```diff
@@ -1,35 +1,148 @@
...
+- [Architecture](#architecture)
+
+#### Conventions
+
+- [Classes](#classes)
+
+
+#### Classes
+
+- Class / Struct attributes should always be declared with the type.
```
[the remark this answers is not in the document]
levieggertcru: This one is just more personal preference.  One company I had worked with in the past had a similar guideline and I'm used to it from there too.

It helps me quickly understand the Class attributes and types by seeing the name and type side by side, just the way my eye automatically looks there.  Some types aren't as quick to discern either like (Float, Int, Double).  There are some types that would require a declaration such as optional types and any immutable instance that is instantiated on init().  Just a preference though when it comes to my eyes viewing class level attribute and types.

README.md
levieggertcru: I think a visual would be helpful here too.  I will make some updates here.

README.md
levieggertcru: Here I prefer static SubViews pointing to the parent (Screen) ViewModel.  It will make for less viewModel layers to jump through and less viewModel layers to pass dependencies through.

PR #1266, opened 2022-07-14
GT Minor AppDiContainer Clean Up

levieggertcru:
- Removed the CrashReporting protocol and class because it's not being used.
- Removed the IsNewUserCacheType protocol. Would like to go the Repository route for data storage and don't see a need to create an interface for every class that persists data when it's unlikely to change and unlikely to have different concrete types.


rachaelblue (approved):
Agreed on moving away from protocols in places where we likely would never have another type conforming to it 👍

PR #1267, opened 2022-07-18
GT-1688 lessons availability message

rachaelblue:
https://jira.cru.org/browse/GT-1688

- I created a new use case for getting the language availability string + checkmark or "X" and removed the extension on LanguageViewModel
- Note that the "language not available" is only supposed to show on lessons whose _primary_ language isn't available, whereas tool cards only show the parallel language available notice, never parallel language unavailable.


godtools/App/Share/Domain/UseCases/GetLanguageAvailabilityStringUseCase/GetLanguageAvailablityStringUseCase.swift
```diff
@@ -0,0 +1,44 @@
...
+    
+    let localizationServices: LocalizationServices
+    let getTranslatedLanguageUseCase: GetTranslatedLanguageUseCase
+    
+    init(localizationServices: LocalizationServices, getTranslatedLanguageUseCase: GetTranslatedLanguageUseCase) {
+        self.localizationServices = localizationServices
+        self.getTranslatedLanguageUseCase = getTranslatedLanguageUseCase
+    }
+    
+    func getLanguageAvailability(for resource: ResourceModel, language: LanguageModel?) -> (isAvailable: Bool, string: String) {
```
levieggertcru, line 21: Hey @rachaelblue the only change I would suggest is not providing a return tuple type here and instead provide a DomainModel to wrap those business attributes.  However, leave the tuple for now because this UseCase will be re-visited when the clean up sprint happens.  This UseCase may even go away and the logic in here will be part of a bigger UseCase for fetching a ToolDomainModel which I will explain at a later time.

rachaelblue: Sounds good!

PR #1277, opened 2022-07-25
GT-1589 shareable analytics

rachaelblue:
https://jira.cru.org/browse/GT-1589


godtools/App/Share/Domain/UseCases/GetShareableImageUseCase/ShareableImageModel.swift
```diff
@@ -0,0 +1,17 @@
...
+//  ShareableImageModel.swift
+//  godtools
+//
+//  Created by Rachael Skeath on 7/25/22.
+//  Copyright © 2022 Cru. All rights reserved.
+//
+
+import UIKit
+
+struct ShareableImageModel {
```
levieggertcru, line 11: Hey @rachaelblue can you change this name to ```ShareableImageDomainModel```.  That is something I would like to shift towards for our domain and data models.  Suffixing with either ```DomainModel``` or ```DataModel```.  That will help us see which is coming from the domain vs data layer.

godtools/App/Share/Domain/UseCases/GetShareableImageUseCase/GetShareableImageUseCase.swift
```diff
@@ -0,0 +1,25 @@
...
+    
+    func getShareableImage(from shareable: Shareable, manifestResourcesCache: ManifestResourcesCache) -> ShareableImageModel? {
+        guard let shareableImage = shareable as? ShareableImage, let resource = shareableImage.resource, let imageToShare = manifestResourcesCache.getImageFromManifestResources(resource: resource) else {
+            return nil
+        }
+        
+        return ShareableImageModel(
+            image: imageToShare,
+            imageId: shareable.id,
+            toolAbbreviation: shareable.manifest.code
```
levieggertcru, line 22: I noticed that the ```manifest.code``` is an optional value.  Would it be safer to fetch the abbreviation from the resource?  Or fallback to the resource abbreviation if manifest code is null?  @frett thoughts here.

levieggertcru (changes requested):

godtools/App/Share/Domain/UseCases/GetShareableImageUseCase/GetShareableImageUseCase.swift
frett: on Android I just pass it as null if it doesn't exist. We've been actively publishing the code and the locale in manifests for several years now, it's just older translations that may be missing them.

If you want to fallback to the what is set in your local resource, that's fine

frett: https://github.com/CruGlobal/mobile-content-api/pull/159

For reference, we've been including those attributes for almost 4 years now when publishing a translation

frett: https://github.com/CruGlobal/mobile-content-api/pull/159

For reference, we've been including those attributes for almost 4 years now when publishing a translation

levieggertcru: I'm fine with mimicking Android here.

godtools/App/Share/Domain/UseCases/GetShareableImageUseCase/ShareableImageModel.swift
rachaelblue: Ooo good idea, I was thinking about that too.  That'll help make things clearer for sure.

levieggertcru (approved):

PR #1278, opened 2022-07-25
GT-1597 quick links other languages

rachaelblue:
https://jira.cru.org/browse/GT-1597
https://jira.cru.org/browse/GT-1656

### DIFF
--- a/godtools/App/Share/Domain/UseCases/GetDeviceLanguageCodeUseCase.swift
+++ b/godtools/App/Share/Domain/UseCases/GetDeviceLanguageCodeUseCase.swift
@@ -1,0 +1,26 @@
+//
+//  GetDeviceLanguageCodeUseCase.swift
+//  godtools
+//
+//  Created by Rachael Skeath on 7/25/22.
+//  Copyright © 2022 Cru. All rights reserved.
+//
+
+import Foundation
+
+class GetDeviceLanguageCodeUseCase {
+    
+    private let deviceLanguage: DeviceLanguageType
+    
+    init(deviceLanguage: DeviceLanguageType) {
+        
+        self.deviceLanguage = deviceLanguage
+    }
+    
+    func getDeviceLanguageCode() -> String {
+        
+        let deviceLocale: Locale = deviceLanguage.preferredLocalizationLocale ?? deviceLanguage.locale
+        
+        return (deviceLocale.languageCode ?? deviceLocale.identifier).lowercased()
+    }
+}
--- a/godtools/App/Share/LanguageCodes.swift
+++ b/godtools/App/Share/LanguageCodes.swift
@@ -1,0 +1,18 @@
+//
+//  LanguageCodes.swift
+//  godtools
+//
+//  Created by Rachael Skeath on 7/25/22.
+//  Copyright © 2022 Cru. All rights reserved.
+//
+
+import Foundation
+
+enum LanguageCodes {
+    
+    static let English = "en"
+    static let French = "fr"
+    static let Latvian = "lv"
+    static let Vietnamese = "vi"
+    
+}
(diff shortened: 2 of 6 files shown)


godtools/App/Share/LanguageCodes.swift
```diff
@@ -0,0 +1,18 @@
...
+//  LanguageCodes.swift
+//  godtools
+//
+//  Created by Rachael Skeath on 7/25/22.
+//  Copyright © 2022 Cru. All rights reserved.
+//
+
+import Foundation
+
+enum LanguageCodes {
```
levieggertcru, line 11: This is definitely needed.  I know there are some other places in the app where we have hard-coded language codes which should pull from here.

Couple things:
- Can you use a ```struct``` or ```class``` instead of an enum since we aren't using cases.

- Can you make those constants lowerCamelCase to mimic swift conventions. 

https://www.swift.org/documentation/api-design-guidelines/#general-conventions

godtools/App/Share/Domain/UseCases/GetDeviceLanguageCodeUseCase.swift
```diff
@@ -0,0 +1,26 @@
...
+//  GetDeviceLanguageCodeUseCase.swift
+//  godtools
+//
+//  Created by Rachael Skeath on 7/25/22.
+//  Copyright © 2022 Cru. All rights reserved.
+//
+
+import Foundation
+
+class GetDeviceLanguageCodeUseCase {
```
levieggertcru, line 11: Nice we are definitely needing this too.  I'm even thinking of phasing out DeviceLanguageType and moving that logic directly in here, but for now this is a start.

godtools/App/Share/Domain/UseCases/GetDeviceLanguageCodeUseCase.swift
```diff
@@ -0,0 +1,26 @@
...
+    
+    private let deviceLanguage: DeviceLanguageType
+    
+    init(deviceLanguage: DeviceLanguageType) {
+        
+        self.deviceLanguage = deviceLanguage
+    }
+    
+    func getDeviceLanguageCode() -> String {
```
levieggertcru, line 20: I have some additional thoughts, instead of this UseCase being specific to a LanguageCode, how about having a general ```GetDeviceLanguageUseCase```.

For the return type instead of a ```String```, a ```DeviceLanguageDomainModel``` which can encapsulate the languageCode.  This would give us some room to attach additional device language related attributes to our domain model if needed.

Thoughts?

levieggertcru (changes requested):

godtools/App/Share/LanguageCodes.swift
levieggertcru: Hey @rachaelblue update here, I think enum is okay here and has the added benefit of not being instantiated. 

Reading through this StackOverflow post here -> https://stackoverflow.com/questions/24244326/swift-global-constant-naming-convention/38078903#38078903

rachaelblue: Yeah, I use `enum` when I essentially want a namespace and not something instantiable.  But I also see the perspective that that's not what `enum` was designed for.  It would be nice for Swift to have something separate specifically for namespaces.

godtools/App/Share/Domain/UseCases/GetDeviceLanguageCodeUseCase.swift
rachaelblue: That sounds good!  That'll make it easy to build on in the future.  I'm wary of having millions of tiny useCases that do almost nothing, so grouping things together when reasonable will help that.

levieggertcru: Yeah I am wary of that as well.  It's something we will most likely have to balance and there may be situations where it's okay not to use a UseCase and directly inject data dependencies into the ViewModel.  We'll have explore that as we go.

levieggertcru (approved):

PR #1283, opened 2022-07-26
upgrade shared parser to 0.6.1-SNAPSHOT

frett:
This includes the new accordion analytics properties

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -13,7 +13,7 @@ use_frameworks!
 def shared_pods
   
   # CruGlobal pods
-  pod 'GodtoolsToolParser', '0.6.0'
+  pod 'GodtoolsToolParser', '0.6.1-SNAPSHOT'
 end
 
 target 'godtools' do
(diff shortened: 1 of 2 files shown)


Podfile
```diff
@@ -13,7 +13,7 @@ use_frameworks!
 def shared_pods
   
   # CruGlobal pods
-  pod 'GodtoolsToolParser', '0.6.0'
+  pod 'GodtoolsToolParser', '0.6.1-SNAPSHOT'
```
rachaelblue, line 16: What does `snapshot` mean?

rachaelblue (approved):

Podfile
frett: it's a concept I took from java maven versioning conventions. SNAPSHOT is a moving target for a build at the tip of the current development branch.

So, `SNAPSHOT` versions can change frequently, but they allow you to access new features before they end up in a stable (unchanging) release.

I try to make sure and create a stable release of the shared parser before we release iOS or Android apps to the store, that way we in theory could always build older versions of the app from git if we need to debug something that happens on a specific version of the app.

frett: https://stackoverflow.com/a/5901460/4721910

PR #1286, opened 2022-07-28
GT-1733 divide by 0 check

rachaelblue:
https://jira.cru.org/browse/GT-1733


godtools/App/Services/Renderer/Views/MobileContent/Views/ContentStack/MobileContentStackView.swift
```diff
@@ -647,13 +647,20 @@ extension MobileContentStackView {
             constrainLeadingToSuperviewLeading = true
             constrainTrailingToSuperviewTrailing = true
             
+            let multiplier: CGFloat
```
levieggertcru, line 650: Hey @rachaelblue for the fallback multiplier can you have it fallback to 1 instead.  With a value of 1 I believe it would still maintain the original height in size.height.

rachaelblue: Sure.  I wasn't sure that height made sense if width is zero.  It's the difference between a big blank space and a small blank space 🤷‍♀️

levieggertcru: Hey @rachaelblue! Hmm, I don't like the idea of having that additional space either.

How about this.  Either:
1. The view isn't rendered in the stack if it has an invalid height (height resolves to 0).
2. Decrease the multiplier to something smaller, say 0.1.  That would cut down on some of the added space.

My main worry with having a multiplier of 0 is that the view height will then resolve to 0 which I believe would cause issues with the MobileContentStackView and how it determines its overall height.

The MobileContentStackView's overall height is determined by all of its children where each child view has a top constraint and bottom constraint pushing the overall height of the stack.  I vaguely remember autolayout having issues in this scenario if a child view's height was 0.  Maybe this isn't an issue anymore, that was the main reason I was thinking that the multiplier should be something greater than 0.  Do you know if there would be an issue here with constraints breaking?

There's also some logic in the relayoutForSpacerViews which would benefit from having the view removed entirely or visibility state set to gone.  Otherwise we will get some extra space added when that calculation is done.  See here -> https://github.com/CruGlobal/godtools-swift/blob/develop/godtools/App/Services/Renderer/Views/MobileContent/Views/ContentStack/MobileContentStackView.swift#L449

Now that I think about it, we could set the visibilityState to ```gone``` on the MobileContentAnimationView.  Take a look at ```visibilityState: MobileContentViewVisibilityState``` on ```MobileContentView```.

Maybe in ```MobileContentAnimationView``` we have something like this?
```
override var visibilityState: MobileContentViewVisibilityState {
        get {
            let isInvalidHeight: Bool = //check for invalid height.
            return isInvalidHeight ? .gone : super.visibilityState
        }
    }
```

rachaelblue: Sorry I'm just finally getting to this!  Setting the `visibilityState` like you have it didn't work for me, that getter never got called, and I was afraid to mess with all the stuff that `visibilityState` seems to do.  So instead I changed the animation asset size so it doesn't default to 0 if the animationData itself is nil.

I also left the multiplier in there but changed it to 0.1, so in the case that there ever is a 0 value, it still won't crash and the empty space is pretty small.  (Which there shouldn't be a 0 value, but better to leave the catch in there and fail gracefully)

levieggertcru: Okay dang I was hoping that override would work.  I think this is fine though with it being such a small case.

levieggertcru (approved):

levieggertcru (approved):

PR #1287, opened 2022-07-29
GT-1605 Accordion Analytics

rachaelblue:
https://jira.cru.org/browse/GT-1605


godtools/App/Services/Renderer/Views/MobileContent/Views/ContentAccordionSection/MobileContentAccordionSectionView.swift
```diff
@@ -86,6 +86,12 @@ class MobileContentAccordionSectionView: MobileContentView, NibBased {
     
     private func toggleContentHidden() {
         setContentHidden(hidden: !contentIsHidden, animated: true)
+        
+        if contentIsHidden {
```
levieggertcru, line 90: Hey @rachaelblue you'll want this logic in the ```setContentHidden``` method  since```toggleContentHidden``` method funnels through there.

rachaelblue: Good catch, I was thinking it should just fire when a user tap occurs, but it should fire whenever it's open regardless of how it happened.

levieggertcru (approved):

PR #1288, opened 2022-08-01
GT-1448 Complete GetToolTranslationsUseCase for downloading tool language translations

levieggertcru:


levieggertcru:
Hey @rachaelblue this is another pretty large PR.

- Cleaned up TranslationsRepository.  Also realized the GodToolsToolParser Manifest can't be shared between threads.  I think this is because of the watchers. 
- Added LanguageSettingsRepository which will replace LanguageSettingsService.
- Added UseCases for working with the language settings repository. 
- Completed GetToolTranslationsUseCase for downloading tool language translations.
- Added downloading of attachment files to AttachmentsRepository.

godtools/App/Share/Data/LanguageSettingsRepository/Cache/LanguageSettingsCache.swift
```diff
@@ -0,0 +1,92 @@
...
+        storeParallelLanguage(id: nil)
+    }
+}
+
+private extension UserDefaults {
+    
+    private static let primaryLanguageCacheKey: String = "kPrimaryLanguageId"
+    private static let parallelLanguageCacheKey: String = "kParallelLanguageId"
+    
+    @objc dynamic var primaryLanguageId: String? {
```
rachaelblue, line 69: Nice to have these in UserDefaults!

godtools/App/Share/Data/TranslationsRepository/Api/MobileContentTranslationsApi.swift
```diff
@@ -36,14 +36,21 @@ class MobileContentTranslationsApi {
         )
     }
     
-    func getTranslationFile(fileName: String) -> AnyPublisher<URLResponseObject, Error> {
+    func getTranslationFile(fileName: String) -> AnyPublisher<URLResponseObject, URLResponseError> {
         
         return session.dataTaskPublisher(for: getTranslationFileRequest(fileName: fileName))
-            .map {
-                return URLResponseObject(data: $0.data, urlResponse: $0.response)
+            .tryMap {
```
rachaelblue, line 42: Haven't seen `tryMap` before, that's cool!

godtools/App/Share/Data/TranslationsRepository/TranslationsRepository.swift
```diff
@@ -34,195 +34,123 @@ class TranslationsRepository {
...
-            
-            return Fail(error: NSError.errorWithDescription(description: errorDescription))
-                .eraseToAnyPublisher()
+        let requests = translations.map {
+            self.downloadAndCacheTranslationFiles(translation: $0)
         }
         
-        return Just(translation.manifestName).setFailureType(to: Error.self)
+        return Publishers.MergeMany(requests)
+            .collect()
```
rachaelblue, line 79: Does `collect` wait until all the publishers in the merge have a response, and then notify subscribers only once with all the values?

godtools/App/Share/Domain/UseCases/GetDeviceLanguageUseCase/GetDeviceLanguageUseCase.swift
```diff
@@ -9,20 +9,36 @@
 import Foundation
 
 class GetDeviceLanguageUseCase {
+        
+    private let getLanguageUseCase: GetLanguageUseCase
```
rachaelblue, line 13: `GetDeviceLanguageUseCase` doesn't seem to be using `GetLanguageUseCase`.  Should it?

godtools/App/Features/SetupParallel/Views/ParallelLanguageListViewModel.swift
```diff
@@ -8,108 +8,57 @@
...
-                }
-            }
-        }
-        
-        dataDownloader.resourcesUpdatedFromRemoteDatabase.addObserver(self) { [weak self] (error: InitialDataDownloaderError?) in
-            DispatchQueue.main.async { [weak self] in
-                if error == nil {
-                    self?.reloadData()
-                }
+        Publishers.Zip(getLanguagesListUseCase.getLanguagesList(), getSettingsPrimaryLanguageUseCase.getPrimaryLanguage())
```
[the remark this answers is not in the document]
levieggertcru: Hey @rachaelblue I had to double check this one, found these marble diagrams that do a good job explaining the difference.
https://rxmarbles.com/#zip
https://rxmarbles.com/#combineLatest

godtools/App/Share/Data/TranslationsRepository/TranslationsRepository.swift
levieggertcru: Hey @rachaelblue from my understanding the collect will gather all of the values of the upstream publishers once those have all completed and completed successfully.  It seems if any of the upstream publishers fail, then the entire thing immediately fails.

https://developer.apple.com/documentation/combine/publishers/mergemany/collect()

godtools/App/Share/Domain/UseCases/GetDeviceLanguageUseCase/GetDeviceLanguageUseCase.swift
levieggertcru: I was contemplating that as well.  There are some differences between them though.
- The LanguageDomainModel is mapped from mobile content api languages with applied business logic.
- Where as DeviceLanguageDomainModel is mapped from the OS System Device languages, in which we aren't applying any special business logic at this time.  We mostly use this for enabling specific device language features in the app.  

There is a relationship between the 2 of these where the language codes from our languages in mobile content api match a Locale identifier.

For now, let's just keep them distinct.  This way we can grab the current device language without depending on the mobile content languages api.

godtools/App/Features/SetupParallel/Views/ParallelLanguageListViewModel.swift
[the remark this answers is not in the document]
frett: they are standard reactive stream operators, `zip` will take elements in pairs. `combineLatest` will combine the latest value from both streams whenever either stream emits a value.

`Zip(['a', 'b', 'c'], [0, 1])` will process the pairs `['a',0], ['b',1]`. if the second stream never produces a 3rd element, then `c` will never be processed from the first stream.

`combineLatest` will always trigger when a new value is received.

the marbles diagrams are great at visualizing this.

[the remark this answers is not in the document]
frett: I've actually not found a usage for zip anywhere in Android yet, usually I'm looking for `combineLatest` to transform 2 streams into a single stream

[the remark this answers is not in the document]
levieggertcru: Not sure if Zip really makes much sense here now.  Seems I would always want the latest of these.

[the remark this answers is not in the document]
levieggertcru: Thanks for the added input Daniel!  Going to update to use latest.

godtools/App/Features/Languages/ChooseLanguage/ChooseLanguageViewModel.swift
```diff
@@ -34,39 +44,43 @@ class ChooseLanguageViewModel: NSObject, ChooseLanguageViewModelType {
...
         case .primary:
             navTitle.accept(value: localizationServices.stringForMainBundle(key: "primary_language"))
         case .parallel:
             navTitle.accept(value: localizationServices.stringForMainBundle(key: "parallel_language"))
         }
         
-        reloadData()
-        
-        setupBinding()
+        Publishers.Zip3(getLanguagesListUseCase.getLanguagesList(), getSettingsPrimaryLanguageUseCase.getPrimaryLanguage(), getSettingsParallelLanguageUseCase.getParallelLanguage())
```
rachaelblue, line 71: Does this need to be `CombineLatest` then, too?  Seems like it should be. Can you combine more than 2 streams?

godtools/App/Share/Domain/UseCases/GetDeviceLanguageUseCase/GetDeviceLanguageUseCase.swift
rachaelblue: Makes sense to me.  Can we remove this useCase dependency then?

godtools/App/Features/Languages/ChooseLanguage/ChooseLanguageViewModel.swift
levieggertcru: Hey @rachaelblue this should also use CombineLatest.   I will include this in GT-1746

rachaelblue (approved):

PR #1291, opened 2022-08-03
GT-1735 remove mock view models 2

rachaelblue:
https://jira.cru.org/browse/GT-1735

I removed the `mock` viewModels, but the `base` viewModels are currently acting like protocols in that they have subclasses that use shared behavior from the base/parent class.  For example, `BaseFavoriteToolsViewModel` is being subclassed by `AllFavoriteToolsViewModel` and `FavoriteToolsViewModel`, since both need logic for pulling out favorite tools.  

It would make more sense to have actual protocols instead, but SwiftUI won't work with protocols.  So all this to say, I think it makes sense to keep the base VMs until/unless we find a better workaround.  I could potentially see the logic that's in `base` basically becoming a single UseCase in some instances, but that wouldn't solve the problem of reusable SwiftUI views accepting different viewModel types.


levieggertcru:
Hey @rachaelblue, let's merge this change for removing the mocks and revisit the base viewModels when starting more of the code refactoring to implement UseCases in all tools and favorites.  I see now the FavoritedList you're sharing.  I'm going to spend a little more time reviewing that.

In the mean time though, let's get this next sprint out before more of the code refactoring.  I did assign this task to you here -> https://jira.cru.org/browse/GT-1580.  This one is pretty involved.  I'm going to get refreshed on it and we can catch up and discuss.

levieggertcru (approved):

PR #1295, opened 2022-08-09
GT-1749 attachments repo for resource cards

rachaelblue:
https://jira.cru.org/browse/GT-1749


godtools/App/Share/Domain/UseCases/GetBannerImageUseCase/GetBannerImageUseCase.swift
```diff
@@ -0,0 +1,38 @@
...
+        
+        self.attachmentsRepository = attachmentsRepository
+    }
+    
+    func getBannerImagePublisher(for attachmentId: String) -> AnyPublisher<Image?, Never> {
+        
+        return attachmentsRepository.getAttachmentImage(id: attachmentId)
+    }
+    
+    func getBannerImage(for attachmentId: String) -> Image? {
```
levieggertcru, line 27: Hey @rachaelblue this PR looks good.  Eventually I would like this method to go away since this method will only work on cached banner images and is only useful for SwiftUI Previews.  Instead we can have a Development Assets Catalog for SwiftUI Previews.  This work doesn't isn't a showstopper for the next TestFlight release and can be done as part of the refactoring in this task here. -> https://jira.cru.org/browse/GT-1755

levieggertcru (approved):

godtools/App/Share/SwiftUI Views/LessonCardView/ViewModels/LessonCardViewModel.swift
```diff
@@ -130,6 +115,15 @@ extension LessonCardViewModel {
                 }
             }
         }
+        
+        getBannerImageUseCase.getBannerImagePublisher(for: resource.attrBanner)
```
levieggertcru, line 119: Hey @rachaelblue for this check out the assign() method https://developer.apple.com/documentation/combine/just/assign(to:on:).

You should be able to do something like this.

```
getBannerImageUseCase.getBannerImagePublisher(for: resource.attrBanner)
            .assign(to: \.bannerImage, on: self)
            .store(in: &cancellables)
```

godtools/App/Features/Tools/SwiftUI/AllTools/ToolCardsView/ToolCardView/ViewModels/ToolCardViewModel.swift
```diff
@@ -120,6 +112,15 @@ extension ToolCardViewModel {
                 }
             }
         }
+        
+        getBannerImageUseCase.getBannerImagePublisher(for: resource.attrBanner)
```
levieggertcru, line 116: Same here, see if you can use assign here.

```
getBannerImageUseCase.getBannerImagePublisher(for: resource.attrBanner)
            .assign(to: \.bannerImage, on: self)
            .store(in: &cancellables)
```

godtools/App/Share/SwiftUI Views/LessonCardView/ViewModels/LessonCardViewModel.swift
rachaelblue: Oh cool!  That cleans it up nicely.  Thanks!

levieggertcru (approved):

PR #1296, opened 2022-08-10
GT-1680 fix prayer screen signup keyboard obstruction

levieggertcru:
This PR fixes the keyboard obstruction on the Sign up for email series card found in Four Spiritual Laws > Did you pray this prayer card > tap I already made this decision > tap Sign up for an email series.

The main issue was the ToolPageCardView was expecting a direct child of MobileContentFormView and some cases the MobileContentFormView was not a direct child, but nested in a paragraph.  So that reference was never made.

form-nested


rachaelblue (approved):
This looks good.  @frett thoughts about ^?

PR #1298, opened 2022-08-11
GT-1753 remove favorited resources cache from tool cards

rachaelblue:
https://jira.cru.org/browse/GT-1753


levieggertcru (changes requested):
Hey @rachaelblue I did a quick look over of this PR and I noticed some additional work that isn't needed for [GT-1753](https://jira.cru.org/browse/GT-1753).  Mainly around adding the extra ```GetAllFavoritedToolIDsUseCase``` and the sorting of resource ids for favorited tools.  Favorited tools for now are sorted by the creation (favorited) date.  In the future there will be some additional rules around how those are sorted, such as by a user sort order.

The main focus of this task was removing the FavoritedResourcesCache dependency in place of those listed UseCases.  I should have also mentioned the ```GetAllFavoritedToolsUseCase``` in the ticket because it looks like that will be needed in the ```BaseFavoriteToolsViewModel```.  I like the ```ToggleToolFavoritedUseCase```, that will be helpful in cutting down the need to inject 2 use cases.


FavoritedResourcesCache-Dependency

rachaelblue:
Hey @levieggertcru, sorry I probably should've clarified what I did here.  

- Renamed `GetAllFavoritedToolsUseCase` --> `GetAllFavoritedToolIDsUseCase`
- Recreated `GetAllFavoritedToolsUseCase` to return `ResourceModel`, since the original version just returns `FavoritedResourceModel` (which is essentially just the tool ID, not the actual tool)

`BaseFavoriteToolsViewModel` needs `ResourceModel` objects, not just the IDs, so I thought it made sense to rename so it's clear what you're actually getting back from the use case.  So, `GetAllFavoritedToolIDsUseCase` is your existing code, and `GetAllFavoritedToolsUseCase` is new code, but it uses `GetAllFavoritedToolIDsUseCase` inside it.  Does that make sense?

levieggertcru:
Hey @rachaelblue on ```develop``` see line 182 of ```BaseFavoriteToolsViewModel``` https://github.com/CruGlobal/godtools-swift/blob/develop/godtools/App/Features/Tools/SwiftUI/Favorites/FavoriteToolsView/ViewModels/BaseFavoriteToolsViewModel.swift#L182 and it's fetching the ```FavoritedResourceModel``` objects.
And then line 32 of ```GetAllFavoritedToolsUseCase``` and it's returning ```FavoritedResourceModel```.

Eventually ```GetAllFavoritedToolsUseCase``` will return a ```ToolDomainModel```.  I just didn't want to introduce those changes yet until the refactor because it's going to be a lot more work involved.

I should have explained that better in the Jira task.

rachaelblue:
@levieggertcru ok gotcha, I'll switch it back then and leave the logic to fetch the tool itself inside the base class.

godtools/App/Features/Tools/SwiftUI/AllTools/ToolCardsView/ToolCardView/ViewModels/ToolCardViewModel.swift
```diff
@@ -96,7 +96,7 @@ extension ToolCardViewModel {
     }
     
     private func setupPublishedProperties() {
-        isFavorited = favoritedResourcesCache.isFavorited(resourceId: resource.id)
+        isFavorited = getToolIsFavoritedUseCase.getToolIsFavorited(tool: resource)
```
levieggertcru, line 99: Hey @rachaelblue this shouldn't be needed now that you are subscribed to a publisher below in the setupBinding() method.
```
getToolIsFavoritedUseCase.getToolIsFavoritedPublisher(tool: resource)
            .assign(to: \.isFavorited, on: self)
            .store(in: &cancellables)
```

levieggertcru (changes requested):

levieggertcru (approved):

PR #1308, opened 2022-08-16
GT-1763 fix realm crash removing all favorited tools

levieggertcru:
The issue was performing a ```realm.write``` from a realm notification, in our case ```getFavoritedResourcesChanged``` notification.  That's not allowed by Realm -> https://github.com/realm/realm-swift/issues/4511.  I ended up adding an extension found in the linked Github Issue to solve the crash. 

For writes I do prefer doing those on a background serial queue, especially for large writes.  That would also solve this crash issue.


rachaelblue (commented):
This fixed the crash, but I'm not sure that the behavior that caused the crash is correct.  When a user removes all favorites, we're immediately trying to add back those 4 default favorites and write that to Realm.  I would think that those 4 default favorites should only be on 1st install, but after that the user should get the "no favorite tools" message if they unfavorite them all.  Thoughts?

levieggertcru:
Hey @rachaelblue this PR addresses that issue :)

https://github.com/CruGlobal/godtools-swift/pull/1310

rachaelblue (approved):
Lol was just about to comment I realized your other PR fixes that 😄

PR #1309, opened 2022-08-16
GT-1765 Fix crash when cacheing RealmCategoryArticle with existing primaryKey

levieggertcru:


rachaelblue:
Hey @levieggertcru, were you able to reproduce this crash?

levieggertcru:
Hey @rachaelblue I actually wasn't able to reproduce this one.  It was crashing around attempting to store a RealmCategoryArticle with existing primary key.  I basically changed the code to write over the top of any existing RealmCategoryArticle objects by using the update all policy.  Also did a little optimization by moving the write outside of the for loop (write all objects at once instead of performing a write per object). 

Firebase: https://console.firebase.google.com/u/1/project/godtools-b2f82/crashlytics/app/ios:org.cru.godtools/issues/7c4fb810a969c6e5bf7984ca825d431f?time=last-seven-days&sessionEventKey=f733932ce16340278d1201d431e55c2e_1710277105940703069

rachaelblue (approved):

PR #1317, opened 2022-08-22
GT-1722 add run script for downloading resources, languages, and attachments to app bundle

levieggertcru:
This PR does the following:
- Removing resources.json, languages.json, and attachment banners from the InitialResources directory.
- Adding run script which will download the resources and languages and save those json files to the app bundle as resources.json and languages.json.  Also parses through resources and saves attachment banners to the app bundle.  This way on a fresh install user's will have this data already available.

Also note that I needed to update the ModelTypes so there wasn't any Realm dependencies necessary for the Run Script.


Scripts/DownloadInitialResources/DownloadInitialResourcesScript.swift
```diff
@@ -0,0 +1,405 @@
...
+                                
+                return deleteTemporaryAssetsDirectory()
+            })
+            .sink(receiveCompletion: { result in
+                semaphore.signal()
+            }, receiveValue: { result in
+                
+            })
+        
+        semaphore.wait()
```
rachaelblue, line 57: I don't think I've ever seen a semaphore actually in use before.  Does this just ensure that nothing else will get executed until this script finishes?

rachaelblue (commented):
Is there a good way to test this?

Scripts/DownloadInitialResources/DownloadInitialResourcesScript.swift
levieggertcru: I'm going to see if I can write some unit tests to ensure those assets are persisted to the app bundle too.

levieggertcru: Hey @rachaelblue for testing this you will want to make sure you can run that script and then run the app without a network connection to see if the resources, languages, and attachments persist.

What you can do is:
1. Ensure Build Phases > Download Initial Resources > For install builds only is unchecked.  Run the a new build with a network connection so the Download Initial Resources script runs and attaches to the device then quit the app.
2. Disable network connection and run the app and see if the lessons, all tools, favorited tools are populated with data and attachment banners are visible on those tool cards. (I need to fix something here because I noticed the lessons and all tools lists aren't loading.  Most likely due to an observable not firing off in InitialDataDownloader when the network request fails due to no network connection.)

frett: If you are looking for another example, I use a semaphore to provide external synchronization for Snowplow events on Android.

relevant code: https://github.com/CruGlobal/android-gto-support/blob/master/gto-support-snowplow/src/main/kotlin/org/ccci/gto/android/common/snowplow/events/EventSynchronizer.kt#L11

levieggertcru: Hey @rachaelblue just pushed 1 commit which addresses the issue of those observables not firing off from InitialDataDownloader in the event the network request fails.   You should be able to test now.

rachaelblue (approved):

PR #1331, opened 2022-08-24
GT-1742 get spotlight tools use case

rachaelblue:
https://jira.cru.org/browse/GT-1742

- Started refactoring to use `ToolDomainModel` instead of `ResourceModel` and made a followup subtask to finish doing this in the `ToolCardViewModel`, since that's the relevant spot for Spotlight.  (I'll also finish mapping ResourceModel in that ticket)
- Created `GetSpotlighToolsUseCase` and `GetAllToolsUseCase`, since spotlight just further filters the All Tools list


godtools/App/Share/Domain/UseCases/GetAllToolsUseCase/GetAllToolsUseCase.swift
```diff
@@ -0,0 +1,52 @@
...
+    
+    private let getToolUseCase: GetToolUseCase
+    private let resourcesRepository: ResourcesRepository
+    
+    init(getToolUseCase: GetToolUseCase, resourcesRepository: ResourcesRepository) {
+        self.getToolUseCase = getToolUseCase
+        self.resourcesRepository = resourcesRepository
+    }
+    
+    func getAllToolsPublisher(sorted: Bool) -> AnyPublisher<[ToolDomainModel], Never> {
```
levieggertcru, line 21: Hey @rachaelblue for the getAllTools won't those always be sorted by the default order?  Not sure this argument would be needed and may cause confusion.

levieggertcru (changes requested):
Hey @rachaelblue this looks like a good start to me for transitioning over to the ToolDomainModels.
Just 1 question on the getAllToolsPublisher.

godtools/App/Share/Domain/UseCases/GetAllToolsUseCase/GetAllToolsUseCase.swift
rachaelblue: The only case is when we dynamically create the category buttons-- we get allTools, but sort them by language availability first and default order second, and then pull out the categories.  I could make this a default parameter like `func getAllToolsPublisher(sorted: Bool = true)` so we otherwise don't have to think about it?

levieggertcru: Hey @rachaelblue I forgot about sorting on the language availability. Lets leave this as is.

levieggertcru (approved):

PR #1334, opened 2022-08-26
GT-1742 finish mapping resource model

rachaelblue:
https://jira.cru.org/browse/GT-1777

- Cleaned up the remaining `ResourceModel` dependencies inside `ToolCardViewModel`
- In `GetLanguageAvailabilityStringUseCase`, `ResourceModel` was being passed in, and in some cases it was a tool and others it was a lesson.  This brings us back to the question of whether or not we want lessons to share tool code or have duplicated code.  I tend think it makes sense to share behavior by using protocols, but still keep the domain models separate-- `ToolDomainModel` and `LessonDomainModel`.  So, I made a protocol `LanguageSupportable` so any domain model that conforms to it can use `GetLanguageAvailabilityStringUseCase` moving forward.  Thoughts?


levieggertcru:
Hey @rachaelblue yeah I see where that is needed to share that UseCase logic between domain models.

levieggertcru (approved):

PR #1335, opened 2022-08-29
GT-1358 route multiplatform parsing errors to crashlytics

levieggertcru:
This PR implements GodToolsToolParser's NapierProxyKt logging for logging errors to Firebase.


godtools/App/Share/Data/GodToolsParserLogger/GodToolsParserLogger.swift
```diff
@@ -0,0 +1,39 @@
...
+        isStarted = true
+                
+        NapierProxyKt.enableCustomLogging { (logLevel: NapierLogLevel, tag: String?, throwable: KotlinThrowable?, message: String?) in
+            
+            guard let error = throwable?.asError() else {
+                return
+            }
+            
+            firebaseCrashReporting.recordError(error: error)
+        }
```
frett, line 37: this is logging just any captured exceptions pushed through the log system. This is not currently including any log messages that may be included with the errors.

you will want to use `firebaseCrashReporting.log()` to record any log messages that are included as well.

see: https://github.com/AAkira/Napier/blob/master/ios/Napier/AppDelegate.swift#L27

frett: instead of separating the logging callback into 2 separate callbacks like they did in their sample, I left it as a single callback that you can log messages and record errors from

levieggertcru: Thanks Daniel!  Going to update.

godtools/App/Share/Data/GodToolsParserLogger/GodToolsParserLogger.swift
```diff
@@ -0,0 +1,44 @@
...
+        NapierProxyKt.enableCustomLogging { (logLevel: NapierLogLevel, tag: String?, throwable: KotlinThrowable?, message: String?) in
+            
+            DispatchQueue.global().async {
+                
+                if let error = throwable?.asError() {
+                    Crashlytics.crashlytics().record(error: error)
+                }
+                else {
+                    Crashlytics.crashlytics().log("\(String(describing: tag)): \(String(describing: message))")
+                }
```
frett, line 40:
```suggestion
                Crashlytics.crashlytics().log("\(String(describing: tag)): \(String(describing: message))")
                if let error = throwable?.asError() {
                    Crashlytics.crashlytics().record(error: error)
                }
```

levieggertcru: I see should log regardless if error or not.

frett (approved):

PR #1342, opened 2022-09-01
GT 1785 updates for shared parser

levieggertcru:
This PR makes some changes for the GodToolsToolParser.
- Replaced ```.tips.isEmpty``` with ```.hasTips``` on ```Manifest```.
- Using ```.withAppVersion(deviceType: .ios, version: appConfig.appVersion)``` on ```ParseTranslationManifestForRenderer.swift```.
- Removed ```parsesRelatedFiles``` on ```TranslationManifestParserType.swift``` since that caused confusion between downloading vs parsing related files and instead exposed ```includeRelatedFiles``` on ```TranslationsRepository``` which controls if related files should be downloaded from remote.


godtools/App/Share/Data/TranslationsRepository/ManifestParser/ParseTranslationManifestForRenderer.swift
```diff
@@ -19,10 +19,16 @@ class ParseTranslationManifestForRenderer: TranslationManifestParser {
         ParserConfigKt.FEATURE_MULTISELECT
     ]
         
-    init(resourcesFileCache: ResourcesSHA256FileCache) {
-    
+    init(appConfig: AppConfig, resourcesFileCache: ResourcesSHA256FileCache) {
+            
+        let parserConfig = ParserConfig()
+            .withParsePages(enabled: true)
+            .withParseTips(enabled: true)
```
frett, line 26: these default to true, so you technically don't need to specify them unless you want to be verbose.

godtools/App/Share/Data/TranslationsRepository/ManifestParser/TranslationManifestParser.swift
```diff
@@ -15,15 +15,20 @@ class TranslationManifestParser {
...
     
-    static func getManifestParser(type: TranslationManifestParserType, resourcesFileCache: ResourcesSHA256FileCache) -> TranslationManifestParser {
+    static func getManifestParser(type: TranslationManifestParserType, appConfig: AppConfig, resourcesFileCache: ResourcesSHA256FileCache) -> TranslationManifestParser {
         
         switch type {
-        case .related:
-            return ParseTranslationManifestForRelatedFiles(resourcesFileCache: resourcesFileCache)
+        
+        case .manifestAndRelatedFiles:
+            let parserConfig = ParserConfig().withParseRelated(enabled: false)
```
frett, line 23: I think you meant `enabled: true` here?

otherwise this is the same config as `.manifestOnly`

frett: hmm, maybe you did mean `enabled: false`. what are the use-cases that drive `.manifestAndRelatedFiles` and `.manifestOnly`? should they just be a single type?

levieggertcru: Hey @frett so parse related true here does that just attach the related files to the manifest?

I will also update this to set parsePages and parseTips to false since that parsing isn't needed when downloading and cacheing the manifest and related files to the file cache.

godtools/App/Share/Data/TranslationsRepository/ManifestParser/ParseTranslationManifestForRenderer.swift
levieggertcru: That's good to know, I think I will keep them just to be verbose.

godtools/App/Share/Data/TranslationsRepository/ManifestParser/TranslationManifestParser.swift
frett: taking a look at where you are using the 2 types, I think they can just be a single type of `.manifestOnly`.

frett: no matter the parser mode the `manifest.relatedFiles` property will always return a list of all related files, there is no parser mode to disable that behavior.

frett: the flags just indicate if the parser follows that list of related files to actually parse those files or not.

PR #1348, opened 2022-09-06
GT-1743 get all tools use case

rachaelblue:
https://jira.cru.org/browse/GT-1743

On the category buttons, I added some logic to use `GetSettingsPrimaryLanguageUseCase` to get the bundle for the category translation because the old logic wasn't updating properly when the language changed.  This all needs to become its own use case (maybe with the tool title translation logic, too?), but just wanted to get it working for now.


godtools/App/Features/Tools/SwiftUI/AllTools/ToolCategoriesView/ToolCategoryButtonView/ViewModels/ToolCategoryButtonViewModel.swift
```diff
@@ -12,29 +12,38 @@ class ToolCategoryButtonViewModel: BaseToolCategoryButtonViewModel {
...
+    init(category: ToolCategoryDomainModel, selectedCategoryName: String?, localizationServices: LocalizationServices, getSettingsPrimaryLanguageUseCase: GetSettingsPrimaryLanguageUseCase) {
+        self.category = category
         self.localizationServices = localizationServices
-        self.languageSettingsService = languageSettingsService
+        self.getSettingsPrimaryLanguageUseCase = getSettingsPrimaryLanguageUseCase
         
-        let bundle = localizationServices.bundleLoader.bundleForPrimaryLanguageOrFallback(in: languageSettingsService)
-        let translatedCategory = localizationServices.toolCategoryStringForBundle(bundle: bundle, attrCategory: attrCategory)
+        let bundle: Bundle
+        if let primaryLanguage = getSettingsPrimaryLanguageUseCase.getPrimaryLanguage() {
```
levieggertcru, line 27: Hey @rachaelblue I would like this logic to go in the ```GetToolCategoriesUseCase```.  The ```ToolCategoryDomainModel``` title would then map to the button title.

godtools/App/Features/Tools/SwiftUI/AllTools/ToolCategoriesView/ToolCategoryButtonView/ViewModels/AllToolsCategoryButtonViewModel.swift
```diff
@@ -9,10 +9,19 @@
 import Foundation
 
 class AllToolsCategoryButtonViewModel: BaseToolCategoryButtonViewModel {
```
levieggertcru, line 11: Hey @rachaelblue I would also like the All Tools category to be included in the ```GetToolCategoriesUseCase```.

godtools/App/Share/Domain/UseCases/GetAllToolsUseCase/GetAllToolsUseCase.swift
```diff
@@ -17,6 +17,24 @@ class GetAllToolsUseCase {
...
+                
+                let tools = self.getAllTools(sorted: true, with: category)
+                
+                return Just(tools)
+                    .eraseToAnyPublisher()
+            }
+            .eraseToAnyPublisher()
+    }
     
     func getAllToolsPublisher(sorted: Bool) -> AnyPublisher<[ToolDomainModel], Never> {
```
levieggertcru, line 39: Hey @rachaelblue I think this method should be private since use cases should expose a single point of execution.  The GetAllToolsUseCase we will always want to filter on the category since that is the intended functionality.  And I like the idea of passing the category CurrentValueSubject and using CombineLatest here.

levieggertcru (changes requested):

godtools/App/Share/Domain/UseCases/GetAllToolsUseCase/GetAllToolsUseCase.swift
rachaelblue: That method wasn't being used anymore so I deleted it.  But, your comment about having only one method exposed per use case brings me to a different question -- `GetAllToolsUseCase` has an additional public method that returns an array of `ResourceModel`s because `GetToolCategoriesUseCase` needs the all tools result before it gets mapped to `ToolDomainModel`.  Thoughts about that?  This may be one potential reason for putting certain filtering logic on the repository itself, when multiple use cases need to use it.

rachaelblue: Also, in order to move the logic that adds the "All Tools" category into the use case, I made a dummy id to keep track of that category- `static let allToolsCategoryId = "allTools"`.  I have this in the `GetToolCategoriesUseCase`, but I'm not sure where else it might fit better.  Does that make sense?

levieggertcru (changes requested):
Just a few more dependencies that can be removed / changed.

levieggertcru (approved):

godtools/App/Features/Tools/SwiftUI/AllTools/ToolCategoriesView/ToolCategoriesViewModel.swift
```diff
@@ -6,80 +6,58 @@
...
     @Published var categoryTitleText: String = ""
-    @Published var buttonViewModels = [BaseToolCategoryButtonViewModel]()
-    @Published var selectedCategory: String?
+    @Published var buttonViewModels = [ToolCategoryButtonViewModel]()
+    @Published var selectedCategoryId: String? = nil
     
     // MARK: - Init
     
-    init(dataDownloader: InitialDataDownloader, languageSettingsService: LanguageSettingsService, localizationServices: LocalizationServices, delegate: ToolCategoriesViewModelDelegate?) {
+    init(dataDownloader: InitialDataDownloader, localizationServices: LocalizationServices, getSettingsPrimaryLanguageUseCase: GetSettingsPrimaryLanguageUseCase, getToolCategoriesUseCase: GetToolCategoriesUseCase, delegate: ToolCategoriesViewModelDelegate?) {
```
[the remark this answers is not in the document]
rachaelblue: Fixed.  Will be more careful about checking for unused references moving forward!

PR #1350, opened 2022-09-08
GT-1413 delete resources after sync

levieggertcru:


levieggertcru:
This PR does the following:
- Deletes old Realm translation objects after syncing resources plus latest translations and latest attachments.
- Added check to renderer where if any translations have been updated, due to a sync completing, then the current renderer will download the latest translations, rebuild the MobileContentRenderer object, and set the current renderer to the new MobileContentRenderer object causing the renderer to update its UI.

rachaelblue (approved):

PR #1352, opened 2022-09-09
GT-1787 language settings feature

levieggertcru:
This PR creates a Feature folder for Language Settings and related presentation and domain use cases.  For now the use cases are consuming the shared ```LanguageDomainModel``` in ```Share/Domain/UseCases/GetLanguageUseCase```.

feature-language-settings


rachaelblue (approved):
I like this structure, seems like it'll make it easier to navigate 👍

PR #1354, opened 2022-09-12
GT-1745 get all favorited tools use case

rachaelblue:
https://jira.cru.org/browse/GT-1745

- I renamed the existing `GetAllFavoritedToolsUseCase` to `GetAllFavoritedResourceModelsUseCase`, and reworked `GetAllFavoritedToolsUseCase` so that it uses `GetAllFavoritedResourceModelsUseCase` and returns `ToolDomainModel`s.
- Favorites listens to `cachedResourcesAvailable` to determine when to hide the loading spinner.  Will this flag get moved over to the ResourcesRepository?


levieggertcru:
Hey @rachaelblue for the ```cachedResourcesAvailable```, that is now replaced by ```ResourcesRepository getResourcesChanged()``` which is triggered anytime resources are stored in the RealmDatabase.  
Favorited resources will be available on first install so I don't think you will need to worry about a spinner, but for added measure you could Combine the```ResourcesRepository getResourcesChanged()``` with the ```GetAllFavoritedToolsUseCase getFavoritedResourcesChanged()```

levieggertcru:
Hey @rachaelblue also instead of having the ```GetAllFavoritedResourceModelsUseCase```,  we should remove the ```GetAllFavoritedToolsUseCase``` from ```GetShortcutItemsUseCase``` and instead inject ```favoritedResourcesRepository: FavoritedResourcesRepository``` in there.

```GetAllFavoritedToolsUseCase``` should really only tie to the ```Home``` screen since we're moving away from the Quasi clean architecture. 

Another issue with ```GetAllFavoritedResourceModelsUseCase``` is it's returning data layer objects which are supposed to be hidden from the presentation layer.

**Comment by rachaelblue** [comment]
@levieggertcru ok changes based on our earlier conversation:
- removed the loading spinner from the favorites tab (not pull to refresh) since we should always have some sort of content there
- moved the logic for determining what language to show for the tool name into ToolDomainModel and encapsulated that result as a `CurrentToolTranslationDomainModel`.  There are other strings in ToolCardViewModel that need to know what language that specific tool is showing in (separate from what the primary language actually is) so they read from `CurrentToolTranslationDomainModel`.
- created a new ticket for removing `GetAllFavoritedResourceModelsUseCase` later: https://jira.cru.org/browse/GT-1789?atlLinkOrigin=c2xhY2staW50ZWdyYXRpb258aXNzdWU%3D

File: godtools/App/Features/Tools/SwiftUI/AllTools/ToolCardsView/ToolCardView/ViewModels/ToolCardViewModel.swift [hunk 1]
```diff
@@ -131,46 +126,31 @@ extension ToolCardViewModel {
...
-        if let primaryLanguage = languageSettingsService.primaryLanguage.value, let primaryTranslation = resourcesCache.getResourceLanguageTranslation(resourceId: tool.dataModelId, languageId: primaryLanguage.id) {
+        switch tool.currentTranslation {
+        case .primaryLanguage(let language, _):
             
-            toolName = primaryTranslation.translatedName
-            languageBundle = bundleLoader.bundleForResource(resourceName: primaryLanguage.code) ?? Bundle.main
-            languageDirection = primaryLanguage.languageDirection
-        }
-        else if let englishTranslation = resourcesCache.getResourceLanguageTranslation(resourceId: tool.id, languageCode: "en") {
+            languageBundle = bundleLoader.bundleForResource(resourceName: language.localeIdentifier) ?? Bundle.main
```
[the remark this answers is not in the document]
rachaelblue: Good idea, that cleaned it up a little more.

godtools/App/Share/Domain/UseCases/GetToolUseCase/ToolDomainModel.swift
```diff
@@ -10,9 +10,10 @@ import Foundation
 
 struct ToolDomainModel {
     
-    // TODO: - finish mapping necessary attributes in GT-1777
+    let abbreviation: String
     let bannerImageId: String
     let category: String
+    let currentTranslationLanguage: LanguageDomainModel?
```
levieggertcru, line 16: Hey @rachaelblue I don't think this needs to be optional now does it?

levieggertcru (changes requested):

godtools/App/Share/Domain/UseCases/GetToolUseCase/ToolDomainModel.swift
rachaelblue: nope, you're right

PR #1364, opened 2022-09-16
GT-1748 get lessons use case

rachaelblue:
https://jira.cru.org/browse/GT-1748

This is a lot of files mostly because I refactored areas that had `ResourceModel` to use `LessonDomainModel` instead, and passing new use cases down the chain.  This only addresses the Lessons page itself, not the featured lessons yet.


godtools/App/Share/Domain/UseCases/GetLanguageAvailabilityUseCase/LanguageAvailabilityDomainModel.swift
```diff
@@ -0,0 +1,15 @@
...
+//  LanguageAvailabilityDomainModel.swift
+//  godtools
+//
+//  Created by Rachael Skeath on 9/15/22.
+//  Copyright © 2022 Cru. All rights reserved.
+//
+
+import Foundation
+
+struct LanguageAvailabilityDomainModel {
```
levieggertcru, line 11: I like this much better than using the tuple.  Helps to see a concrete type.

levieggertcru (approved):

PR #1370, opened 2022-09-16
GT upgrade Github Actions to Xcode_14.0

levieggertcru:


rachaelblue (approved):

.github/workflows/build.yml
```diff
@@ -42,6 +42,13 @@ jobs:
           key: ${{ runner.os }}-konan-${{ github.sha }}
           restore-keys: ${{ runner.os }}-konan-
 
+      - name: Archive ktlint results
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: ktlint-results
+          path: "**/build/reports/ktlint/"
```
frett, line 50: So, you will want to modify this to actually archive test results/output files that would be beneficial for diagnosing failures after the fact

frett: Ktlint is a printer for Kotlin code, that utility isn't relevant for the iOS repo

frett (commented):
The archive step should be updated to archive files you care about and not files that are only relevant in Kotlin projects

frett (approved):

PR #1371, opened 2022-09-16
GT Clean up ColorPalette

levieggertcru:
Use rawValue as the colorName from AssetCatalog and UIColor and Color both instantiate from the colorName (rawValue)


rachaelblue (approved):
much cleaner 👍

PR #1378, opened 2022-09-19
GT Increase minimum deployment target to iOS 14

levieggertcru:
With the release of iOS 16 it's a good time to increase the minimum deployment target to iOS 14.

Most of our users are running iOS 15 and a very small percentage of our users are running iOS 13.

iOS 13-16 last 30 days


rachaelblue (approved):
🎉

PR #1380, opened 2022-09-20
GT-1795 Fix Lottie frame sizing in ToolDetails 

levieggertcru:
Fixed this by following this example here (https://medium.com/swlh/how-to-use-lottie-animations-in-swiftui-caaf19944d96)

The nice thing is now the frame size doesn't need to be injected into the UIViewRepresentable and can be configured from SwiftUI.


rachaelblue (approved):
Nice!  Glad you found that tutorial!

PR #1386, opened 2022-09-22
GT-1804 restore button to delete parallel language

levieggertcru:
When choosing a parallel language from language settings, there was a Clear button in the top right of the navigation bar that allowed users to delete the parallel language from settings.  That logic got lost during the refactor.  This PR fixes that and shows the button to delete the settings parallel language if a parallel language exists.


rachaelblue (approved):
When you clear the parallel language, the tool card doesn't remove the language name from the subtitle-- pretty sure this is the same issue as #1388, so I'll fix that in there.

levieggertcru:
Thanks Rachael!

PR #1395, opened 2022-09-23
GT-1768 Jumpy Home Screen

rachaelblue:
https://jira.cru.org/browse/GT-1768

I couldn't reproduce the specific thing Ann was seeing in the ticket (seems likely that it got resolved since she made that ticket).  But, the launch experience in general felt a little glitchy, so I did add the loading spinner back for the initial load only.  After that, you shouldn't see that spinner again.  Also, iOS 15 animates everything in more smoothly than iOS 16 for some reason-- there's a small jump on first launch when the favorites row appears, but it seems to be iOS 16 only.


godtools/App/Features/Lessons/Lessons/LessonsView/LessonsViewModel.swift
```diff
@@ -81,8 +82,10 @@ extension LessonsViewModel {
             .receiveOnMain()
             .sink { [weak self] lessons in
                 
-                self?.isLoading = lessons.isEmpty
-                self?.lessons = lessons
+                withAnimation {
```
levieggertcru, line 85: Hey @rachaelblue do you know if the withAnimation block will work keeping it in the View?  It doesn't seem correct to have that in the ViewModel since animation logic is a responsibility of the View.

godtools/App/Features/Tools/SwiftUI/Favorites/FavoriteToolsView/FavoriteToolsView.swift
```diff
@@ -23,7 +23,11 @@ struct FavoriteToolsView: View {
             
             FavoriteToolsHeaderView(viewModel: viewModel, leadingPadding: leadingPadding)
             
-            if viewModel.tools.isEmpty {
+            if viewModel.isInitialLoad {
```
levieggertcru, line 26: I think an enum would be good here too for this view state, ```loadingTools```, ```noTools```, ```tools```.

levieggertcru:
I looked into this a little and I'm not sure why there is that initial delay.  The resources in Realm should be immediately readable, but for some reason there is that split delay.  Let's keep the loading spinner and I might come back to this later on because I'm curious as to why there is that slight delay.

godtools/App/Features/Lessons/Lessons/LessonsView/LessonsViewModel.swift
rachaelblue: I can try adding implicit animations to the view instead, I believe that should have the same effect.

levieggertcru (approved):

PR #1401, opened 2022-09-27
GT-1141 Update training tip ask icon to fix clipping at bottom of icon

levieggertcru:
training_tips_colored_speech_bubble_clipping

new-speech-bubble-no-clipping


rachaelblue (approved):
Not sure how you noticed that, I had to zoom in my already-magnified screen to see it 😄 Although, it bugs me a little that the speech bubble isn't centered in the pink square, not sure if that's intentional or not.

levieggertcru:
Hey @rachaelblue I didn't notice the centering, but that is off and I think it would be nice to have that centered.  I'll look into centering that in this PR.  Those are two separate images so we should be able to center it.

levieggertcru:
Hey @rachaelblue going to merge this PR and create a new ticket for centering the icon.

PR #1416, opened 2022-10-17
GT-1580 Add arguments for contentLanguage and secondaryContentLanguage in FirebaseAnalytics

levieggertcru:
This PR removes the ```LanguageSettingsService.swift``` dependency from ```FirebaseAnalytics.swift``` in order that contentLanguage and secondaryContentLanguage can be injected into FirebaseAnalytics.   
Most cases the contentLanguage and secondaryContentLanguage will default to the settings primary and parallel language with some exceptions (example tool renderer).

### DIFF
--- a/godtools/App/AppDelegate.swift
+++ b/godtools/App/AppDelegate.swift
@@ -160,9 +160,23 @@ extension AppDelegate {
             
         case .tool:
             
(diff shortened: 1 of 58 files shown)


levieggertcru:
Not sure why the build status hasn't updated in the PR, but I did manually trigger github actions and it passed. https://github.com/CruGlobal/godtools-swift/actions/runs/3268355662

PR #1426, opened 2022-10-28
GT-1834 custom tab bar view

rachaelblue:


godtools/App/Features/Dashboard/Presentation/Dashboard/DashboardView.swift
```diff
@@ -17,41 +17,67 @@ struct DashboardView: View {
     init(viewModel: DashboardViewModel) {
         self.viewModel = viewModel
         
-        UITabBar.appearance().barTintColor = .white
-        UITabBar.appearance().unselectedItemTintColor = UIColor(red: 170 / 255, green: 170 / 255, blue: 170 / 255, alpha: 1)
-        UITabBarItem.appearance().setTitleTextAttributes([.font: FontLibrary.sfProTextRegular.uiFont(size: 12) as Any], for: .normal)
+        if #available(iOS 16.0, *) { } else {
+            UITabBar.appearance().isHidden = true
```
rachaelblue, line 21: @levieggertcru unfortunately I'm still having to use `appearance` to hide the default tab bar for anything less than iOS 16, at which point a modifier exists to hide it.  Although, technically the custom tab bar sits on top of the `TabView`, so if we don't hide the default tab bar programmatically it's still not actually visible.  But leaving it in there seemed kinda hacky 😄

levieggertcru: Hey @rachaelblue I think it's a good idea to hide the default tab bar too.  I was out yesterday, going to review this PR now.

levieggertcru (approved):

PR #1427, opened 2022-11-07
GT-1818 auth for mobile content api

rachaelblue:
Hey @levieggertcru, this is a first stab at the auth flow.  You can enable login in the menu, and upon logging in it will fetch the mobile content auth token and store it in the keychain.

### DIFF
--- a/godtools/App/Features/Menu/Domain/UseCases/AuthenticateUserUseCase/AuthenticateUserUseCase.swift
+++ b/godtools/App/Features/Menu/Domain/UseCases/AuthenticateUserUseCase/AuthenticateUserUseCase.swift
@@ -15,30 +15,20 @@ class AuthenticateUserUseCase {
     private let cruOktaAuthentication: CruOktaAuthentication
     private let emailSignUpService: EmailSignUpService
     private let firebaseAnalytics: FirebaseAnalytics
-    private let mobileContentAuthTokenRepository: MobileContentAuthTokenRepository
     private let snowplowAnalytics: SnowplowAnalytics
     
-    init(cruOktaAuthentication: CruOktaAuthentication, emailSignUpService: EmailSignUpService, firebaseAnalytics: FirebaseAnalytics, mobileContentAuthTokenRepository: MobileContentAuthTokenRepository, snowplowAnalytics: SnowplowAnalytics) {
+    init(cruOktaAuthentication: CruOktaAuthentication, emailSignUpService: EmailSignUpService, firebaseAnalytics: FirebaseAnalytics, snowplowAnalytics: SnowplowAnalytics) {
         
         self.cruOktaAuthentication = cruOktaAuthentication
         self.emailSignUpService = emailSignUpService
         self.firebaseAnalytics = firebaseAnalytics
-        self.mobileContentAuthTokenRepository = mobileContentAuthTokenRepository
         self.snowplowAnalytics = snowplowAnalytics
     }
     
     func authenticatePublisher(authType: AuthenticateUserAuthTypeDomainModel) -> AnyPublisher<Bool, Error> {
         
         return authenticateByAuthTypePublisher(authType: authType)
-            .flatMap({ (accessToken: OktaAccessToken) -> AnyPublisher<String?, Error> in
-                
-                return self.mobileContentAuthTokenRepository.fetchRemoteAuthToken(oktaAccessToken: accessToken.value)
-                    .mapError { urlResponseError in
-                        return urlResponseError.getError()
-                    }
-                    .eraseToAnyPublisher()
-            })
-            .flatMap({ _ -> AnyPublisher<CruOktaUserDataModel, Error> in
+            .flatMap({ (accessToken: OktaAccessToken) -> AnyPublisher<CruOktaUserDataModel, Error> in
                                 
                 return self.cruOktaAuthentication.getAuthUserPublisher()
                     .mapError { oktaError in
--- a/godtools/App/DependencyContainer/AppDomainLayerDependencies.swift
+++ b/godtools/App/DependencyContainer/AppDomainLayerDependencies.swift
@@ -57,7 +57,6 @@ class AppDomainLayerDependencies {
             cruOktaAuthentication: dataLayer.getCruOktaAuthentication(),
             emailSignUpService: dataLayer.getEmailSignUpService(),
             firebaseAnalytics: dataLayer.getAnalytics().firebaseAnalytics,
-            mobileContentAuthTokenRepository: dataLayer.getMobileContentAuthTokenRepository(),
             snowplowAnalytics: dataLayer.getAnalytics().snowplowAnalytics
         )
     }


godtools/App/Share/Domain/UseCases/GetMobileContentAuthTokenUseCase/GetMobileContentAuthTokenUseCase.swift
```diff
@@ -0,0 +1,62 @@
...
+//
+//  Created by Rachael Skeath on 11/7/22.
+//  Copyright © 2022 Cru. All rights reserved.
+//
+
+import Foundation
+import OktaAuthentication
+import Combine
+
+class GetMobileContentAuthTokenUseCase {
```
levieggertcru, line 13: Here we don't need this UseCase since there isn't any business rules being applied.  The ```AuthenticateUserUseCase``` should communicate directly with the ```AuthTokenRepository``` and there should always be a new auth token fetched when a user authenticates.

godtools/App/Share/Data/MobileContentAuthTokenRepository/MobileContentAuthTokenRepository.swift
```diff
@@ -0,0 +1,61 @@
...
+import Foundation
+import OktaAuthentication
+import Combine
+
+class MobileContentAuthTokenRepository {
+    
+    private let api: MobileContentAuthTokenAPI
+    private let cache: MobileContentAuthTokenCache
+    
+    private var cancellables = Set<AnyCancellable>()
```
levieggertcru, line 18: Hey @rachaelblue looks like cancellables snuck in here and isn't needed.

godtools/App/Share/Data/MobileContentAuthTokenRepository/MobileContentAuthTokenRepository.swift
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    private func getCachedAuthToken(for userId: Int?) -> String? {
+        
+        guard let userId = userId else { return nil }
+        
+        return cache.getAuthToken(for: userId)
+    }
+    
+    private func fetchAuthToken(oktaAccessToken: String) -> AnyPublisher<String?, URLResponseError> {
```
levieggertcru, line 47: Hey @rachaelblue I think this method would be nice to have exposed.  The ```AuthenticateUserUseCase``` could then make use of this and fetch a new auth token when a user authenticates.
It would also be nice to include in the method name that this is fetching a token from Remote too.

godtools/App/Share/Data/MobileContentAuthTokenRepository/MobileContentAuthTokenRepository.swift
```diff
@@ -0,0 +1,61 @@
...
+    
+    private var cancellables = Set<AnyCancellable>()
+    
+    init(api: MobileContentAuthTokenAPI, cache: MobileContentAuthTokenCache) {
+        
+        self.api = api
+        self.cache = cache
+    }
+    
+    func getAuthTokenPublisher(for userId: Int? = nil, oktaAccessToken: String) -> AnyPublisher<String?, URLResponseError> {
```
levieggertcru, line 26: I think here it would be nice if this could fallback to fetching from Remote if fetching from cache doesn't provide a value.

godtools/App/Features/Menu/Domain/UseCases/AuthenticateUserUseCase/AuthenticateUserUseCase.swift
```diff
@@ -30,6 +32,8 @@ class AuthenticateUserUseCase {
         return authenticateByAuthTypePublisher(authType: authType)
             .flatMap({ (accessToken: OktaAccessToken) -> AnyPublisher<CruOktaUserDataModel, Error> in
                 
+                self.getMobileContentAuthToken.refreshAuthToken(with: accessToken.value)
```
levieggertcru, line 35: Hey @rachaelblue also note here the fetching of the authToken we want part of the ```authenticateByAuthTypePublisher``` chain ```(add additional flatMap)```.  This way if ```authenticateByAuthTypePublisher``` is cancelled, everything is cancelled in this chain.

levieggertcru (changes requested):
Hey @rachaelblue looking good so far.  I left some comments and let me know if you want to catch up on any comments I have made.

godtools/App/Share/Data/MobileContentAuthTokenRepository/Api/Models/MobileContentAuthTokenDataModel.swift
```diff
@@ -0,0 +1,42 @@
...
+    
+    init(from decoder: Decoder) throws {
+        
+        let container = try decoder.container(keyedBy: RootKeys.self)
+        
+        let dataContainer = try container.nestedContainer(keyedBy: DataKeys.self, forKey: .data)
+        let attributesContainer = try dataContainer.nestedContainer(keyedBy: AttributesKeys.self, forKey: .attributes)
+        
+        token = try attributesContainer.decode(String.self, forKey: .token)
+        expirationDate = try attributesContainer.decodeIfPresent(String.self, forKey: .expiration) ?? ""
```
levieggertcru, line 39: One more thing, I'm not sure if we will use the expirationDate, but it would be nice to have this decoded to a Date that way we have a Date type to work with if we need to do any sort of comparison or date checking.  ```ISO8601DateFormatter().date(from: <String>)```

godtools/App/Services/KeychainService/KeychainService.swift
```diff
@@ -0,0 +1,131 @@
...
+//  godtools
+//
+//  Created by Rachael Skeath on 11/2/22.
+//  Copyright © 2022 Cru. All rights reserved.
+//
+
+import Foundation
+import Security
+
+class KeychainService {
```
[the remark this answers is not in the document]
levieggertcru: Hey @rachaelblue service isn't necessary in the name.  The ```Service``` naming ended up coming from a tutorial I was reading on ```MVVM``` where they encapsulated data layer (networking, cache) in ```Services```.  Same as our ```Repository``` classes.  

service

As far as class naming goes, I'm trying to get better at implementing SOLID and naming classes to be more specific to the responsibility of the class.  Here is a good place to name more specifically to the responsibility of the class  and ```MobileContentAuthTokenKeychainAccessor``` is a great name for this class.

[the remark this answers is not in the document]
rachaelblue: Ok cool, that makes sense!  Thanks for explaining!

godtools/App/Features/Menu/Domain/UseCases/AuthenticateUserUseCase/AuthenticateUserUseCase.swift
levieggertcru: Hey @rachaelblue hmm that's true, we don't need the auth token value here and at this point we really don't need to fetch the auth token at all.

We could just leave it out and allow the auth token to be fetched when an auth request is made.  The benefit of having it here is we would have a auth token in place for that future initial request.  Were you thinking along these lines too?

If we do include the fetch request in the chain here, we would still get the return value, we just wouldn't do anything with it which might be confusing.  It would look something like this...

```swift
return authenticateByAuthTypePublisher(authType: authType)
            .flatMap({ (accessToken: OktaAccessToken) -> AnyPublisher<String?, URLResponseError> in
                
                return self.getMobileContentAuthToken.fetchTokenFromRemote(oktaAccessToken: accessToken.value)

                    .eraseToAnyPublisher()
            })
            .flatMap({ (authToken: String?) -> AnyPublisher<CruOktaUserDataModel, Error> in
                                                                
                return self.cruOktaAuthentication.getAuthUserPublisher()
                    .mapError { oktaError in
                        return oktaError.getError()
                    }
                    .eraseToAnyPublisher()
            })
            .flatMap({ (authUser: CruOktaUserDataModel) -> AnyPublisher<Bool, Error> in
                
                self.postEmailSignUp(authUser: authUser)
                self.setAnalyticsUserProperties(authUser: authUser)
                
                return Just(true).setFailureType(to: Error.self)
                    .eraseToAnyPublisher()
            })
            .eraseToAnyPublisher()
```

rachaelblue: Ah I see, I was thinking the Okta token was used elsewhere in the chain, so commandeering it for `fetchTokenFromRemote()` wasn't going to work.  But since that's not the case, I think it's ok to ignore the auth token value as long as it makes sense to wait until the auth token responds to trigger `getAuthUserPublisher()`.

We probably still want to fetch the token here rather than waiting til we need it.  I think Android is doing it this way too, attempts to use an existing token and only refreshes it if the auth request fails.

So if that all makes sense, I think we can go with your solution ^

wrandall22: cc @frett

godtools/App/Share/Data/MobileContentAuthTokenRepository/MobileContentAuthTokenRepository.swift
rachaelblue: It's doing that already, but should be more obvious since renaming to `fetchRemoteAuthToken()`

godtools/App/Features/Menu/Domain/UseCases/AuthenticateUserUseCase/AuthenticateUserUseCase.swift
levieggertcru: Hey @rachaelblue I'd say let's fetch that auth token here that way it's stored and ready for our authorized requests.

It would be nice if we could execute the ```fetchTokenFromRemote()``` and ```getAuthUserPublisher``` concurrently as well.  Maybe we can use CombineLatest or MergeMany.  I need to look those up again and refresh myself on the differences.

frett: on Android we don't pro-actively fetch the auth token, we fetch it on the first call to an API that requires authentication. after we fetch it for the authenticated API request we cache it to reuse as long as we can so that we aren't constantly requesting it every time we use an authenticated API.

I also have synchronization in place, so if multiple authenticated API requests fire at the same time they will allow block until a single auth token is retrieved, and then all of them resume. This is a nice optimization, but isn't absolutely necessary.

frett: Also of note is the fact that we only use authentication on user APIs currently. There is no need to change how tools are synced

levieggertcru: Thanks @frett.  @rachaelblue after Daniel's comment lets keep it consistent with Android and only fetch the auth token when it's actually needed via an authorized request.

rachaelblue: Sounds good, fixed!

godtools/App/Share/Data/MobileContentAuthTokenRepository/MobileContentAuthTokenRepository.swift
levieggertcru: Oh yeah it is !

levieggertcru (approved):

PR #1431, opened 2022-11-14
GT-1859 add realm configuration with option for memory or disk

levieggertcru:
This PR makes some updates to ```RealmDatabase.swift```
- Moved RealmDatabase.swift files to Share/Data/RealmDatabase/
- Add files for configuring the RealmDatabase with either a memory or disk cache.  I think I will need the memory cache for writing tests.  Will possibly move SwiftUIPreviewDiContainer to use a memory cache.
- Schema version is now configured from ```RealmDatabaseProductionConfiguration.swift```.
- Also moved SwiftUIPreviewDiContainer.swift to Preview/DependencyContainer/


rachaelblue (approved):
This should make writing tests much easier!

godtools/App/Share/Data/RealmDatabase/RealmDatabaseConfiguration+RealmConfig.swift
```diff
@@ -0,0 +1,35 @@
...
+        var realmConfig: Realm.Configuration
+        
+        switch cacheType {
+            
+        case .disk(let fileName, let migrationBlock):
+            
+            realmConfig = Realm.Configuration()
+            realmConfig.fileURL = realmConfig.fileURL?.deletingLastPathComponent().appendingPathComponent(fileName)
+            realmConfig.schemaVersion = schemaVersion
+            realmConfig.migrationBlock = migrationBlock
```
levieggertcru, line 25: Hey @rachaelblue I had to make one more change in this PR which was removing the MigrationBlock closure from the extension here and instead have that be injected with the RealmDatabaseCacheType disk.

The migrationBlock is now created in RealmDatabaseProductionConfiguration.
```
let migrationBlock = { (migration: Migration, oldSchemaVersion: UInt64) in

            if (oldSchemaVersion < 1) {
                // Nothing to do!
                // Realm will automatically detect new properties and removed properties
                // And will update the schema on disk automatically
            }
        }
```

rachaelblue (approved):

PR #1432, opened 2022-11-15
GT-1860 refactor activity view to swiftui and combine

levieggertcru:
This PR changes AccountView.swift and associated subviews to use SwiftUI and Combine Publishers rather than UIKit.  Changes in this PR include:

- Removed old AccountView classes (subviews, viewModels) that were using UIKit.
- Added new AccountView.swift and (subviews, viewModels) which use SwiftUI.
- Changed GlobalAnalyticsService to GlobalAnalyticsRepository and updated to use Combine publishers.
- Added UseCase for fetching global activity for AccountView > Global Activity .
- Added UseCase for checking if account creation is supported.
- Renamed ProgressView.swift which was a custom view for displaying a download progress bar.  This was renamed to DownloadProgressView.  Was having conflicts when trying to use the SwiftUI ProgressView available in iOS 14.


godtools/App/Features/Menu/Presentation/Account/AccountViewModel.swift
```diff
@@ -8,87 +8,111 @@
...
-        accountItems.accept(value: items)
+        trackSectionViewedAnalytics(sectionName: "Global Dashboard")
     }
     
-    func settingsTapped() {
-        print("TODO: AccountViewModel Implement settings.")
-    }
-
-    func accountPageDidAppear(page: Int) {
+    func globalActivityAnalyticsItemWillAppear(index: Int) -> AccountGlobalActivityAnalyticsItemViewModel {
```
rachaelblue, line 114: Naming things `willAppear` seems more event-driven/UIKit-y.  Since we're using SwiftUI now, it might be good to do away with this naming pattern and just call it `getGlobalActivityAnalyticsItem`.  Thoughts?

godtools/App/Features/Menu/Presentation/Account/AccountView.swift
```diff
@@ -2,156 +2,95 @@
...
+                        
+                        AccountGlobalActivity(
+                            viewModel: viewModel,
+                            sectionFrameWidth: sectionFrameWidth
+                        )
+                    }
+                }
+                .frame(width: sectionFrameWidth, alignment: .leading)
+                .padding(EdgeInsets(top: 32, leading: sectionHorizontalPadding, bottom: 20, trailing: sectionHorizontalPadding))
+            }//end ScrollView
```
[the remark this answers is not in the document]
levieggertcru: Hey @rachaelblue that's a good idea.  We should try and keep views into smaller components.  I will update this.

godtools/App/Features/Menu/Presentation/Account/AccountViewModel.swift
levieggertcru: Yeah you're right it does seem UIKit-y.  I will update this as well.  There will be some other places in the app that can be updated as we go along too.

levieggertcru:
Hey @rachaelblue this is ready for a second look.

rachaelblue (approved):
Looks good!

PR #1434, opened 2022-11-16
GT-1835 mobile content api auth session

rachaelblue:
This will be way easier to test once we start actually making authenticated requests.  But, I tested by passing in a get user details request URL and using Charles to rewrite the response to 401.  It did recognize the 401, fetched a new auth token, and reattempted the user request.


frett:
> This will be way easier to test once we start actually making authenticated requests. But, I tested by passing in a get user details request URL and using Charles to rewrite the response to 401. It did recognize the 401, fetched a new auth token, and reattempted the user request.

another way to "force" re-authentication is to delete the user on the mobile-content-api using the authenticated request: `DELETE /users/me`. This will cause your current session to become invalid, so your next request will require re-authentication.

godtools/App/Share/Common/URLSession/MobileContentApiAuthSession.swift
```diff
@@ -0,0 +1,115 @@
...
+    
+    private let requestBuilder: RequestBuilder = RequestBuilder()
+    
+    init(ignoreCacheSession: IgnoreCacheSession, mobileContentAuthTokenRepository: MobileContentAuthTokenRepository) {
+     
+        self.session = ignoreCacheSession.session
+        self.mobileContentAuthTokenRepository = mobileContentAuthTokenRepository
+    }
+    
+    func sendDataRequest(with urlString: String) -> AnyPublisher<Data?, URLResponseError> {
```
levieggertcru, line 26: Hey @rachaelblue I should have made this note in the Jira task, but I would like an additional method that accepts a (URLSession and URLRequest).  The idea here is MobileContentApiAuthSession would mutate the URLRequest by adding the Auth header and then use the provided URLSession to execute the URLRequest.  This would give us more modularity as the URLSession and URLRequest can be created completely separate from MobileContentApiAuthSession.

I think the urlString method is nice as well for convenience, just note that we are locked into using the IgnoreCacheSession and http methods with GET.  Maybe the naming could be a little more specific to that.  ```sendGetRequestIgnoringCache``` or something along those lines.

levieggertcru (changes requested):

godtools/App/Share/Common/URLSession/MobileContentApiAuthSession.swift
rachaelblue: Fixed!

godtools/App/Share/Common/URLSession/MobileContentApiAuthSession.swift
```diff
@@ -0,0 +1,129 @@
...
+//
+//  Created by Rachael Skeath on 11/10/22.
+//  Copyright © 2022 Cru. All rights reserved.
+//
+
+import Foundation
+import RequestOperation
+import Combine
+
+class MobileContentApiAuthSession {
```
levieggertcru, line 13: Hey @rachaelblue can you also move this class into Share/Data/MobileContentApiAuthSession/.  Common is reserved for generic classes not related to GodTools.

godtools/App/Share/Data/MobileContentAuthTokenRepository/MobileContentAuthTokenRepository.swift
```diff
@@ -35,7 +37,13 @@ class MobileContentAuthTokenRepository {
         }
     }
     
-    func fetchRemoteAuthToken(oktaAccessToken: String) -> AnyPublisher<String?, URLResponseError> {
+    func fetchRemoteAuthToken(oktaAccessToken: String? = nil) -> AnyPublisher<String?, URLResponseError> {
+        
+        // TODO: -  `getAccessTokenFromPersistentStore` should be updated to `withFreshTokenPerformAction` based on https://github.com/okta/okta-oidc-ios/blob/e5450a3aab5c194a7470addeef176e769a374650/Sources/AppAuth/OKTAuthState.m#L90-L93
```
levieggertcru, line 42: Hey @rachaelblue for this we do have a method available from the ```OktaAuthentication``` class which ```CruOktaAuthentication``` class derives from.  There is a publisher ```cruOktaAuthentication.renewAccessTokenPublisher()```.
I'm fine if we create another task for this.

godtools/App/Share/Common/URLSession/MobileContentApiAuthSession.swift
frett: to further add to rationale behind this, several of the new protected APIs take POST and PUT requests to submit data to the API. So, something hardcoded to be GET only has limited utility

godtools/App/Share/Common/URLSession/MobileContentApiAuthSession.swift
rachaelblue: Yep!

levieggertcru (approved):

PR #1437, opened 2022-11-18
use `renewAccessTokenPublisher` to fetch fresh Okta token

rachaelblue:
https://jira.cru.org/browse/GT-1869

Branched off of #1434


godtools/App/Share/Data/MobileContentAuthTokenRepository/MobileContentAuthTokenRepository.swift
```diff
@@ -58,6 +56,39 @@ class MobileContentAuthTokenRepository {
...
+                .setFailureType(to: URLResponseError.self)
+                .eraseToAnyPublisher()
+            
+        } else {
+            
+            return renewOktaAccessTokenPublisher()
+        }
+    }
+    
+    private func renewOktaAccessTokenPublisher() -> AnyPublisher<String, URLResponseError> {
```
levieggertcru, line 73: Hey @rachaelblue I've been thinking about the MobileContentAuthTokenRepository class and I would like to limit it's responsibilities by removing the CruOktaAuthentication dependency.

MobileContentAuthTokenRepository responsibilities would consist of communication with our mobile content api and auth token persistence.  A nice feature to have in this class would be to check the expiration date before fetching from the persisted cache and if an auth token has expired fetch a new one.  I also envision this class will be expanded on to support access tokens by type such as (Okta, Facebook, Google SignIn, Apple SignIn) as the Mobile Content API will need to provide endpoints for those additional access tokens if we plan to drop Okta in GodTools.  Just something to think about.

As far as obtaining new access tokens and renewing access tokens, we really don't have a place for this.  At the moment we directly authenticate from a UseCase using CruOktaAuthentication.  I'm thinking it would be nice to have a data layer class that is responsible for the application's authentication state.  It could end up encapsulating authentication classes such as CruOktaAuthentication, classes for Facebook sign-in, Google sign-in and Apple sign-in etc.  It could also track if the app is currently authenticated, attempt to renew access tokens, perform sign out, etc. 

What are your thoughts on having that separate class?  I'm trying to think more SOLID and at the same time find a good balance.

frett: just to chime in, the proposed mechanism to support multiple authentication methods on the mobile-content-api is to just tweak the authentication request based on the provider. the auth token that mobile-content-api returns will be consistent no matter the authentication provider.

On Android I abstracted out generation of the auth request so that it can be implemented by each supported auth provider. Everything else about authenticated requests stays the same

levieggertcru: Thanks @frett that is good to know.

rachaelblue: Hey @levieggertcru, I think you're right that the Okta renewal piece shouldn't be in the repository.  I think that's why I originally had a separate use case for getting the auth token, to first grab the Okta token and pass it into the repository.  If we had a separate class that encapsulated authentication and returned fresh tokens, I guess MobileContentApiAuthSession could be responsible for passing the access token into the repo?

levieggertcru: Hey @rachaelblue yep that's right, MobileContentApiAuthSession would pass the access token along to the MobileContentAuthRepository to fetch the api auth token.
It would also reference the class that handles authentication and fetch the access token via a getAccessTokenPublisher().

rachaelblue: Ok I moved the Okta token logic into a new class `UserAuthentication` that can hopefully evolve as other sign on options are added.  I had to move some of the logic from `MobileContentAuthRepository` to  `MobileContentApiAuthSession` since we only need an Okta token if there's not already a cached api auth token, and that check was previously handled in the repo.

levieggertcru (approved):
Hey @rachaelblue looks good.  Created an additional task for UserAuthentication.swift which will need to go through our UseCase once we start tracking authentication  (https://jira.cru.org/browse/GT-1874?atlLinkOrigin=c2xhY2staW50ZWdyYXRpb258aXNzdWU%3D).

PR #1438, opened 2022-11-23
GT-1865 fetch user details

rachaelblue:
https://jira.cru.org/browse/GT-1865

Fetching user details when a user opens their account page from menu and displaying the "Joined on" string:
Screen Shot 2022-11-28 at 1 00 15 PM


levieggertcru (changes requested):

godtools/App/Share/Data/UserDetailsRepository/Cache/RealmUserDetailsCache.swift
```diff
@@ -0,0 +1,41 @@
...
+    
+    func getUserDetailsChanged() -> AnyPublisher<Void, Never> {
+        
+        return realmDatabase.openRealm().objects(RealmUserDetails.self)
+            .objectWillChange
+            .eraseToAnyPublisher()
+    }
+    
+    func getUserDetails() -> UserDetailsDataModel? {
+        guard let realmUserDetails = realmDatabase.openRealm().objects(RealmUserDetails.self).first else { return nil }
```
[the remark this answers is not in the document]
rachaelblue: I did previously have the user ID in the keychain as part of the primary key that was used to retrieve the token.  But the issue with that was that later I didn't necessarily know what the user ID was in order to use it to retrieve the auth token.  So, I took it out of the primary key under the assumption that we should really only ever have 1 auth token in the keychain.  And I would think same goes from UserDetail objects-- we should realistically only ever have one in Realm, right?

You're right though that it would still be safer to retrieve the UserDetails using the ID (even if there's only 1 object in Realm).  How do you feel about having the user ID stored in the keychain separately from the auth token?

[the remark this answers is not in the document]
levieggertcru: Hmm I do like the idea of using the user ID as a primary key for fetching the auth token too.  It feels safer to have a primary key for fetching those.  And I think storing that user ID separately is a good idea.  If a user were to sign out we could always do some clean up by deleting their auth token / user ID from the Keychain.  I think the UserDetails we can always leave in the RealmDatabase.  If a user were to sign in with multiple accounts on their device we would then have that data persisted and available.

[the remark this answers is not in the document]
frett: a potential flow that could cause more than 1 `UserDetail` object to be present in realm:
1. User logs into app as User A
2. User realizes they used the wrong account and logs out
3. User then logs into app as User B

there would be multiple `UserDetail` objects in realm then unless you delete user related data on logout (which could be flaky)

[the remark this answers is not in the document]
rachaelblue: Sounds good!  Yeah I was assuming we'd probably want to clear everything out on sign out.  But if Okta logs out automatically sometimes that could be harder to track.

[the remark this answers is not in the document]
frett: yeah, and with us moving to support other authentication mechanisms, I'm not sure what they would support for callbacks

rachaelblue:
Ok I added the userID to the keychain when we get the response from the auth request, and then that's used to pull user details from Realm (and the auth token from the keychain too) @levieggertcru

godtools/App/Share/Data/UserDetailsRepository/Cache/RealmUserDetailsCache.swift
```diff
@@ -0,0 +1,44 @@
...
+    }
+    
+    func getUserDetailsChanged() -> AnyPublisher<Void, Never> {
+        
+        return realmDatabase.openRealm().objects(RealmUserDetails.self)
+            .objectWillChange
+            .eraseToAnyPublisher()
+    }
+    
+    func getUserDetails() -> UserDetailsDataModel? {
```
levieggertcru, line 33: Hey @rachaelblue I think this method could be a little more descriptive since it's accessing authTokenRepository to fetch the authorized userId.  Something like ```getAuthUserDetails()```.

godtools/App/Share/Data/MobileContentAuthTokenRepository/Cache/MobileContentAuthTokenCache.swift
```diff
@@ -29,8 +29,25 @@ class MobileContentAuthTokenCache {
...
     }
     
-    func getAuthToken() -> String? {
+    func getAuthToken(for userId: String) -> String? {
         
-        return keychainAccessor.getMobileContentAuthToken()
+        return keychainAccessor.getMobileContentAuthToken(userId: userId)
+    }
+    
+    func storeUserId(_ userId: String) {
```
levieggertcru, line 37: I would suggest forcing both the userId and authToken to be stored together since those are paired together rather than having separate methods to store each.  Having separate methods to store each could cause inconsistencies if one value were stored and another wasn't.

godtools/App/Share/Data/MobileContentAuthTokenRepository/MobileContentAuthTokenRepository.swift
```diff
@@ -36,7 +37,14 @@ class MobileContentAuthTokenRepository {
...
     func getCachedAuthToken() -> String? {
-                
-        return cache.getAuthToken()
+        
+        guard let userId = getUserId() else { return nil }
+        
+        return cache.getAuthToken(for: userId)
+    }
+    
+    func getUserId() -> String? {
```
levieggertcru, line 46: Same here, I would suggest returning a single type that encapsulates the authToken and userId since those are paired together.

levieggertcru (changes requested):

godtools/App/Share/Data/MobileContentAuthTokenRepository/MobileContentAuthTokenRepository.swift
rachaelblue: The issue is that they aren't actually stored together in the keychain, so it seems unnecessary
 to pull out both the token and the ID from the keychain when sometimes only the ID is needed.  Thoughts?

levieggertcru: Hey @rachaelblue I think having additional methods to fetch a single id and single token is okay too since the repository is just a store and in this case storing an auth object.   

I do think there should be a method that encapsulates those into a single type provided by the repository since those attributes belong to one another.  From an outside view into the repository it has know knowledge of the underlying cache.  Just that an auth token or auth object can be fetched from the store.

rachaelblue: Ok I made a `MobileContentAuthTokenRepoModel` that includes the token and id.  Is there a naming pattern that makes more sense for model objects that come from the repository other than "repoModel"?  We're kind of creating a separate layer between data and domain now.

levieggertcru: Hey @rachaelblue I think something suffixed with ```DataModel``` for the repository.  It's going to have a similar shape to our api models and cache models.  I'm wondering now if our api models should be suffixed with something other than ```DataModel```, such as ```Codable``` or ```Decodable```?  
A lot of our repositories are currently returning the underlying api codable / decodable models, but after our call yesterday, it seems we want to be creating a separate DataModel for the repositories.

rachaelblue:
Hey @levieggertcru, made the change from:
-  `MobileContentAuthTokenDataModel` --> `MobileContentAuthTokenDecodable` 
- `MobileContentAuthTokenRepoModel` --> `MobileContentAuthTokenDataModel`

And now the authTokenRepo returns that data model (in addition to the individual token/id getters).

godtools/App/Share/Domain/UseCases/GetUserDetailsUseCase/GetUserDetailsUseCase.swift
```diff
@@ -0,0 +1,79 @@
...
+    private let localizationServices: LocalizationServices
+    
+    init(repository: UserDetailsRepository, localizationServices: LocalizationServices) {
+        
+        self.repository = repository
+        self.localizationServices = localizationServices
+    }
+    
+    
+    func getUserDetailsPublisher() -> AnyPublisher<UserDetailsDomainModel?, Never> {
```
levieggertcru, line 24: Hey @rachaelblue a couple suggestions here.

*  What if instead of adding the ```fetchRemoteUserDetailsPublisher``` here, we provide a ```reloadFromRemote``` flag on the ```UserDetailsRepository getUserDetailsChanged()``` which would then trigger the changed publisher again after remote has finished.  Then we get some re-usability if we added this to any other UseCases.
```swift
func getUserDetailsChanged(reloadFromRemote: Bool) -> AnyPublisher<Void, Never> {
        
        if reloadFromRemote {
            
            fetchRemoteUserDetails()
                .sink { finished in
                    
                } receiveValue: { (userDataModel: UserDetailsDataModel) in
                    
                }
                .store(in: &cancellables)

        }
        
        return cache.getUserDetailsChanged()
    }
```
* Another is changing the return type to a non-optional value ```UserDetailsDomainModel``` and instead provide a default value the UI can display.  This way the ViewModel doesn't have to do any unwrapping and can always have a value ready to display.  Which in our case is just an empty string.

```swift
private func getDefaultUserDetails() -> UserDetailsDomainModel {
        return UserDetailsDomainModel(
            joinedOnString: ""
        )
    }
    
    func getUserDetailsPublisher() -> AnyPublisher<UserDetailsDomainModel, Never> {
        
        return repository.getUserDetailsChanged(reloadFromRemote: true)
            .flatMap { () in
                    
                guard let userDetails = self.repository.getAuthUserDetails() else {
                    
                    return Just(self.getDefaultUserDetails())
                        .eraseToAnyPublisher()
                }
                
                let joinedOnString = self.buildJoinedOnString(from: userDetails)
                
                return Just(UserDetailsDomainModel(joinedOnString: joinedOnString))
                    .eraseToAnyPublisher()
            }
            .eraseToAnyPublisher()
    }
```

levieggertcru (changes requested):
Hey @rachaelblue looking good!  Just a couple other suggestions and this is ready!

godtools/App/Share/Data/MobileContentAuthTokenRepository/Cache/MobileContentAuthTokenKeychainAccessor.swift
```diff
@@ -11,34 +11,44 @@ import Security
 
 class MobileContentAuthTokenKeychainAccessor {
```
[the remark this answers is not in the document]
rachaelblue: Yep!  If there's an existing userId/authToken in the keychain already, it does an update query to change the existing item.

godtools/App/Share/Domain/UseCases/GetUserDetailsUseCase/GetUserDetailsUseCase.swift
rachaelblue: Good idea about the `UseretailsDomainModel`, I went ahead and made that change!

It feels a little odd to me for `getUserDetailsChanged` to have such a huge side effect as pulling from the remote.  Also then the repository will be responsible for holding onto that cancellable instead of the class that initiated it.  It seems like if there's a scenario where you'd want to pull from the remote and get the userDetails changed publisher, then it'd probably make sense to just use `GetUserDetailsUseCase`?

levieggertcru: Hey @rachaelblue I need to think about this one some more.  I feel like the data layer should be responsible for invalidating the cache which would trigger a new fetch from remote.  Currently we don't have a common solution for this. It would be nice if there could be some common cache control policy on the data layer that allows more control of how data can be fetched whether it be from cache only, remote only, cache and remote, etc.  I was looking through the Apollo GraphQL SDK a little bit and I like their setup https://www.apollographql.com/docs/ios/fetching/fetching-data/#cache-policies.   For now I'm fine with leaving the fetch in the UseCase until we can come up with a better solution too. 

Another thing I did just think of is the location of this UseCase in ```Share/Domain/``` and possibly the naming.  I would like us to start getting more specific with UseCases and having those live in a ```{Feature_Name}/Domain/UseCases/``` folder . 

The Account presentation files currently live in ```Features/Menu/``` so a good place for this UseCase would be in ```Features/Menu/Domain/UseCases/```.  As the Account grows feel free to create a specific Feature for the Account as well if you'd like because I know you will be adding more logic to the Account section soon.

As far as the UseCase naming, we could have 2 use cases, 1 responsible for the account creation date and another for the account user name ```GetUserAccountCreationDateUseCase``` and ```GetUserAccountNameUseCase``` or a single UseCase ```GetUserAccountDetailsUseCase``` for handling fetching the name and creation date and applying business rules.  This way we get away from the "Quazi" clean pattern.

wrandall22: One thing that Daniel B has noted on the web side is that Apollo is a fairly large library there. It might be something to watch for here as well for app size.

levieggertcru (approved):

godtools/App/Share/Domain/UseCases/GetUserDetailsUseCase/GetUserDetailsUseCase.swift
levieggertcru: Hey @wrandall22 that's good to know about the Apollo library size.  I'm still hesitant about adding Apollo into our own projects. There would be some benefits, but as far as cache syncing we'd still have to find alternative solutions for that.

levieggertcru: Hey @rachaelblue yes a ticket would be good to have down for this.  Our repositories will most likely be duplicating logic around fetching and syncing of data, so some type of common solution for this would be nice to have that our repositories can make use of.

rachaelblue: Ticket here: https://jira.cru.org/browse/GT-1887

PR #1441, opened 2022-12-01
GT-1374 standardize clickable content

levieggertcru:
This PR implements the following:
- Adds protocol ```ClickableMobileContentViewModel``` with an extension for sharing the analytics tracking when a clickable view is tapped.
- Updated the Clickable views to send URL events if a URL is attached to the Clickable model.
- Changed arguments ```url: String``` to ```url: URL``` because ```Clickable``` provides a URL type.


frett:
@levieggertcru let's talk about this on a call, there might be some better abstraction that can happen for this ticket

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentButton/MobileContentButtonView.swift
```diff
@@ -201,7 +201,10 @@ class MobileContentButtonView: MobileContentView {
             super.sendEventsToAllViews(eventIds: viewModel.buttonEvents, rendererState: viewModel.rendererState)
         
         case .url:
```
frett, line 203: you should actually be moving away form using the button type, you should treat a button just like any other `Clickable` object. Trigger any content events attached to it and open any url defined.

levieggertcru: Hey @frett that sounds good!  Let's plan that for next week Monday.  I'm going to be out tomorrow and rest of day today.

godtools/App/Services/Renderer/Views/MobileContent/Views/ClickableMobileContentViewModel.swift
```diff
@@ -0,0 +1,39 @@
...
+import GodToolsToolParser
+
+protocol ClickableMobileContentViewModel {
+        
+    func getClickableUrl() -> URL?
+}
+
+extension ClickableMobileContentViewModel {
+    
+    func getClickableEvents(model: Clickable) -> [AnalyticsEvent] {
```
frett, line 19: maybe `getClickableAnalyticsEvents()`

godtools/App/Services/Renderer/Views/MobileContent/Views/ClickableMobileContentViewModel.swift
```diff
@@ -0,0 +1,39 @@
...
+//  Created by Levi Eggert on 11/30/22.
+//  Copyright © 2022 Cru. All rights reserved.
+//
+
+import Foundation
+import GodToolsToolParser
+
+protocol ClickableMobileContentViewModel {
+        
+    func getClickableUrl() -> URL?
```
frett, line 14: you could also include `getClickableEvents() -> [EventId]`

levieggertcru:
@frett @rachaelblue this is ready for another review.  There's a lot of file changes because I ended up removing all of the ```ViewModel``` ```ViewModelType``` ```protocols``` that were in the ```Renderer``` which is something myself and Rachael have been in the process of as we switch over to SwiftUI.  

This PR includes the following changes with most important at the top of this list:

* Added ```MobileContentViewModel.swift``` which is a Base ViewModel that Base ```MobileContentView``` references.  In this class is a reference to a ```GodToolsToolParser``` ```BaseModel```.  ```MobileContentViewModel``` contains shared logic for Clickable models and shared logic for tracking events when a MobileContentView enters and exits view.  
The MobileContentViewModel also references ```renderedPageContext: MobileContentRenderedPageContext``` which encapsulates information about the current renderer state.

* Added a ```viewTapped()``` method to ```MobileContentView``` which processes shared logic for ```Clickables```` and calls MobileContentViewModel viewTapped().

* Removed ```ViewModelType``` protocols for Views in the ```Renderer```.

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentButton/MobileContentButtonView.swift
```diff
@@ -195,21 +195,7 @@ class MobileContentButtonView: MobileContentView {
...
-            super.sendButtonWithUrlEventToRootView(url: viewModel.buttonUrl)
-       
-        case .unknown:
-            break
-        default:
-            break
-        }
-        
-        viewModel.buttonTapped()
+        super.viewTapped(mobileContentAnalytics: viewModel.mobileContentAnalytics)
```
frett, line 198: 💯

godtools/App/Services/Renderer/Views/MobileContent/Views/MobileContentViewModel.swift
```diff
@@ -0,0 +1,117 @@
...
+        mobileContentAnalytics.trackEvents(
+            events: getClickableAnalyticsEvents(),
+            renderedPageContext: renderedPageContext
+        )
+    }
+    
+    func getClickableAnalyticsEvents() -> [AnalyticsEvent] {
+        
+        guard let clickableModel = baseModel as? Clickable, clickableModel.isClickable,
+              let modelHasAnalyticsEvents = clickableModel as? HasAnalyticsEvents else {
```
frett, line 92: after this next release I'm going to look at everything that implements `Clickable` and see if they also implement `HasAnalyticsEvents`. If so I might make `Clickable` extend `HashAnalyticsEvents`, so this check would become redundant.

But that will be a future cleanup we can do and not something to include as part of this PR.

frett (approved):

PR #1443, opened 2022-12-02
GT-1873 fetch user counters

rachaelblue:
Added a user counter API to fetch all user counters and increment a single counter, as well as repository and cache to store them in Realm.  The UserCounter objects maintains the API count separately from the incremented value from the device, since we only include the "delta" value in the request and not the total count.


godtools/App/Share/Data/UserCountersRepository/UserCountersRepository.swift
```diff
@@ -0,0 +1,73 @@
...
+                return self.cache.syncUserCounters(userCountersResponse.userCounters)
+                    .mapError { error in
+                        return URLResponseError.otherError(error: error)
+                    }
+                    .eraseToAnyPublisher()
+            }
+            .eraseToAnyPublisher()
+    }
+    
+    func incrementUserCounter(_ userCounter: UserCounterDataModel) -> AnyPublisher<RealmUserCounter, URLResponseError> {
```
levieggertcru, line 36: Hey @rachaelblue same comment here as in ```UserCountersAPI```.  I'd suggest changing the argument here to ```(id: String, increment: Int)``` or a type that encapsulates the attributes required for an increment.

levieggertcru (changes requested):
Hey @rachaelblue let me know if you want to catch up on some of these change requests.  I'd like to get a better understanding behind the sync too if you have some time today.

rachaelblue:
Hey @levieggertcru, this is ready for a 2nd look.  I split the user counter data model up into two separate ones like we discussed a while back-- `UserCounterDataModel` for the repo/cache layer and `UserCounterDecodable` for the API layer.  This allows the "decodable" one to only contain the attributes we get directly from the API response, but we can still track the "increment/delta" counter value through "dataModel" and Realm.  You'll see in the sync logic that we do have to account for this when syncing to make sure increments don't get overwritten when we update from the remote.

I also adjusted the increment logic, thinking that it may make more sense to batch increment requests.  So, when we want to increment a counter, calling `incrementCachedUserCounterBy1` will increment the local value, but not push it to the remote.  Then we'd later call `syncUpdatedUserCountersWithRemote` to push all the updates to the remote at once, rather than doing them one at a time.  I think this is the way Android does it based on the pseudocode @frett had sent me.  Let me know if you have questions about anything.

godtools/App/Share/Data/UserCountersRepository/Api/Models/GetUserCountersResponseDecodable.swift
```diff
@@ -0,0 +1,25 @@
...
+//  GetUserCountersResponseDecodable.swift
+//  godtools
+//
+//  Created by Rachael Skeath on 11/29/22.
+//  Copyright © 2022 Cru. All rights reserved.
+//
+
+import Foundation
+
+struct GetUserCountersResponseDecodable: Decodable {
```
levieggertcru, line 11: Hey @rachaelblue created this task that will do away with having to create multiple Decodable types when decoding ```data``` attributes.  There's quite a few Decodable types I've created as well.  (https://jira.cru.org/browse/GT-1894)

levieggertcru (approved):
Hey @rachaelblue looks good!  I like keeping the Decodable model clean and adding those additional attributes on the Cache Model layer since that logic is handled there.

godtools/App/Share/Data/UserCountersRepository/Api/Models/GetUserCountersResponseDecodable.swift
rachaelblue: Sounds good!

PR #1456, opened 2022-12-27
GT-1739 Fix Xcode Warnings in godtools target

levieggertcru:


levieggertcru:
This PR contains the following changes:

- Silence local variable not being used warnings.
- Silence return value not being used warnings.
- Remove deprecated mainThreadRealm instance variable from RealmDatabase.
- Remove references to deprecated LanguageSettingsService in place of LanguageSettingsRepository. 
- Remove IsNewUserService in place of using Repositories and UseCases for checking if Onboarding Tutorial is available and if Setup Parallel Language is available for presenting to the user.

PR #1463, opened 2023-01-04
Upgrading GitHub Actions Workflows Xcode Version

levieggertcru:
Upgrade to Mac OS 13 and Xcode 14.2.


.github/workflows/build.yml
```diff
@@ -15,9 +15,9 @@ concurrency:
 
 jobs:
   build:
-    runs-on: macos-12
+    runs-on: macos-13
     env:
-      DEVELOPER_DIR: /Applications/Xcode_14.0.app/Contents/Developer
+      DEVELOPER_DIR: /Applications/Xcode_14.2.app/Contents/Developer
```
frett, line 20: it's kind of annoying that you have to specify the Xcode version like this. I did find a GHA action that supposedly will select the XCode version for you: https://github.com/marketplace/actions/setup-xcode-version

It might be a bit more readable that just a directory path being specified

frett (dismissed):

.github/workflows/build.yml
levieggertcru: Thanks @frett.  I will look into using Xcode select.

It appears I will have to put this PR on hold as mac-os 13 isn't available yet (https://github.com/actions/runner-images/issues/6426)

levieggertcru: @frett @rachaelblue this is ready again for review.
GitHub Action workflows are set to run on macos-13 and use Xcode version 14.2

frett (approved):

PR #1464, opened 2023-01-05
GT-1895 remove deprecated translation downloader related files

levieggertcru:
This PR removes files related to the old translation downloading since switching over to use Combine publishers.  

This biggest change is removing protocol ResourceItemInitialDownloadProgress.swift which our ToolCard ViewModels were implementing. @rachaelblue this means that the progress bars on the ToolCards could also be removed since we are no longer tracking a download progress.


rachaelblue (approved):
Looks good!  Shall I make a followup ticket to remove the progress bars on ToolCard?

levieggertcru:
Hey @rachaelblue yeah can you create a ticket to remove those, I don't see those getting introduced back.

PR #1468, opened 2023-01-10
GT-1898 increment counters for interactions

rachaelblue:


godtools/App/Services/Renderer/Views/MobileContent/Views/Pages/MobileContentPagesViewModel.swift
```diff
@@ -61,12 +63,31 @@ class MobileContentPagesViewModel: NSObject {
...
     }
     
+    private func setUpCountLanguageUsages() {
+        
+        currentPageRenderer.flatMap { currentPageRenderer -> AnyPublisher<UserCounterDomainModel, Error> in
+            
+            let currentLocaleId = currentPageRenderer.language.localeIdentifier
+            let locale = Locale(identifier: currentLocaleId)
+            
+            return self.incrementUserCounterUseCase.incrementUserCounter(for: .languageUsed(locale: locale))
```
levieggertcru, line 81: Hey @rachaelblue I like this idea here, the only problem is there are situations when a page renderer may re-trigger the same page.  So we would end up duplicating language counts here.

One case is when a tool is first launched I believe AutoLayout did finish layout constraints will re-trigger a page renderer so this would get captured twice when a tool opens.  There may be other cases that re-trigger a current page to render.

Instead of sinking you could go down below into the ```setPageRenderer(pageRenderer:``` method and check for a language change and if so trigger the counter.

```

func setPageRenderer(pageRenderer: MobileContentPageRenderer) {

let languageChanged: Bool = currentPageRenderer.value.language.localeIdentifier != pageRenderer.language.localeIdentifier
        
        if languageChanged {
            print("trigger language counter")
        }

....
```

levieggertcru (changes requested):

godtools/App/Features/ToolSettings/LoadToolRemoteSession/LoadToolRemoteSessionViewModel.swift
```diff
@@ -39,4 +45,15 @@ class LoadToolRemoteSessionViewModel: LoadingViewModelType {
         
         flowDelegate?.navigate(step: .cancelledLoadingToolRemoteSession)
     }
+    
+    func pageViewed() {
+        
+        incrementUserCounterUseCase.incrementUserCounter(for: .screenShare(tool: resourceId))
```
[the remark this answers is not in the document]
frett: just user initiated ones

rachaelblue:
Thanks @levieggertcru!  Got those changes made.

godtools/App/Features/Lesson/Lesson/LessonViewModel.swift
```diff
@@ -27,6 +29,18 @@ class LessonViewModel: MobileContentPagesViewModel {
         closeTapped()
     }
     
+    override func viewDidFinishLayout(window: UIViewController, safeArea: UIEdgeInsets) {
+        super.viewDidFinishLayout(window: window, safeArea: safeArea)
+        
+        incrementUserCounterUseCase.incrementUserCounter(for: .lessonOpen(tool: resource.id))
```
levieggertcru, line 35: Oh yeah this isn't even needed on the pageDidAppear() life cycle.  I didn't catch that these were opens.

levieggertcru: viewDidFinishLayout() works.

levieggertcru (approved):

PR #1470, opened 2023-01-13
GT-1908 pusher user counters to remote

rachaelblue:


godtools/App/Share/Domain/UseCases/IncrementUserCounterUseCase/IncrementUserCounterUseCase.swift
```diff
@@ -41,9 +41,19 @@ class IncrementUserCounterUseCase {
                 .eraseToAnyPublisher()
         }
         
+        // Only need to count language usage once per language.
+        if case .languageUsed = interaction, let languageAlreadyUsedCounter = userCountersRepository.getUserCounter(id: userCounterId) {
```
levieggertcru, line 45: Hey @rachaelblue is this essentially capping a language usage at count of 1?

rachaelblue: Hey @levieggertcru, yes-- this is a business rule that we only need to log language usage one time (per language), so count will be limited to 1 here.  Although per @frett, it won't hurt if we're continuing to count usage beyond 1, it's just not necessary.

godtools/App/Share/Data/UserCountersRepository/UserCountersRepository.swift
```diff
@@ -42,6 +49,14 @@ class UserCountersRepository {
     
     func syncUpdatedUserCountersWithRemote() {
         
+        if cancellables.isEmpty == false {
```
levieggertcru, line 52: This is fine as long as ```cancellables``` is strictly used for syncing updated counters.   If other publishers were to sink and store cancellables in the ```cancellables``` ```Set``` then those could potentially be cancelled as well.  For situations like these instead of using ```.store(in: &cancellables)``` I will create an instance variable of ```AnyCancellable?``` that can reference the returned AnyCancellable and cancel the reference.

godtools/App/Share/Domain/UseCases/IncrementUserCounterUseCase/IncrementUserCounterUseCase.swift
levieggertcru: Hey @rachaelblue it doesn't hurt to cap it on our end as well.  I would probably just let the counter increment and instead on the fetch if the count is greater than 1 cap it to 1.  That way the data layer stays unchanged, we're just capping the returned value as part of the business rules.  But either way has advantages so I'm good with capping here too.

levieggertcru (changes requested):
Hey @rachaelblue looks good, I think instead of cancelling the entire ```Set``` though use an instance variable that can reference the AnyCancellable and be cancelled.  This would avoid any potential future publishers being cancelled if that  ```cancellables``````Set``` were to be shared.

godtools/App/Share/Domain/UseCases/IncrementUserCounterUseCase/IncrementUserCounterUseCase.swift
frett: capping the language usage counter is a really soft business rule at this point. When consuming the language counters for `UserActivity` we just check to see if a language counter is `>= 1`. I capped counting within a single tool session so that if the end user switches back and forth between parallel languages it doesn't artificially inflate the numbers.

Long term we might use these counts to determine a user's most frequently used language or something like that, at which point we might get a bit more strict on what constitutes incrementing a count, and would make capping the numbers when synced from the API untenable. But that isn't on any immediate roadmaps

frett: which what I described with how I count languages on Android is different than how Rachael is counting here. Reopening a tool will still increment the language counter even if it was already incremented at least once. It's just within a single tool usage session we only increment it once per language

levieggertcru: Hey @frett that makes sense.  How do you define a session or new session in Android?  Would changing user accounts start a new session?  Would the app have to be in the background for certain time period to start a new session?

levieggertcru: Hey @frett does incrementing ```.languageUsed``` include ```lessons``` and ```cyoa``` or should it be ```tracts``` only?

frett: I'm incrementing it for cyoa and lessons

levieggertcru: Okay thanks then we are good since the base ```MobileContentPagesViewModel.swift``` covers all of those.

frett: I did the same thing, I put it in my base renderer UI that is shared across all the type :)

godtools/App/Services/Renderer/Views/MobileContent/Views/Pages/MobileContentPagesViewModel.swift
```diff
@@ -367,17 +368,35 @@ class MobileContentPagesViewModel: NSObject {
     
     private func countLanguageUsage(localeId: String) {
         
+        if languageUsageAlreadyCountedThisSession(localeId: localeId) { return }
```
levieggertcru, line 371: Hey @rachaelblue just tested and working good!

godtools/App/Share/Data/UserCountersRepository/UserCountersRepository.swift
rachaelblue: Hey @levieggertcru, I went ahead and moved this entire method into a separate class `RemoteUserCountersSync`.  You're totally right, we need to keep the cancellables from the remote sync completed separated, and just having a separate instance variable still felt dangerous and potentially confusing in the future if we didn't remember which to use where.  So I think it's better encapsulated in a separate class altogether!

levieggertcru: I like!

levieggertcru (approved):

PR #1475, opened 2023-01-18
GT-1875 Fix strange scroll effect in iOS 16

levieggertcru:
There was a strange scroll effect that was occurring in iOS 16 for favorites, tool spotlight, and tool categories. (https://jira.cru.org/browse/GT-1875)

Found a related issue posted to Stackoverflow here:  https://stackoverflow.com/questions/74042635/nested-scrollview-in-a-list-refreshable-strange-behaviour-in-ios-16

Seems to be fixed by using a ScrollView instead of a List.  It's nice to use a ScrollView here too since we don't have to worry about how the System might style the List element and since we have custom content going into the ScrollView.

Also removed some iOS 13 code in BackwardsCompatibleList.swift since we're now pointing to iOS 14.

### DIFF
--- a/godtools/App/Features/Dashboard/Presentation/Home/FavoritesContentView.swift
+++ b/godtools/App/Features/Dashboard/Presentation/Home/FavoritesContentView.swift
@@ -43,6 +43,7 @@ struct FavoritesContentView: View {
                             .foregroundColor(ColorPalette.gtGrey.color)
                             .padding(.top, 12)
                             .padding(.bottom, 15)
+                            .padding(.leading, leadingTrailingPadding)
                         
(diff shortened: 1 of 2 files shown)


rachaelblue (commented):
This seems to have caused some text to shift into the center in a couple spots.  Unless it was already like that? 🤔
Screen Shot 2023-01-19 at 2 46 44 PM Screen Shot 2023-01-19 at 2 47 24 PM

levieggertcru:
It might have!  I will check.

levieggertcru:
Hey @rachaelblue yeah the LazyVStack must default to a ```center``` alignment.  Updated the alignment to ```leading```.

I also set the Welcome to GodTools Text leading to be the leadingTrailingPadding defined for the FavoritesView.

leading-20

rachaelblue (approved):

PR #1476, opened 2023-01-20
GT-1894 consolidate response data decoding

levieggertcru:
This PR implements the following:

* Adds a ```Common/JsonApi/JsonApiResponseData.swift``` (https://jsonapi.org/) for quick decoding of data objects. 
* Refactors GlobalAnalyticsRepository to work against a RealmDatabase cache instead of UserDefaults and also introduces a GlobalAnalyticsDataModel that the GlobalAnalyticsRepository will return instead of returning the mobile content api decodable model.

```JsonApiResponseData``` ```Usage```

```swift
struct JsonApiResponseData<T: Decodable>: Decodable {
    
    let data: T
    
    enum CodingKeys: String, CodingKey {
        case data = "data"
    }
}
```

This reduces the need of having to create duplicate Decodable structs that wrap data objects when decoding.  We can decode both arrays and single objects of data.

Example decoding a data object with a single object.

```swift
.decode(type: JsonApiResponseData<MobileContentGlobalAnalyticsDecodable>.self, decoder: JSONDecoder())
            .mapError {
                return URLResponseError.decodeError(error: $0)
            }
            .map {
                return $0.data
            }
            .eraseToAnyPublisher()
```

Example decoding a data object with an array of objects.

```swift
.decode(type: JsonApiResponseData<[LanguageModel]>.self, decoder: JSONDecoder())
            .mapError {
                return URLResponseError.decodeError(error: $0)
            }
            .map {
                return $0.data
            }
            .eraseToAnyPublisher()
```


rachaelblue (approved):
Nice!  Will have to go back and start using this when decoding.

PR #1477, opened 2023-01-20
GT-1834 tab bar spacing

rachaelblue:
1. adjusted individual tab item spacing
2. added padding at the bottom of the scrollViews so the last item isn't cut off


godtools/App/Features/Dashboard/Presentation/Tools/AllToolsContentView.swift
```diff
@@ -39,6 +39,10 @@ struct AllToolsContentView: View {
                         
                         AllToolsList(viewModel: viewModel, width: width, leadingTrailingPadding: leadingTrailingPadding)
                         
+                        Rectangle()
+                            .frame(height: 80)
```
levieggertcru, line 43: Hey @rachaelblue this would work if the Dashboard TabBar height was fixed to 80 and that value was stored in a constant that could be used in these 3 places.  

I would prefer something more adaptive here instead of a fixed spacing.  

That could be achieved by having a containing view in the Dashboard where the bottom of the container touches the top of the TabBar.  Then any views in the dashboard would be placed inside of the container. 

dashboard-view

levieggertcru (changes requested):

godtools/App/Features/Dashboard/Presentation/Tools/AllToolsContentView.swift
rachaelblue: @levieggertcru ok so I changed the `ZStack` that held the tabs and tab bar to a `VStack`, which solved in the initial problem.  But in doing so, I discovered that the view modifier to hide the system tab bar wasn't working like I had expected, and on iOS 16 it was now visible above the custom tab bar when a tab appears for the 2nd time.  I had trouble finding much information about how it should be used... it seems like people either use the system tab bar, or they create a custom one completely from scratch (not the hybrid method we currently have).  So I guess that's an option moving forward if we continue to have difficulties, but for now doing `UITabBar.appearance().isHidden = true` works properly on all iOS versions.

levieggertcru: Thanks @rachaelblue I'll take a look!  Yeah sometimes I prefer to create from scratch, especially if I'm fighting the system.  It's sometimes easier to just build those components from scratch and this way you know it's going to work correctly across iOS versions.

godtools/App/Features/Dashboard/Presentation/Dashboard/DashboardView.swift
```diff
@@ -17,41 +17,25 @@ struct DashboardView: View {
     init(viewModel: DashboardViewModel) {
         self.viewModel = viewModel
         
-        if #available(iOS 16.0, *) { } else {
-            UITabBar.appearance().isHidden = true
-        }
+        // Note: In iOS 16+, the preferred way to hide the tab bar is through the view modifier `.toolbar(.hidden, for: .tabBar)`, but there seems to be a bug where it only hides on the tab's first appearance.  May want to reinvestigate this in future versions.
+        UITabBar.appearance().isHidden = true
```
levieggertcru, line 21: Just to be safe it would be good to use the ```whenContainedInInstancesOf``` argument to restrict this to the DashboardView.

https://developer.apple.com/documentation/uikit/uiappearance/1615013-appearance

godtools/App/Features/Dashboard/Presentation/Dashboard/DashboardTabBarView/DashboardTabBarView.swift
```diff
@@ -16,23 +16,20 @@ struct DashboardTabBarView: View {
...
+                    .frame(maxWidth: .infinity)
                 
-                doubleSpacer()
             }
             .padding(.top, 16)
             .padding(.bottom, 8)
+            .padding([.leading, .trailing], 8)
             .frame(maxWidth: .infinity)
             .background(
                 Color.white
```
levieggertcru, line 35: I noticed the shadow appears to have a white backing behind it which looks odd when content is clipped behind the shadow.  I changed the HStack to a red background and it's more evident.

iphone8-ios16


iphone8-ios16-red

levieggertcru (changes requested):

godtools/App/Features/Dashboard/Presentation/Dashboard/DashboardTabBarView/DashboardTabBarView.swift
rachaelblue: Good catch, fixed!

rachaelblue:
Hey @levieggertcru, I got the paging added in and have it animating only when the tab selection changes.  So should be ready for another look.  Thanks!

levieggertcru (approved):
Woohoo dashboard is back animating!

PR #1480, opened 2023-01-25
GT-1907 user badges UI

rachaelblue:
The counters for lesson completion and tip completion aren't implemented yet (future tickets), so these badges can't be earned right now. All the other badges should work correctly.


godtools/App/Features/Menu/Domain/UseCases/GetUserActivityBadgeUseCase/GetUserActivityBadgeUseCase.swift
```diff
@@ -0,0 +1,160 @@
...
+    private func getBadgeText(badgeType: Badge.BadgeType, progressTarget: Int) -> String {
+        
+        if progressTarget == 1 {
+            
+            return getSingularString(for: badgeType)
+            
+        } else {
+            
+            return getPluralString(for: badgeType, progressTarget: progressTarget)
+        }
```
frett, line 53: does iOS have a native way to handle plurals?

The 1/other rule applies to english, but it's not the same for all foreign languages.

frett: a quick google search found this: https://developer.apple.com/documentation/xcode/localizing-strings-that-contain-plurals

frett: I'm also not sure how well OneSky supports iOS plurals, I know it has some outdated rules for French plurals on Android

godtools/Assets.xcassets/Features/User Badge Icons/articles-1.imageset/Contents.json
```diff
@@ -0,0 +1,16 @@
+{
+  "images" : [
+    {
+      "filename" : "Group 793.pdf",
```
frett, line 4: hmm, the pdf format for SVGs was mentioned in one kotlin multiplatform shared image library I had found. Maybe I will take a look at it later to try it out.

godtools/App/Features/Menu/Domain/UseCases/GetUserActivityBadgeUseCase/GetUserActivityBadgeUseCase.swift
frett: For reference, here is where a couple plurals are defined in Android:

https://github.com/CruGlobal/godtools-android/blob/develop/app/src/main/res/values/strings_tool_details.xml#L9-L19

and you can see how those same plurals get translated to Russian:
https://github.com/CruGlobal/godtools-android/blob/develop/app/src/main/res/values-ru/strings_tool_details.xml#L9-L21

rachaelblue: Ohhhh interesting!  Good catch!  I'm thinking back to other apps I've worked on and wondering now if they were localizing correctly for plurals or not 🤔  Will make a separate ticket.

rachaelblue: ticket here: https://jira.cru.org/browse/GT-1918

godtools/App/Features/Menu/Presentation/Account/Subviews/AccountActivityView/Subviews/AccountActivityBadgeView.swift
```diff
@@ -0,0 +1,65 @@
...
+                    
+                    Image("badge_background")
+                        .foregroundColor(badge.iconBackgroundColor)
+                    Image(badge.iconImageName)
+                        .foregroundColor(badge.iconForegroundColor)
+                }
+                
+                Text(badge.badgeText)
+                    .font(FontLibrary.sfProTextRegular.font(size: 10))
+                    .foregroundColor(badge.isEarned ? ColorPalette.gtGrey.color : lightGreyTextColor)
```
levieggertcru, line 35: I think the badge textColor could even be part of the UseCase rather than checking isEarned here.

godtools/App/Features/Menu/Domain/UseCases/GetUserActivityBadgeUseCase/UserActivityBadgeDomainModel.swift
```diff
@@ -0,0 +1,29 @@
...
+//
+
+import Foundation
+import GodToolsToolParser
+import SwiftUI
+
+struct UserActivityBadgeDomainModel {
+        
+    let badgeText: String
+    let badgeType: Badge.BadgeType
```
levieggertcru, line 16: I would say ```badgeType``` and ```variant``` shouldn't be exposed here since those attributes don't apply to the presentation layer.  Instead this struct should require the ```Identifiable``` ```id``` on initialization.  Something like ```identifiableId``` would be more clear.

levieggertcru (changes requested):

godtools/Assets.xcassets/Features/User Badge Icons/articles-1.imageset/Contents.json
levieggertcru: How do PDF files work in Xcode?  Do they still get bundled with the app?  I would like to look into this one a little more, my hope is they don't increase the app size more than using compressed PNG files.

godtools/App/Features/Menu/Domain/UseCases/GetUserActivityBadgeUseCase/UserActivityBadgeDomainModel.swift
```diff
@@ -0,0 +1,23 @@
+//
+//  UserActivityBadgeDomainModel.swift
+//  godtools
+//
+//  Created by Rachael Skeath on 1/17/23.
+//  Copyright © 2023 Cru. All rights reserved.
+//
+
+import Foundation
+import GodToolsToolParser
```
levieggertcru, line 10: With the ```Badge.BadgeType``` removed we no longer need the ```GodToolsToolParser``` dependency here which does make this more clean since the domain model doesn't know about the data layer.
We haven't been fully implementing clean in our UseCases since those do reference our concrete data layer classes, but maybe in the future that is something we can start to integrate as we create a more distinct data layer.

godtools/Assets.xcassets/Features/User Badge Icons/articles-1.imageset/Contents.json
levieggertcru: I'm a little hesitant about PDF files in Xcode.  Here are some pitfalls to look out for (https://medium.com/livefront/pdf-pitfalls-in-xcode-dc372bc0df5).

Also, it looks like XCode 12 may support SVG which would be worth looking into more https://www.avanderlee.com/xcode/svg-image-assets/

frett: the SVG support is available starting with iOS 13 according to the second link you posted, does that line up with our iOS minimum version for GodTools?

levieggertcru: Yep GodTools is supporting minimum of iOS 14.

levieggertcru (approved):

PR #1483, opened 2023-01-30
GT-1926 Update shared buttons to include SwiftUI Button element.

levieggertcru:
This PR updates the shared Button views to include the SwiftUI Button element.  By doing so we get the system highlight when a button is tapped which will highlight the Text element.  Basically any elements within the Button content or highlighted and since the Rectangle is a clear color there will be no highlight effect on the Rectangle element.
- Added a Rectangle element within the Button content which is needed for the tap area of the button.


rachaelblue (commented):
Looks good with the highlight!

rachaelblue (approved):

PR #1490, opened 2023-02-13
GT-1916 capture user analytics

rachaelblue:


godtools/App/Services/Renderer/Views/MobileContent/Views/MobileContentViewModel.swift
```diff
@@ -53,27 +53,19 @@ class MobileContentViewModel: NSObject {
...
         
-        for event in analyticsEvents {
+        for event in visibleAnalyticsEvents {
             
-            let trigger: AnalyticsEvent.Trigger = event.analyticsEvent.trigger
-
-            if trigger == .visible {
-                event.trigger()
-            }
+            event.trigger()
```
rachaelblue, line 60: Talked with @frett about this last week.  Some events (like user analytics) weren't getting triggered here because `getAnalyticsEvents(type: .visible)` sometimes returns a `.default` type rather than explicitly `.visible` (on pages where visible is the default).  So we just need to make sure to use `getAnalyticsEvents(type)` to do the filtering for us.

levieggertcru: Thanks @rachaelblue!  So it seems before I was assuming all analytics events (including visible) was getting included hence the check for ```trigger == .visible```.  But, that wasn't the case right?

godtools/App/Services/Renderer/Views/Tool/Views/Page/ToolPageViewModel.swift
```diff
@@ -25,7 +25,7 @@ class ToolPageViewModel: MobileContentPageViewModel {
         self.analytics = analytics
         self.hidesCallToAction = pageModel.isLastPage
                 
-        self.analyticsEventsObjects = MobileContentAnalyticsEvent.initAnalyticsEvents(
+        self.visibleAnalyticsEventsObjects = MobileContentAnalyticsEvent.initAnalyticsEvents(
```
levieggertcru, line 28: It seems I duplicated some of this logic here for tracking page analytics in the mobile content renderer.  I'm going to create a ticket and look into consolidating some of this.

godtools/App/Services/Renderer/Views/MobileContent/Views/MobileContentViewModel.swift
rachaelblue: Well, it depended on how you got the analytics events.  In `ToolPageCardViewModel`, you were using `cardModel.analyticsEvents` rather than `cardModel.getAnalyticsEvents(type: .visible)`, so in that case any sort of event could get returned (making the `trigger == .visible` check necessary).  I changed that to `cardModel.getAnalyticsEvents(type: .visible)`, and all the other spots were already using that, so we no longer need the check.

levieggertcru: Okay that makes sense.  Thank you for updating that.

levieggertcru (approved):

PR #1491, opened 2023-02-14
GT-1834 tab content bottom spacing

rachaelblue:
Made sure the spacing is consistent at the bottom of each tab's content view.  Also fixed an issue with the blue banner view having a small white bar at the bottom (like the custom tab bar did at the top at one point).

### DIFF
--- a/godtools/App/Features/Dashboard/Presentation/Home/FavoritesContentView.swift
+++ b/godtools/App/Features/Dashboard/Presentation/Home/FavoritesContentView.swift
@@ -19,7 +19,7 @@ struct FavoritesContentView: View {
     
     var body: some View {
         
-        VStack {
+        VStack(spacing: 0) {
             if viewModel.hideTutorialBanner == false {
                 
                 OpenTutorialBannerView(viewModel: viewModel.getTutorialBannerViewModel())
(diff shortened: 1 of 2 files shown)


godtools/App/Features/Dashboard/Presentation/Home/FavoritesContentView.swift
```diff
@@ -51,9 +51,8 @@ struct FavoritesContentView: View {
                         
                         FavoriteToolsView(viewModel: viewModel.favoriteToolsViewModel, width: width, leadingPadding: leadingTrailingPadding)
                             .listRowInsets(EdgeInsets())
-                        
-                        Spacer()
-                        
+                            .padding(.bottom, 23)
```
levieggertcru, line 54: Hey @rachaelblue this fix is good for now so we can get a release out. 
I would really like to see if we can come up with a solution where the bottom padding is consistent between tabs without the need to add a specific bottom padding depending on the tab.  Can you create a ticket for that work?  I don't think it's high priority either.

levieggertcru (approved):

godtools/App/Features/Dashboard/Presentation/Home/FavoritesContentView.swift
levieggertcru: Having a default content inset on the dashboard view would be preferred.  That way it would apply to any tabbed view added to the dashboard.

PR #1498, opened 2023-02-21
GT-1933 Fix settings primary language changed observing

levieggertcru:
Hey @rachaelblue I noticed after merging ```GT-1933``` which fixed the language title observing that when I chose a new primary language from the languages list, the primary language button title wouldn't update on the prior screen.

This screen here wouldn't update after choosing Arabic:
primary-language-button-title

I didn't realize, but ```Empty()``` will cause a publisher to finish immediately.  So what happened was the first time that button was set the publisher finished and no longer received changes.
https://developer.apple.com/documentation/combine/empty

It also appears that using ```Empty(completeImmediately: false)``` still wouldn't send changes.  

```Empty()``` seems a bit misleading in the naming as you would expect an empty response. 

Instead changed to use ```map``` to return ```Void```.  Could have also used ```flatMap``` and returned ```Just(()).eraseToAnyPublisher()```.


rachaelblue:
Thanks @levieggertcru, glad you caught this!  That is odd, and not what I would think "empty" implies 🤔

PR #1506, opened 2023-02-28
GT-1945 Realm User Counter crash

rachaelblue:
Hey @levieggertcru, I was never able to reproduce this crash, but I went ahead and made the sync class return all dataModels so we don't risk having realm objects potentially being accessed from different threads.


levieggertcru:
That sounds good.  I couldn't reproduce the crash either.  Ran the app offline and generated some increments then went back online and generated some more increments. Hard to say what exactly happened.  I know those threads too can be out of our control and the system will create them as needed.

levieggertcru (approved):

PR #1509, opened 2023-03-03
GT-1951 Fix deeplink to cyoa

levieggertcru:
This PR fixes deeplinking to a ```cyoa``` tool with a pageId that is a category.  In iOS if the page is a category page, not (```intro``` or ```categories```) then we will build the intro, categories, and then category page in the renderer navigation stack.

Changes in this PR:
- Remove ```page: Int?``` from ```MobileContentPagesViewModel``` ```init()``` and replace with ```initialPage: MobileContentPagesPage?``` which encapsulates either a ```pageId: String``` or ```pageNumber: Int``` for the initial page to renderer.
- Change ```MobileContentPagesViewModel``` ```currentPage``` attribute to ```currentRenderedPageNumber```.
- Update renderer navigation to include page ids.

### DIFF
--- a/godtools/App/Services/Renderer/Views/MobileContent/Views/Pages/MobileContentPagesViewModel.swift
+++ b/godtools/App/Services/Renderer/Views/MobileContent/Views/Pages/MobileContentPagesViewModel.swift
@@ -286,8 +286,9 @@ extension MobileContentPagesViewModel {
                 
                 let introPageId: String = "intro"
                 let categoriesPageId: String = "categories"
+                let pageIsCategoryPage: Bool = value != categoriesPageId && value != introPageId
                 
-                guard value != categoriesPageId && value != introPageId else {
+                guard pageIsCategoryPage else {
                     return
                 }
                 
--- a/godtools/App/Services/Renderer/Views/MobileContent/Views/Pages/MobileContentPagesViewModel.swift
+++ b/godtools/App/Services/Renderer/Views/MobileContent/Views/Pages/MobileContentPagesViewModel.swift
@@ -25,7 +25,7 @@ class MobileContentPagesViewModel: NSObject {
     
     private(set) var renderer: CurrentValueSubject<MobileContentRenderer, Never>
     private(set) var currentPageRenderer: CurrentValueSubject<MobileContentPageRenderer, Never>
-    private(set) var currentPage: Int = 0
+    private(set) var currentRenderedPageNumber: Int = 0
     private(set) var highestPageNumberViewed: Int = 0
     private(set) var trainingTipsEnabled: Bool = false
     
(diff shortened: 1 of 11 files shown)


godtools/App/Services/Renderer/Views/MobileContent/Views/Pages/MobileContentPagesViewModel.swift
```diff
@@ -68,8 +64,379 @@ class MobileContentPagesViewModel: NSObject {
...
+        let shouldReloadPagesUI: Bool
+        
+        switch page {
+            
+        case .pageId(let value):
+                     
+            if isChooseYourOwnAdventure {
+                
+                let introPageId: String = "intro"
+                let categoriesPageId: String = "categories"
```
frett, line 288: you shouldn't hardcode these, this is specific to the "openers" tool only and wouldn't be applicable to other cyoa tools. There is a `page.parentPage` property that will return the configured page that should be added to the back stack. You should be able to recurse through the hierarchy to build the back stack

frett: https://github.com/CruGlobal/kotlin-mpp-godtools-tool-parser/blob/main/module/parser/src/commonMain/kotlin/org/cru/godtools/shared/tool/parser/model/page/Page.kt#L112

frett: this overall logic could look something like:

```
if isChooseYourOwnAdventure {
   var pages = [renderablePages.first(where: {$0.id == value})]
   
   while {
      let parent = pages[0].parentPage
      if (parent == null || pages.contains(parent)) break
      pages = parent + pages
   }
   
   pagesToRender = pages
}
```

levieggertcru: Thanks @frett !

levieggertcru: Hey @frett this is ready for a second look.  I was able to replace the entire pageId navigation with the logic of moving up the parent.

godtools/App/Services/Renderer/Views/MobileContent/Views/Pages/MobileContentPagesViewModel.swift
```diff
@@ -281,54 +281,30 @@ extension MobileContentPagesViewModel {
...
-                        break
-                    }
+                guard !pageModelsToRenderUpToPageToNavigateTo.contains(parentPage) else {
+                    break
                 }
                 
-                pagesToRender = pageModelsToRenderUpToPageToNavigateTo
-                navigateToPageModel = pageModelMatchingPageId
-                shouldReloadPagesUI = true
+                pageModelsToRenderUpToPageToNavigateTo.insert(parentPage, at: 0)
```
frett, line 301: because you are using the `insert` method here, can you use `let` instead of `var` for `pageModelsToRenderUpToPageToNavigateTo`?

levieggertcru: Swift doesn't allow the ```insert``` on ``immutable``` since ```Array``` is a value type.

Screenshot 2023-03-03 at 2 16 36 PM

frett: k, wasn't sure the semantics of it :)

frett (approved):

PR #1511, opened 2023-03-06
GT-1953 Fix categories list spacing issue

levieggertcru:
Because the categories page was inserted as the starting page on a deeplink it was using the content insets of 60 that is applied to the intro page.

### DIFF
--- a/godtools/App/Services/Renderer/Views/ChooseYourOwnAdventure/ViewFactory/ChooseYourOwnAdventurePageViewFactory.swift
+++ b/godtools/App/Services/Renderer/Views/ChooseYourOwnAdventure/ViewFactory/ChooseYourOwnAdventurePageViewFactory.swift
@@ -24,16 +24,18 @@ class ChooseYourOwnAdventurePageViewFactory: MobileContentPageViewFactoryType {
         
         if let contentPage = renderableModel as? ContentPage {
             
-            let page: Int = renderedPageContext.page
             let contentInsets: UIEdgeInsets
             let itemSpacing: CGFloat
             
-            if page == 0 {
+            let isIntroPage: Bool = contentPage.id == "intro"
+            let isCategoriesPage: Bool = contentPage.id == "categories"
+            
+            if isIntroPage {
                 
                 contentInsets = UIEdgeInsets(top: 0, left: 60, bottom: 0, right: 60)
                 itemSpacing = 28
             }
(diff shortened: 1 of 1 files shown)


godtools/App/Services/Renderer/Views/ChooseYourOwnAdventure/ViewFactory/ChooseYourOwnAdventurePageViewFactory.swift
```diff
@@ -24,16 +24,18 @@ class ChooseYourOwnAdventurePageViewFactory: MobileContentPageViewFactoryType {
...
             
-            let page: Int = renderedPageContext.page
             let contentInsets: UIEdgeInsets
             let itemSpacing: CGFloat
             
-            if page == 0 {
+            let isIntroPage: Bool = contentPage.id == "intro"
+            let isCategoriesPage: Bool = contentPage.id == "categories"
+            
+            if isIntroPage {
```
frett, line 33: We really shouldn't be doing custom rendering logic based on a page id. What was the intent behind the custom inset logic here?

levieggertcru: Hey @frett it's for the leading and trailing for the page content.  I remember I was basing it off Figma and the intro page had greater leading and trailing content insets.

Simulator Screen Shot - iPhone 8 - 2023-03-06 at 15 00 23


Simulator Screen Shot - iPhone 8 - 2023-03-06 at 15 00 26

frett: we'll have to revisit some of the layout capabilities of the renderer once we convert the mobile version to SwiftUI/Jetpack Compose

frett (approved):
This is fine for now, but really needs to be revisited in the future so you don't need to hardcode different behaviors based on an arbitrary page id.

On android we currently render the 2 pages the same and don't have the extra margin on the intro page

levieggertcru:
At least they're are set in ```ChooseYourOwnAdventurePageViewFactory``` which does narrow it down a little bit.  But yeah, it would be nice not to have to set those at all based on pageId.

PR #1518, opened 2023-03-17
GT-1847 delete auth token on logout

rachaelblue:
When a user logs out, delete the cached userId and auth token.  This won't account for the case where Okta logs a user out automatically-- not sure that there's a good way to know that.  But since Okta is being removed anyway, we may want to revisit this as other login methods are added.  It's also not strictly necessary that the auth token get deleted on logout, but preferable to leaving it there.


levieggertcru:
Hey @rachaelblue as for deleting the auth token on logout, I'm really not sure which is preferred.  It seems since the user is invoking a logout then maybe we'd want to clear that data for them.  It doesn't hurt anything.

Okta's access token will also expire and if we are unable to refresh that token then a new sign in with Okta would be required.  I don't think we need to worry about clearing the api auth token from the keychain in this case.  That token will most likely be expired as well and end up getting refreshed once it is needed.

levieggertcru (approved):

PR #1522, opened 2023-03-22
GT-1974 your activity UI

rachaelblue:


godtools/App/Features/Menu/Domain/UseCases/GetUserActivityStatsUseCase/GetUserActivityStatsUseCase.swift
```diff
@@ -0,0 +1,68 @@
...
+    
+    init(localizationServices: LocalizationServices) {
+        
+        self.localizationServices = localizationServices
+    }
+    
+    func getUserActivityStats(from userActivity: UserActivity) -> [UserActivityStatDomainModel] {
+        
+        let toolOpensStat = UserActivityStatDomainModel(
+            iconImageName: "tool-opens",
```
levieggertcru, line 25: Hey @rachaelblue for the image names can you place these in ```ImageCatalog.swift``` and pull from there.

godtools/App/Features/Menu/Domain/UseCases/GetUserActivityStatsUseCase/GetUserActivityStatsUseCase.swift
```diff
@@ -0,0 +1,68 @@
...
+            text: localizationServices.stringForMainBundle(key: "account.activity.toolOpens"),
```
levieggertcru, line 26: For the localizable string keys I remember you had mentioned creating a file for storing those.  Can you go ahead and create a file for storing localizable keys?  
I had created a ```Strings.swift```, but I don't think it's very clean to store all strings in there.  That file could get massive.   It would be more clean to create specific files for storing relevant strings.

godtools/App/Features/Menu/Domain/UseCases/GetUserActivityStatsUseCase/GetUserActivityStatsUseCase.swift
```diff
@@ -0,0 +1,68 @@
...
+            textColor: Color(red: 5 / 255, green: 105 / 255, blue: 155 / 255),
```
levieggertcru, line 27: For creating a Color with RGB can you create a ```Color extension``` for this. 
I had created similar logic ```getColorWithRGB``` in ```ColorPalette.swift```, but it would be more clean to put this in an extension as ColorPalette pertains to colors that are stored in the assets catalog.

godtools/App/Features/Menu/Presentation/Account/Subviews/AccountActivityView/Subviews/AccountActivityStatView.swift
```diff
@@ -0,0 +1,34 @@
...
+
+struct AccountActivityStatView: View {
+    
+    let stat: UserActivityStatDomainModel
+    
+    var body: some View {
+        
+        HStack {
+            
+            Image(stat.iconImageName)
```
levieggertcru, line 19: What are your thoughts on ensuring the Image size stays consistent with the Figma design by adding additional modifiers?

By default SwiftUI Image will size itself based on the asset which would cause problems if we add additional assets that are larger.

https://swiftwithmajid.com/2020/05/27/mastering-images-in-swiftui/

levieggertcru (changes requested):

godtools/App/Features/Menu/Presentation/Account/Subviews/AccountActivityView/Subviews/AccountActivityStatView.swift
rachaelblue: Yeah that's fair.  Ideally we'd get images that are sized correctly to begin with, but from past experience that's sometimes harder than it sounds 😄

levieggertcru (approved):

PR #1527, opened 2023-03-28
GT-1976 move common extensions into SharedAppleExtensions package

levieggertcru:


levieggertcru:
That one is pretty common to conflict :D

PR #1528, opened 2023-03-28
GT-1963 Update article downloading to include cache timestamp up to next hour in article url

levieggertcru:
Updated article downloading to include a timestamp to invalidate the json cache. 

Timestamp will be calculated based on the next hour ```timeIntervalSince1970```.  If the cache is ignored then a timestamp is not included in the URL.

Also increased json levels from 10 to 999.


godtools/App/Share/Data/ArticlesRepository/Repository/Downloader/ArticleAemDownloadGetCacheTimeInterval.swift
```diff
@@ -0,0 +1,49 @@
...
+            dateUpToNextHourComponents.minute = minutesUpToNextHour
+            dateUpToNextHourComponents.second = currentSeconds * -1
+            
+            guard let dateUpToNextHour = calendar.date(byAdding: dateUpToNextHourComponents, to: currentDate) else {
+                return nil
+            }
+            
+            let timeInterval: TimeInterval = dateUpToNextHour.timeIntervalSince1970
+            
+            return timeInterval
```
frett, line 43: no way to get seconds/milliseconds since epoch? if you have that you can just do a fairly simple `secs - (secs % secs_in_hour)` or an equivalent for milliseconds.

godtools/App/Share/Data/ArticlesRepository/Repository/Downloader/ArticleAemDownloadGetCacheTimeInterval.swift
```diff
@@ -0,0 +1,49 @@
...
+            
+        case .ignoreCache:
+            return nil
```
frett, line 46: counter-intuitively you should always return the current secs/millis since epoch if you are trying to ignore the cache

frett: just no rounding to an "hour" in that case

godtools/App/Share/Data/ArticlesRepository/Repository/Downloader/ArticleAemDownloadGetCacheTimeInterval.swift
frett: how I do the actual rounding on Android
https://github.com/CruGlobal/godtools-android/blob/develop/ui/article-aem-renderer/src/main/kotlin/org/cru/godtools/article/aem/service/AemArticleManager.kt#L346-L348

levieggertcru: Hmm, I didn't think of that.  That would be less code.  

Seconds since epoch would be ```Date().timeIntervalSince1970```.

https://developer.apple.com/documentation/foundation/nsdate/1407504-timeintervalsince1970

levieggertcru: Going to update.

godtools/App/Share/Data/ArticlesRepository/Repository/Downloader/ArticleAemDownloadGetCacheTimeInterval.swift
levieggertcru: Thanks @frett for reviewing, will update this.

godtools/App/Share/Data/ArticlesRepository/Repository/Downloader/ArticleAemDownloadGetCacheTimeInterval.swift
frett:
```swift
var timestamp = Date().timeIntervalSince1970
if (!force) {
    timestamp = timestamp - (timestamp % 3600)
}
return timestamp
```

frett (approved):
looks good, probably a good unit test candidate at some point in the future :)

PR #1539, opened 2023-04-10
GT-1965 Fix videos refreshing too soon

levieggertcru:
This PR adds ```MobileContentPageRendererPagesViewDataCache.swift``` which is a place we can store state for any views in a rendered page.  This will allow the state of a view to be restored say if navigating back to a previous page view.  Currently it only keeps the previous page and next page state data stored and will clear data as a user navigates.


rachaelblue (approved):
Video position in maintained when swiping forward/back a single page.

PR #1558, opened 2023-04-14
GT-1997 parse manifest async

levieggertcru:
Hey @frett was able to use async.  This PR includes:

- Replace IosManifestParser with ManifestParser to parse async.
- Needed to ensure parse async is called from main thread.
- Also on the getToolTranslationsFilesPublisher there is a closure downloadStarted which needed to be dispatched to the main thread in DownloadToolTranslationsFlow.swift.

### DIFF
--- a/godtools/App/Share/Data/TranslationsRepository/ManifestParser/TranslationManifestParser.swift
+++ b/godtools/App/Share/Data/TranslationsRepository/ManifestParser/TranslationManifestParser.swift
@@ -8,10 +8,11 @@
 
 import Foundation
 import GodToolsToolParser
+import Combine
 
 class TranslationManifestParser {
     
-    private let parser: IosManifestParser
+    private let parser: ManifestParser
     private let parserConfig: ParserConfig
     private let resourcesFileCache: ResourcesSHA256FileCache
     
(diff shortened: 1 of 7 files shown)


godtools/App/Share/Data/TranslationsRepository/ManifestParser/TranslationManifestParser.swift
```diff
@@ -39,32 +40,57 @@ class TranslationManifestParser {
...
+                return
             }
             
         case .failure(let error):
-            return .failure(error)
+            completion(.failure(error))
+            return
         }
         
-        let result: ParserResult = self.parser.parseManifestBlocking(fileName: manifestName, config: self.parserConfig)
```
frett, line 58: guess you weren't using the `parseManifestBlocking(fileName: manifestName)` variant that didn't require passing a config object.

frett (changes requested):
Don't merge this yet, it looks like the `openFile()` method is currently blocking and called on the same thread that initially calls `parseManifest()`

I'm going to do some research to see if I have access to background threads within the parsing logic.

frett (approved):

frett:
@levieggertcru I merged in the background thread parsing change to GodToolsShared. 0.8.3-SNAPSHOT will include this change.

frett (approved):

PR #1559, opened 2023-04-14
GT-1289 Fix mobile content renderer Link URL redirect

levieggertcru:
The MobileContentLinkView tap gesture was not receiving taps since a UIButton was part of the view hierarchy and taking the taps from the tap gesture in Base MobileContentView.

Fixed by removing the UIButton and replacing it with a UIView and UILabel.


rachaelblue (commented):
The link works, but it doesn't give the visual feedback to highlight on tap like a button would.  Not sure how easy it would be to add that in?

levieggertcru:
Hey @rachaelblue that's a good point.  I'm wondering now if I should update the Clickable elements to highlight like the UIButtons do.  Going to a little more investigation into this and the different clickable elements.

levieggertcru:
Hey @rachaelblue I looked into this some more and I noticed the MobileContentButtonView which is clickable is also setup similarly to the MobileContentLinkView because the button view supports multiple lines of text as well as optional leading and trailing images. ```https://github.com/CruGlobal/godtools-swift/blob/develop/godtools/App/Services/Renderer/Views/MobileContent/Views/ContentButton/MobileContentButtonView.swift#L85```
Since buttons aren't highlighting when tapped I think it's okay that links behave the same.  However, I think it would still be nice to have an option for clickable elements to highlight when tapped.  That could be something implemented when transitioning the renderer to SwiftUI.  I will create a ticket in the backlog for this.

PR #1568, opened 2023-04-21
GT-2005 social sign in view

rachaelblue:
Also included https://jira.cru.org/browse/GT-2005 here.  But just now realizing I didn't hook up the "Create account" button.  Is that supposed to go to the exact same page as login?


levieggertcru:
Hey @rachaelblue yep Create Account will go to the same page as Login.

rachaelblue:
@levieggertcru got the Create Account button routed.

levieggertcru:
Hey @rachaelblue, this is looking great.  I know you're out for some of this week so I did end up pushing some changes to this branch.  Below are some of the changes I added.

- Compressed the loginBackground asset via TinyPng which took the overall file size for those 3 assets from 750kb to 200kb. 

- Removed the white nav close icon that was added because we do have a nav_close icon which we can change to white by adjusting the tint color.  I also ended up moving this button to the navigation bar.

- Moved the social sign button inputs into the SocialSignInViewModel. One thing I noticed with moving logic into subview viewModels is that we end up having to inject dependencies further downward into subview viewModels and also delegate back up which tends to get messy and difficult to follow sometimes just because it creates more layers to follow the data flow through.  What are your thoughts on this?  

- Added SocialSignInAuthenticationType enum which tells us if the SocialSignInViewModel was launched from login or createAccount.  We will need this information for working with the mobile content api.  This was realized after a discussion today with Daniel.

- Updated ModalNavigationController status bar to be configurable via init().  This change came about when moving the SocialSignInView close button to the navigation bar.  I wanted to configure the status bar to be light color on the blue background.

levieggertcru:
Oh and one other change, I moved the SocialSignInButtonView background outside of the Button label.  I noticed when tapping the button the entire button high-lighted which looked a bit off.  By moving the background color outside of the Button label now just the title and icon highlight when tapped.

icon-title-highlight-only

entire-button-highlight

levieggertcru (approved):

rachaelblue:
Hey Levi, thanks for making these changes!  Looks great!  And the highlight on the button does look better with just the text-- good call.

As far as moving the button inputs into SocialSignInViewModel, I was going back and forth on that.  It seems like any way you do it, you end up injecting dependencies, setting delegates, or passing closures.  Sometimes I think allowing the subview to have some dependencies breaks things up a little so the main view model isn't so big.  But, the way use cases are working, view models are staying pretty slim anyway, which is great!  So I'm happy trying to keep logic like that in the main view model rather than the subs.

PR #1589, opened 2023-05-10
user counter unit tests

rachaelblue:
Just realized I never actually opened a PR for these 😆

Writing unit tests for the user counter logic was my KDA this year.  I went through some of the [Swift by Sundell articles](https://www.swiftbysundell.com/discover/unit-testing/) on unit testing, and basically wrote a test for each method in UserCountersRepository.  Also created a mock for the UserCountersAPI.

The biggest hurdle was getting Realm to work in-memory-- it unfortunately didn't work right out of the box just to set the cache type to in-memory, and I ended up with weird crashes.  Every time `RealmDatabase` created a new instance of the Realm object, even with the same config, it was creating a whole new in-memory Realm.  So for testing to work, we have to hold onto a single Realm instance and only use that for the duration of the test.


levieggertcru:
Hey @rachaelblue I'm glad you were able to write some tests, we desperately have a need for it in GodTools.  I'll read through the article you linked, that will be a good read for me.  I'll spend a little time with Realm as well.  Were the crashes you were seeing due to realm being accessed from an incorrect thread?   

realm-threading

https://www.mongodb.com/docs/realm/sdk/swift/crud/threading/

godtools/App/Share/Data/RealmDatabase/RealmDatabase.swift
```diff
@@ -15,18 +15,34 @@ class RealmDatabase {
     private let config: Realm.Configuration
     private let backgroundQueue: DispatchQueue = DispatchQueue(label: "realm.background_queue")
     
+    private lazy var isInMemory: Bool = {
+        return config.inMemoryIdentifier != nil
+    }()
+    private lazy var inMemRealm: Realm = {
```
levieggertcru, line 21: Hey Rachael,

I think I see why inMemRealm needs to be setup this way by exiting the ```background(async:``` method early.   Realm instances aren't thread safe and can only be accessed from the thread they're created from.  

The way inMemRealm is setup now is that it will be allocated on the thread it's first accessed from (since it's lazy) and then at that point forward it can only be accessed from that thread otherwise we will get the Realm accessed from incorrect thread crash.  I don't really want to constrain configurations with inMemory in this way.  I'd still like them to take advantage of creating Realm's on a background thread.

Instead, could you add some additional functionality to RealmDatabase.swift that would allow us to allocate a single threaded RealmDatabase instance from the thread it's created.  Then in the ```background(async:``` you can check for this and exit early with the Realm instance.  

This will allow for inMemory configurations to still take advantage of using the ```background(async``` and also give us the option to lock a RealmDatabase instance to a single thread from which it was allocated.

levieggertcru (changes requested):

levieggertcru (approved):

PR #1592, opened 2023-05-16
GT-1963 fix article loading

levieggertcru:
This PR makes the following changes to ArticleWebView:

- Adds method to ArticleWebViewModel that will stop a current WKWebView load request.
- Adds a view state (loadingArticle, errorMessage, viewingArticle) that the ArticleWebView can observe.
- Adds a timer which will change the view state to viewingArticle after 2 seconds of loading url request for the WKWebView. (In some devices the WKNavigationDelegate didFinish method wasn't firing).


rachaelblue (approved):
This is odd, I've never seen the article not load like Jill is saying 🤔

PR #1593, opened 2023-05-16
GT-2027 Integrate Apple Sign In

rachaelblue:
Hey @levieggertcru, I went ahead and created the provider interface extension for AppleAuthentication, but I know we had talked about getting the token renewal from MobileContentAPI.  Is that still the way it's going to work?  I'm assuming work needs to be done on that end still.

### DIFF
--- a/godtools/App/Share/Data/UserAuthentication/AuthenticationProviders/AppleAuthentication+AuthenticationProviderInterface.swift
+++ b/godtools/App/Share/Data/UserAuthentication/AuthenticationProviders/AppleAuthentication+AuthenticationProviderInterface.swift
@@ -1,0 +1,55 @@
+//
+//  AppleAuthentication+AuthenticationProviderInterface.swift
+//  godtools
+//
+//  Created by Rachael Skeath on 5/16/23.
+//  Copyright © 2023 Cru. All rights reserved.
+//
+
+import Foundation
+import UIKit
+import SocialAuthentication
+import Combine
+
+extension AppleAuthentication: AuthenticationProviderInterface {
+    func getPersistedAccessToken() -> AuthenticationProviderAccessToken? {
+        
+        // TODO: - access token renewal will come through MobileContentAPI?
+        
+        return nil
+    }
+    
+    func authenticatePublisher(presentingViewController: UIViewController) -> AnyPublisher<AuthenticationProviderAccessToken?, Error> {
+        
+        return authenticatePublisher()
+            .map { (response: AppleAuthenticationResponse) in
+                
+                guard let idToken = response.identityToken else {
+                    return nil
+                }
+                
+                // TODO: - include name once AppleAuth is updated
+                return AuthenticationProviderAccessToken.apple(
+                    idToken: idToken,
+                    givenName: "first name",
+                    familyName: "last name"
+                )
+            }
+            .eraseToAnyPublisher()
+    }
+    
+    func renewAccessTokenPublisher() -> AnyPublisher<AuthenticationProviderAccessToken, Error> {
+        
+        // TODO: - access token renewal will come through MobileContentAPI?
+        
+        return Just(AuthenticationProviderAccessToken.apple(idToken: "", givenName: "", familyName: ""))
+            .setFailureType(to: Error.self)
+            .eraseToAnyPublisher()
+    }
+    
+    func signOutPublisher() -> AnyPublisher<Void, Error> {
+        
+        return Just(()).setFailureType(to: Error.self)
+            .eraseToAnyPublisher()
+    }
+}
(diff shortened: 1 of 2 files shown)


levieggertcru:
Hey @rachaelblue that is correct, for token renewal some work will need to be done on the api side.   Unfortunately for now we won't be able to do any token refresh with Apple SignIn.

levieggertcru:
Hey @rachaelblue I merged the latest from ```develop``` into your branch.  I also added a new method ```func getAuthUserPublisher() -> AnyPublisher<AuthUserDomainModel?, Error>``` to your extension for the ```AuthenticationProviderInterface``` with a ```TODO```.

rachaelblue:
Thanks @levieggertcru!  I pulled in the family/given name change.  Is there anything else that should happen in this PR for now?  Or everything else will go into GT-2042 once the API is ready?

godtools/App/Share/Data/UserAuthentication/AuthenticationProviders/AppleAuthentication+AuthenticationProviderInterface.swift
```diff
@@ -0,0 +1,66 @@
...
+                    givenName: givenName,
+                    familyName: familyName
+                )
+            }
+            .eraseToAnyPublisher()
+    }
+    
+    func renewAccessTokenPublisher() -> AnyPublisher<AuthenticationProviderAccessToken, Error> {
+        
+        // TODO: - access token renewal will come through MobileContentAPI?
```
levieggertcru, line 46: Hey @rachaelblue for this let's instead Fail with an Error until refreshing is available.  This way the mobile-content-api won't be called and can fail immediately.

godtools/App/Share/Data/UserAuthentication/AuthenticationProviders/AppleAuthentication+AuthenticationProviderInterface.swift
```diff
@@ -0,0 +1,66 @@
...
+    
+    func signOutPublisher() -> AnyPublisher<Void, Error> {
+        
+        return Just(()).setFailureType(to: Error.self)
+            .eraseToAnyPublisher()
+    }
+    
+    func getAuthUserPublisher() -> AnyPublisher<AuthUserDomainModel?, Error> {
+        
+        // TODO: Should return an AuthUserDomainModel if user info can be obtained from AppleAuthentication. Implement in GT-2027. ~Levi
```
levieggertcru, line 61: For this we will need to return the AuthUserDomainModel.  I realize now we will need some additional work in social-authentication-ios so we can get user info for AuthUserDomainModel.

Google and Facebook SDKs will persist data for a current authenticated user such as access/refresh tokens, user info, etc.  See GoogleSignIn Framework Reference for ```GIDGoogleUser``` https://developers.google.com/identity/sign-in/ios/reference/Classes/GIDGoogleUser

It would be nice if we could do something similar for AppleAuthentication.swift this way we can fetch a current user and add the email, first name, and last name to the AuthUserDomainModel.

levieggertcru (changes requested):
Hey @rachaelblue as far as token refresh this is far as we can go for now.  We will still need the user email, firstName, and lastName for AuthUserDomainModel.  See comments in the change request.

godtools/App/Share/Data/UserAuthentication/AuthenticationProviders/AppleAuthentication+AuthenticationProviderInterface.swift
```diff
@@ -0,0 +1,67 @@
...
+        // TODO: - implement in GT-2042
+        
+        let error: Error = NSError.errorWithDescription(description: "Access token renewal not yet implemented")
+
+        return Fail(error: error)
+            .eraseToAnyPublisher()
+    }
+    
+    func signOutPublisher() -> AnyPublisher<Void, Error> {
+
```
levieggertcru, line 55: Hey @rachaelblue just noticed that signOut will also need to be called on AppleAuthentication.

rachaelblue: good catch!

godtools/App/Share/Data/UserAuthentication/AuthenticationProviders/AppleAuthentication+AuthenticationProviderInterface.swift
```diff
@@ -0,0 +1,81 @@
...
+    
+    func authenticatePublisher(presentingViewController: UIViewController) -> AnyPublisher<AuthenticationProviderAccessToken?, Error> {
+        
+        return authenticatePublisher()
+            .map { (response: AppleAuthenticationResponse) in
+                
+                guard
+                    let idToken = response.identityToken,
+                    let givenName = response.fullName?.givenName,
+                    let familyName = response.fullName?.familyName
```
levieggertcru, line 30: Hey @rachaelblue just thought of something here as well.  I'm wondering if we ever run into a situation where we get an identity token, but given name or family name is null.  I'm good with passing this for now, but will need to make a ticket and do some research here and some testing as well to see if we can pass a null givenName or familyName to the mobile-content-api auth.

levieggertcru (approved):

PR #1597, opened 2023-05-17
GT-2029 MobileContentAPI create_user

rachaelblue:


godtools/App/Share/Data/MobileContentApiAuthSession/MobileContentApiAuthSession.swift
```diff
@@ -76,17 +76,15 @@ class MobileContentApiAuthSession {
         }
     }
     
-    private func fetchRemoteAuthToken() -> AnyPublisher<String, URLResponseError> {
-        
-        // TODO: The internals of this method will need to be updated since we no longer support okta.  See GT-2025.  ~Levi
-        
+    private func fetchRemoteAuthToken(createUser: Bool = false) -> AnyPublisher<String, URLResponseError> {
```
rachaelblue, line 79: I'm anticipating some of this may need to be restructured when we start calling `/auth` independently rather than as needed when making other authenticated requests.  Just giving `create_user` a default value here for the moment til it actually gets hooked up.

levieggertcru: Hey @rachaelblue createUser=false works here.  The MobileContentApiAuthSession assumes oidc identity / access tokens already exist and just attempts to refresh them once they've expired which in turn makes a new /auth call to mobile-content-api.

levieggertcru (approved):

PR #1607, opened 2023-05-24
GT-2010 Add UI and UseCase for deleting an account

levieggertcru:
This PR adds the UI flow and use case to delete a user account.  The delete account use case will call mobile-content-api DELETE /users/me, will sign out of authentication providers revoking tokens, and delete the mobile-content-api auth token from keychain.

Upon successful completion the user is presented with an alert that account deletion was successful.

If an error occurs the user is presented with an alert of the error description.
 
1-delete-account

2-confirm-delete-account

3-deleting-account-progress

4-account-deleted-confirmation


godtools/App/Features/Menu/Presentation/DeleteAccountProgress/DeleteAccountProgressViewModel.swift
```diff
@@ -0,0 +1,61 @@
...
+                
+                DispatchQueue.main.async {
+                    
+                    if let deleteAccountError = deleteAccountError {
+                        self?.flowDelegate?.navigate(step: .didFinishAccountDeletionWithErrorFromDeleteAccountProgress(error: deleteAccountError))
+                    }
+                    else {
+                        self?.flowDelegate?.navigate(step: .didFinishAccountDeletionWithSuccessFromDeleteAccountProgress)
+                    }
+                }
```
rachaelblue, line 54: Shouldn't we already be on main here?

godtools/App/Features/Menu/Domain/UseCases/DeleteAccountUseCase/DeleteAccountUseCase.swift
```diff
@@ -0,0 +1,50 @@
...
+                    .mapError { (error: Error) in
+                        return .otherError(error: error)
+                    }
+                    .eraseToAnyPublisher()
+            })
+            .flatMap({ (void: Void) -> AnyPublisher<Void, Never> in
+                        
+                self.mobileContentAuthTokenRepository.deleteCachedAuthToken()
+                
+                return SetTimeout.setTimeoutPublisher(seconds: 1)
```
rachaelblue, line 40: What does this timeout do?  Make sure the auth token gets removed from the cache before returning?

rachaelblue (commented):
I'm wondering if the "delete account" option shouldn't show up if you're not logged in?

godtools/App/Features/Menu/Presentation/DeleteAccountProgress/DeleteAccountProgressViewModel.swift
levieggertcru: Hey @rachaelblue good catch!  For some reason I did have a thread crash here earlier when implementing, but maybe I didn't have the ```receiveOnMain()``` in place.  I can't exactly remember now, but after running the app a few times it is calling on the mainThread so will remove the DispatchQueue.main.async.

godtools/App/Features/Menu/Domain/UseCases/DeleteAccountUseCase/DeleteAccountUseCase.swift
levieggertcru: Ah this was because I would like the UI to display a little longer here in the event the api request is really fast.  Not sure about having this in the UseCase though.  I'm going to look to moving this delay to the Flow with a combination of elapsed time.

levieggertcru: Hmm, I might be able to use debounce or delay.

https://developer.apple.com/documentation/combine/publisher/debounce(for:scheduler:options:)

https://developer.apple.com/documentation/combine/publisher/delay(for:tolerance:scheduler:options:)

godtools/App/Features/Menu/Presentation/DeleteAccountProgress/DeleteAccountProgressViewModel.swift
levieggertcru: Hey @rachaelblue right! I don't think the delete account option should show if you're not signed in.  And I just realized too that in order for a user to actually delete their account they will need a valid mobile-content-api access token.  Going to create a separate ticket to hide the Delete Account menu option.

godtools/App/Features/Menu/Presentation/DeleteAccountProgress/DeleteAccountProgressViewModel.swift
```diff
@@ -27,9 +28,17 @@ class DeleteAccountProgressViewModel: ObservableObject {
...
+        deleteAccount()
+    }
+    
+    private func deleteAccount() {
+        
+        let startDeleteAccountTime = Date()
+        
         deleteAccountUseCase.deleteAccountPublisher()
             .receiveOnMain()
+            .delay(for: .seconds(getRemainingSecondsToDisplayDeleteAccountProgress(startTime: startDeleteAccountTime)), scheduler: DispatchQueue.main)
```
rachaelblue, line 41: Good to know about this!

rachaelblue (approved):

PR #1612, opened 2023-05-30
GT Cache mobile-content-api auth token expiration date

levieggertcru:
Updated the MobileContentAuthTokenCache to include the expiration date in the event we need to check if the mobile-content-api auth token has expired.
- Cache the auth token expiration date.
- Add isExpired to check if the auth token has expired.

### DIFF
--- a/godtools/App/Share/Data/MobileContentAuthTokenRepository/Cache/MobileContentAuthTokenCache.swift
+++ b/godtools/App/Share/Data/MobileContentAuthTokenRepository/Cache/MobileContentAuthTokenCache.swift
@@ -10,38 +10,13 @@ import Foundation
 
 class MobileContentAuthTokenCache {
     
-    enum CacheKey: String {
-        
-        case expirationDate = "expirationDate"
-        
-        func getUserKey(userId: String) -> String {
-            return "MobileContentAuthTokenCache.\(userId).\(rawValue)"
-        }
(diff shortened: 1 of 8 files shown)


godtools/App/Share/Data/MobileContentAuthTokenRepository/Cache/MobileContentAuthTokenCache.swift
```diff
@@ -10,11 +10,38 @@ import Foundation
...
+        case expirationDate = "expirationDate"
+        
+        func getUserKey(userId: String) -> String {
+            return "MobileContentAuthTokenCache.\(userId).\(rawValue)"
+        }
+    }
+    
+    private let sharedUserDefaults: SharedUserDefaultsCache
+    
     let keychainAccessor: MobileContentAuthTokenKeychainAccessor
```
rachaelblue, line 24: Would it be worth it to separate out the user default logic since the keychain logic is separate already too?

levieggertcru: Hey @rachaelblue it would be cleaner.  I'm going to do that and will also look into using Realm rather than UserDefaults.

rachaelblue (approved):

PR #1614, opened 2023-05-30
GT-2041 send auth on login with create_user

rachaelblue:


godtools/App/Features/Menu/Presentation/SocialSignIn/SocialSignInViewModel.swift
```diff
@@ -60,26 +60,30 @@ class SocialSignInViewModel: ObservableObject {
...
+            
+            switch subscriberCompletion {
+            case .finished:
+                authenticationError = nil
+            case .failure(let error):
+                authenticationError = error
             }
-            .store(in: &cancellables)
+            
+            self?.handleAuthenticationCompleted(error: authenticationError)
```
levieggertcru, line 81: Hey @rachaelblue looked into this crash a little more and I think I see what's going on, I'm just not entirely sure on a fix for it.

Our custom ```receiveOnMain()``` will successfully switch to the main thread for Self.Output using the flatMap.  However, since this crash happens from a thrown Error, flatMap won't get called in our custom ```receiveOnMain()```. 

```
public extension Publisher {
    
    func receiveOnMain() -> AnyPublisher<Self.Output, Self.Failure> {
        
        return self.flatMap({ value -> AnyPublisher<Self.Output, Self.Failure> in
            
            if Thread.isMainThread {
                
                return Just(value)
                    .setFailureType(to: Self.Failure.self)
                    .eraseToAnyPublisher()
                
            } else {
                
                return Just(value)
                    .setFailureType(to: Self.Failure.self)
                    .eraseToAnyPublisher()
                
            }
            
        }).eraseToAnyPublisher()
    }
}
```

levieggertcru: Went ahead and created this task here -> https://jira.cru.org/browse/GT-2052.

levieggertcru (approved):

PR #1617, opened 2023-06-01
GT Include serverClientId for Google Sign-In

levieggertcru:
This PR contains the following changes:

- Adding AppEnvironment enum for staging and production.
- In AppBuild.swift set environment based on the build configuration.
- AppConfig.swift will now configure values based on the environment rather than configuration to reduce any inconsistencies between staging and production.
- Added staging and production clientId and serverClientId for Google SignIn.
- Added staging reverse client id for Google SignIn.
- Added AppleAuthentication as a provider to UserAuthentication.swift.


rachaelblue (approved):
Fixed the Google sign in error!

PR #1622, opened 2023-06-05
GT-2050 refactor authentication provider response type

levieggertcru:
This PR is some refactor to UserAuthentication as a result of the comments in (https://github.com/CruGlobal/godtools-swift/pull/1588).

Each of our authentication providers implement AuthenticationProviderInterface which returns a type AuthenticationProviderAccessToken.  AuthenticationProviderAccessToken didn't make much sense and has been changed to AuthenticationProviderResponse which will encapsulate an identity token, access token, refresh token, and user profile information.

There is now an extension on AuthenticationProviderResponse (+ MobileContentAuthToken) which will return a MobileContentAuthToken type for use against the mobile-content-api /auth endpoint.


godtools/App/Share/Data/MobileContentAuthTokenRepository/Api/MobileContentAuthTokenAPI.swift
```diff
@@ -22,13 +22,13 @@ class MobileContentAuthTokenAPI {
         self.baseURL = config.mobileContentApiBaseUrl
     }
     
-    private func getAuthTokenRequest(providerAccessToken: AuthenticationProviderAccessToken, createUser: Bool) -> URLRequest {
+    private func getAuthTokenRequest(authToken: MobileContentAuthToken, createUser: Bool) -> URLRequest {
```
rachaelblue, line 25: I'm a little confused about the naming now, because the `MobileContentAuthToken` struct isn't actually the auth token, it's the properties that are needed to fetch the auth token, right?

levieggertcru: Hey @rachaelblue that definitely is more confusing now.  I'm thinking it should be something like ```providerToken``` instead of ```providerAccessToken```.

levieggertcru: Right it's the properties needed to fetch the mobile-content-api auth token.

levieggertcru: And those properties come from the OIDC token (identity token, access token) etc.

levieggertcru: Maybe the struct should be ```MobileContentProviderToken``` ?

levieggertcru:
Hey @rachaelblue just pushed a commit.  Let me know your thoughts on this name change.

Changed ```MobileContentAuthToken``` to ```MobileContentAuthProviderToken```.

Method argument is now ```(providerToken: MobileContentAuthProviderToken```.

rachaelblue (approved):
Hey @levieggertcru, that name change makes a lot more sense to me.  Thanks!

PR #1632, opened 2023-06-13
GT-2042 apple signin token refresh

rachaelblue:


godtools/App/Share/Data/UserAuthentication/AuthenticationProvider/Response/AuthenticationProviderResponse+MobileContentAuthProviderToken.swift
```diff
@@ -21,11 +21,18 @@ extension AuthenticationProviderResponse {
...
             
-            guard let idToken = self.idToken, !idToken.isEmpty else {
-                return .failure(NSError.errorWithDescription(description: "Missing apple idToken."))
+            if let authCode = self.appleSignInAuthorizationCode, authCode.isEmpty == false {
+                
+                return .success(.appleAuth(authCode: authCode, givenName: profile.givenName, familyName: profile.familyName))
+                
+            } else if let refreshToken = self.refreshToken, refreshToken.isEmpty == false {
+                
+                return .success(.appleRefresh(refreshToken: refreshToken))
```
levieggertcru, line 30: Hey @rachaelblue just realized something that this refreshToken portion will never evaluate to anything since the Apple refreshToken isn't part of the provider.

godtools/App/Share/Data/UserAuthentication/UserAuthentication.swift
```diff
@@ -97,11 +115,56 @@ class UserAuthentication {
...
+                let persistedAppleRefreshToken = self.mobileContentAuthTokenRepository.getCachedAuthTokenModel()?.appleRefreshToken
+                
+                let appleAuthProviderResponse = AuthenticationProviderResponse(
+                    accessToken: nil,
+                    appleSignInAuthorizationCode: nil,
+                    idToken: nil,
+                    profile: authProviderProfile,
+                    providerType: .apple,
+                    refreshToken: persistedAppleRefreshToken
+                )
```
rachaelblue, line 159: Hey @levieggertcru, I used `AuthenticationProviderResponse` inside `UserAuthentication` when I fetch the refresh token, so the logic in `AuthenticationProviderResponse+MobileContentAuthProviderToken` will get evaluated.  Would you still prefer to move that logic and avoid doing it this way?

levieggertcru: Hey @rachaelblue I see what you were able to do.  I didn't see that.

But yeah let's still move that logic since we're doing all the mobile-content-api related things in UserAuthentication due to having the MobileContentAuthTokenRepository reference in there now.

rachaelblue: Fixed!

levieggertcru: Thanks @rachaelblue !

levieggertcru (approved):

PR #1696, opened 2023-08-21
GT create user flag

levieggertcru:


levieggertcru:
Remove create_user flag from apple refresh.

PR #1714, opened 2023-09-05
GT Adding SwiftUIPreview RealmDatabase and AppConfigInterface

levieggertcru:
This PR contains some prep work for fixing SwiftUI preview crashes.  The preview still crashes and more work will need to be completed following this PR.

Changes in this PR:

- Added SwiftUIPreviewDatabase.swift which will create an in memory database with some realm objects for SwiftUI Preview.
- Add AppConfigInterface which SwiftUI preview may require it's own config.
- Change AppConfig computed properties to get methods.
- Add enum AppBuildConfiguration.swift.  Removed from AppBuild.swift.
- Moved some classes out of AppDiContainer and into AppDataLayerDependencies.
- Remove protocol FirebaseInAppMessagingType.


godtools/App/Preview/Data/RealmDatabase/SwiftUIPreviewDatabaseConfiguration.swift
```diff
@@ -0,0 +1,20 @@
...
+//
+
+import Foundation
+
+class SwiftUIPreviewDatabaseConfiguration: RealmDatabaseConfiguration {
+    
+    init() {
+        
+        super.init(
+            cacheType: .inMemory(identifier: "godtools_swiftui_preview_database"),
```
rachaelblue, line 16: So do previews require the database to be in memory?  I know previews were kind of working at some point, so just curious what changed.

levieggertcru: Hey @rachaelblue it shouldn't, but there are instances those objects may not exist in the database yet since those are populated on app launch.  

There's a couple other issues going on with previews crashing right now.  One in Firebase Analytics when attempting to track an event.  I also noticed that when the preview is compiled it actually triggers UIApplicationDidBecomeActive in the AppFlow which then causes a crash in the shared parser parse translation manifest async method.

PR #1715, opened 2023-09-05
GT-2098 tools filter UI

rachaelblue:
- built out filter button UI on Tools
- created new filter search/selection page (language/category shares the same view and has separate view model subclasses)
- set up navigation with app flow
- haven't hooked up any data yet, so leaving some of the existing category logic in place as some of it might get reused


godtools/App/Features/Dashboard/Presentation/Tools/Presentation/ToolsView.swift
```diff
@@ -100,6 +98,43 @@ struct ToolsView: View {
             viewModel.pageViewed()
         }
     }
+    
+    @ViewBuilder func filterSection(width: CGFloat) -> some View {
```
levieggertcru, line 102: Hey @rachaelblue instead of using @ViewBuilder here I would like a View to encapsulate this instead.  I think it's cleaner and more readable to use a separate View file and if we started adding additional ViewBuilder methods for subviews it could bloat this file.

godtools/App/Share/SwiftUI Views/SearchBar/SearchBar.swift
```diff
@@ -0,0 +1,70 @@
...
+//
+
+import SwiftUI
+
+// adapted from https://www.appcoda.com/swiftui-search-bar/
+
+struct SearchBar: View {
+    
+    @Binding var searchText: String
+    let localizationServices: LocalizationServices
```
levieggertcru, line 16: Here we don't want to View referencing a DataLayer class.  The View shouldn't know anything about the DataLayer.  That would violate our clean architecture.

godtools/App/Share/SwiftUI Views/SearchBar/SearchBar.swift
```diff
@@ -0,0 +1,70 @@
...
+                                .foregroundColor(.gray)
+                                .padding(.trailing, 10)
+                        }
+                    }
+                }
+            )
+            
+            if isEditing {
+                Button(action: {
+                    DispatchQueue.main.async {
```
levieggertcru, line 58: It seemed strange to me that DispatchQueue.main.async is needed here, but it also seems that the state won't trigger without it?  Is that what you found?

godtools/App/Share/SwiftUI Views/SearchBar/SearchBar.swift
```diff
@@ -0,0 +1,70 @@
...
+                        self.isEditing = false
+                        self.searchText = ""
+                    }
+                    
+                    UIApplication.shared.sendAction(#selector(UIResponder.resignFirstResponder), to: nil, from: nil, for: nil)
```
levieggertcru, line 63: We haven't had much experience using Keyboard with SwiftUI, too bad we have to use UIApplication this way to resignFirstResponder on the TextField.  Seems a bit wonky, but probably the only way at the moment.

godtools/App/Features/Dashboard/Presentation/Tools/Presentation/Subviews/ToolFilterSelectionView/ViewModel/ToolFilterSelectionViewModel.swift
```diff
@@ -0,0 +1,43 @@
...
+//  Created by Rachael Skeath on 8/25/23.
+//  Copyright © 2023 Cru. All rights reserved.
+//
+
+import Foundation
+import Combine
+
+class ToolFilterSelectionViewModel: ObservableObject {
+    
+    var selectedCategory: ToolCategoryDomainModel? = nil
```
levieggertcru, line 14: Should these domain model references be immutable or at least private?  Unless they will need to be mutated by the subclasses?

levieggertcru (changes requested):
Hey @rachaelblue the UI is looking good!  Left a couple change requests.
There's also a folder structure change I'd like as well with the location of the TooFilterSelectionView.swift, I have a separate branch created for that structure change I'd like to go through on a quick call.

godtools/App/Features/Dashboard/Presentation/Tools/Presentation/ToolsView.swift
rachaelblue: That makes sense.  I was thinking you didn't want so many layers of subviews, but were you more concerned about layers of view models?

godtools/App/Share/SwiftUI Views/SearchBar/SearchBar.swift
rachaelblue: I hated having a whole viewModel just to translate "cancel", when this should be a built-in SwiftUI component anyway (there is one in SwiftUI, but it's not customizable), but I realize it does violate clean architecture.  Would you think I should move all the vars into a viewModel, or just localizationServices?

godtools/App/Share/SwiftUI Views/SearchBar/SearchBar.swift
rachaelblue: I *think* it's something about `resignFirstResponder`.  When I remove that UIApplication call, the `cancel` button at least works properly (but the keyboard doesn't get dismissed).  I can look into this a little more, maybe there's a different way this is supposed to happen in SwiftUI.

godtools/App/Features/Dashboard/Presentation/Tools/Presentation/Subviews/ToolFilterSelectionView/ViewModel/ToolFilterSelectionViewModel.swift
rachaelblue: I ended up removing these in favor of a `ToolFilterSelection` current value publisher in my follow-up branch.  But yes, that's a huge downside of using inheritance when we really just want a protocol-- we end up with a lot of public stuff.  If there's a better way to reuse views with different viewModel types, I'm all for it, just not sure what it is 🤔

rachaelblue: That makes sense, I hadn't thought of having the interfaces injected into a single viewModel.  I imagine that could still be tricky because there's likely specific behavioral differences that could be difficult to model in an interface.  I'll keep going as is, but it wouldn't be hard to make two separate views if that ends up making more sense.  I know we've had similar conversations to this before, I think I'm so used to protocols that I really want the baseViewModel thing to work 😅

rachaelblue:
Hey @levieggertcru, I made the folder structure change, moved the filter section into a view, and added a SearchBarViewModel.  I didn't touch the domain model references since I ended up removing them in the next branch and using a publisher instead.

godtools/App/Share/SwiftUI Views/SearchBar/SearchBarViewModel.swift
```diff
@@ -0,0 +1,22 @@
...
+//  Created by Rachael Skeath on 9/14/23.
+//  Copyright © 2023 Cru. All rights reserved.
+//
+
+import Foundation
+import Combine
+
+class SearchBarViewModel: ObservableObject {
+    
+    var searchTextPublisher: CurrentValueSubject<String, Never>
```
levieggertcru, line 14: This variable should be immutable.  Immutability is always preferred.

godtools/App/Features/Dashboard/Presentation/ToolFilterSelectionView/ViewModel/ToolFilterSelectionViewModel.swift
```diff
@@ -0,0 +1,53 @@
...
+//
+
+import Foundation
+import Combine
+
+class ToolFilterSelectionViewModel: ObservableObject {
+    
+    var selectedCategory: ToolCategoryDomainModel? = nil
+    var selectedLanguage: LanguageDomainModel? = nil
+    var searchTextPublisher: CurrentValueSubject<String, Never> = CurrentValueSubject("")
```
levieggertcru, line 16: This variable too should be immutable.  It will never need to be mutated.

godtools/App/Features/Dashboard/Presentation/Tools/Presentation/Subviews/ToolFilterSelectionView/ViewModel/ToolFilterSelectionViewModel.swift
levieggertcru: Yeah interfaces are the preferred way.  And it can get a bit tricky too with interfaces.  But yeah keep at it, will be interesting to see how it goes.

levieggertcru (approved):

PR #1724, opened 2023-09-15
GT-2102 fix tab view bug for iOS 16

levieggertcru:
This PR fixes iOS 16 bugs with the TabView when in a right to left language.  Ended up having to force a TabView and any accompanying TabBarView or PageControl that controls the TabView to always be in a left to right environment.  Then manually adjust them for right to left languages. 

As of this PR only left to right languages will be supported on PagedView.  Once we start implementing the new App Language feature these will be able to support right to left.  Will create a PR pointing to the app-languages-feature to address that.  

- Add PagedView which wraps a TabView and forced the environment to left to right and manually adjust for right to left languages.
- Updated PageControl to always be in a left to right environment and manually adjust for right to left languages.
- Updated DashboardTabBarView to always be in a left to right environment and manually adjust for right to left languages.
- Updated any remaining views using a TabView to use PagedView instead.
- Removed all flipsForRightToLeftLayoutDirection modifiers.


rachaelblue (commented):
Hey @levieggertcru, what language is good to test this in?  I tried changing the system language to Arabic, but I can't get past the splash screen.  Is that normal?

levieggertcru:
Hey @rachaelblue any language at the moment.  It should behave the same (Left to Right Layout) no matter the language.  That's strange, you should be able to get past the splash screen.  Let me check Arabic and see if I can reproduce.

levieggertcru:
Also I will be creating a follow up PR after this one is merged to address right to left languages.  But that will be branched from the app-languages-feature.

levieggertcru:
I got past the splash screen running Arabic, iPhone 12 Pro 16.2.  

What are you seeing when you get stuck on the splash?  Is it just the logo you see and nothing else?

rachaelblue:
@levieggertcru well now I can't reproduce it, which is making me think maybe it was just a connection issue.  It happened when I had the app uninstalled, changed the system language from English to Arabic, and then installed fresh (just got stuck on the splash screen).  But I wasn't having that problem in English, so who knows.  At any rate, I haven't been able to reproduce it today.

rachaelblue (approved):

PR #1746, opened 2023-09-24
GT-2115 filter tools

rachaelblue:


godtools/App/Features/Dashboard/Domain/GetAllToolsUseCase/GetAllToolsUseCase.swift
```diff
@@ -14,32 +14,68 @@ class GetAllToolsUseCase {
     private let getToolUseCase: GetToolUseCase
     private let resourcesRepository: ResourcesRepository
     
+    private var allToolsInMem: [ToolDomainModel]?
```
rachaelblue, line 17: Hey @levieggertcru, wanted to ask what you thought about this.  I was running into the issue that the language filter selection page had a slight delay in presenting due to the logic required to count how many tools occur in each language.  Counting "all tools" is computationally expensive, since it requires sorting out the variants/default variants etc before actually counting.  So, to get around making the allTools request every time, I'm just doing it once and storing them here, and then filtering this array.  Thoughts?  I realize that if this "optimization" gets used anywhere else, we run the risk of presenting old data since this array currently never gets updated (not really an issue here since the page gets created fresh every time it opens, but could be risky elsewhere).

levieggertcru: Hey @rachaelblue I see, is it this method in RealmResourcesCache ```func getAllTools(sorted: Bool, category: String? = nil, languageId: String? = nil) -> [ResourceModel]``` that's heavy on the computations?

Yeah I'd say just store that in memory for now.  Maybe we can re-visit tweaking that once everything is in and tested.  I'm wondering if a more efficient way would be to store a list of ResourceModels on our LanguageModel type? Currently each ResourceModel does point to a list of language ids.  But the LanguageModel doesn't point back.

levieggertcru: Hey @rachaelblue and right there is the possibility of that inMemory list not getting updated.

rachaelblue: Yes, that's the method that does all the complicated logic for metatools/variants.  It seems like it would be more efficient if we had the list of resources attached to the LanguageModel, although I'm thinking we'd still need to do the variant sorting on each individual list of ResourceModels.  Otherwise the tool count may still be off, if I understand it correctly.

levieggertcru: Let's stick with the inMemory for now.  I think that's the quickest best solution.  We can create a separate task to come back to this and fine tune it.  And right I don't think variants are included in that list since we do filter those out in the All Tools UI.

levieggertcru (approved):

levieggertcru (approved):

PR #1751, opened 2023-09-27
GT Add simple test to test sorting on app languages list by english language

levieggertcru:


levieggertcru:
Pushed some more tests.  Curious what the code diff will do now.

PR #1756, opened 2023-10-02
GT-2122 Add more app languages 

levieggertcru:
Hey @rachaelblue I had the original list commented out and was only including ar and en.  Updated to include more languages.  Might be nice to see more languages as you implement that list.


levieggertcru:
Feel free to merge this in if you need to.

rachaelblue (approved):
Thanks @levieggertcru!  Was playing with just the two languages and it's pretty cool how everything swaps sides with Arabic.

PR #1763, opened 2023-10-06
GT-2122 bind search text in both directions

levieggertcru:
Hey @rachaelblue here's an example of the two directional binding.  I need to add the publisher back to the SearchBar, I think this will break Tool Filtering Search.


levieggertcru:
Okay added that publisher.send back to the SearchBar.  Eventually that can go away when implementing the 2 way binding with tool filtering search.

PR #1766, opened 2023-10-09
GT-2090 Plural Activity Counter Strings

rachaelblue:


godtools/App/Features/Menu/Domain/UseCases/GetUserActivityStatsUseCase/GetUserActivityStatsUseCase.swift
```diff
@@ -23,46 +23,56 @@ class GetUserActivityStatsUseCase {
...
     }
+    
+    private func getUserActivityText(stringKey: String, activityCount: Int32) -> String {
+        
+        let formatString = localizationServices.stringForSystemElseEnglish(
+            key: stringKey,
+            fileType: .stringsdict
+        )
+        
+        return String.localizedStringWithFormat(formatString, activityCount)
```
levieggertcru, line 76: Hey @rachaelblue I'm good with leaving this PR as is for now since we're using LocalizationServices in the Account presentation.

Eventually we'll need to update this entire view interface strings to be based on the app language which will most likely effect the ```String.localizedStringWithFormat``` since that defaults to the current Locale (https://developer.apple.com/documentation/foundation/nsstring/1497301-localizedstringwithformat) and instead will need to use (https://developer.apple.com/documentation/foundation/nsstring/1497317-initwithformat).

rachaelblue: Hey @levieggertcru, yeah I was wondering when that should happen.  Didn't know if we had a ticket to rip out LocalizationServices altogether or not.

rachaelblue: Also I merged dev back into this branch, so conflicts are resolved.

levieggertcru: Okay nice I'm going to approve this PR.  I'd say go ahead and create a Jira task for updating the account activity to translate based on the app language.

levieggertcru (approved):

godtools/App/Features/Menu/Domain/UseCases/GetUserActivityStatsUseCase/GetUserActivityStatsUseCase.swift
rachaelblue: Sounds good, made a ticket here: https://jira.cru.org/browse/GT-2159

PR #1773, opened 2023-10-12
GT-2157 consolidate app languages

levieggertcru:


levieggertcru:
Hey @rachaelblue I'm going to go ahead and merge this PR, I rolled back a lot of the UseCases I had created in the app langauge feature per our discussion yesterday.  This is a high level of some of the changes:
- Removed GetAppLanguageNameUseCase and GetAppLanguageNameInAppLanguageUseCase.
- Removed AppLanguageNameDomainModel and use a String type instead.
- Pushed the logic for building the app languages list items into the repository that implements the domain interface.
- Removed the observe methods on the use cases and now back to having a single method.  We just have to be sure to keep any publishers that trigger changes at the top of the publisher chain.
-  Added a use case for fetching the languages settings interface strings.

PR #1794, opened 2023-10-26
GT-2123 confirm app language UI

rachaelblue:
It turns out NSAttributedString isn't supported in SwiftUI, they have their own version AttributedString, but that requires iOS 15+.  And it's odd there's not really a good workaround.  So for now, in iOS 15 or above you'll see the language name in blue, and in 14 you won't.  I'm gonna see what Jill thinks about it, but I wasn't sure it's worth a lot of effort.


levieggertcru:
Morning @rachaelblue let's see what Jill says.  I don't think it's worth a ton of effort either especially with iOS 14 most likely being a small percentage of our users.

godtools/App/Features/AppLanguage/Domain/UseCases/GetConfirmAppLanguageInterfaceStringsUseCase.swift
```diff
@@ -0,0 +1,91 @@
...
+    
+    private func getAttributedMessageStringPublisher(selectedLanguage: AppLanguageCodeDomainModel, translatedInLanguage: AppLanguageCodeDomainModel) -> AnyPublisher<NSAttributedString, Never> {
+        
+        return getInterfaceStringRepositoryInterface.getStringPublisher(languageCode: translatedInLanguage, stringId: "languageSettings.confirmAppLanguage.message")
+            .flatMap { formatString in
+                
+                let languageName = self.localeLanguageName.getDisplayName(forLanguageCode: selectedLanguage, translatedInLanguageCode: translatedInLanguage) ?? ""
+                let languageNameAttributed = NSAttributedString(
+                    string: languageName,
+                    attributes: [NSAttributedString.Key.foregroundColor: ColorPalette.gtBlue.uiColor]
```
levieggertcru, line 72: I'm wondering if this NSAttributedString logic should be moved out into the View or ViewModel layer maybe?  I don't know if the coloring here is domain layer specific since the domain layer really shouldn't know anything about the view.  
Let's leave this as is for now and think about that some more and can come back to this UseCase.

levieggertcru (changes requested):
Hey @rachaelblue looking good.  

Just a couple change requests on using the CloseButton.swift and making those flow steps unique.

I need to think about that UseCase a little more too.  I don't think we should be doing that ColorPalette logic in there since that is more View specific. It's tricky though because highlighting of the specific text does seem business rule related.  Let's think about it some more and also see what Jill says about iOS 14.

godtools/App/Features/AppLanguage/Presentation/ConfirmAppLanguage/ConfirmAppLanguageViewModel.swift
```diff
@@ -0,0 +1,68 @@
...
+// MARK: - Inputs
+
+extension ConfirmAppLanguageViewModel {
+    
+    func confirmLanguageButtonTapped() {
+        flowDelegate?.navigate(step: .appLanguageChangeConfirmed(appLanguage: selectedLanguage))
+    }
+    
+    func nevermindButtonTapped() {
+        dismissView()
```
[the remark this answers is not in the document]
rachaelblue: That makes sense and makes things more consistent.  Just another place where I need to be reprogrammed to stop consolidating haha.

godtools/App/Features/AppLanguage/Presentation/ConfirmAppLanguage/ConfirmAppLanguageView.swift
```diff
@@ -0,0 +1,102 @@
...
+                    }
+                    
+                    GTWhiteButton(title: viewModel.nevermindButtonTitle, fontSize: 15, width: buttonWidth, height: 48) {
+                        
+                        viewModel.nevermindButtonTapped()
+                    }
+                }
+                
+                Spacer()
+                Spacer()
```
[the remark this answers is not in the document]
rachaelblue: I've noticed the spacers work relative to each other.  I have 1 spacer above the content and 2 spacers below, so 1/3 of the space should be above and 2/3 below.

godtools/App/Features/AppLanguage/Domain/UseCases/GetConfirmAppLanguageInterfaceStringsUseCase.swift
rachaelblue: Yeah I went back and forth on this... the color has to be applied before the string formatting happens, and the string formatting makes sense to be in the useCase.  But I guess both could happen in the viewModel instead?  Or they take turns and the viewModel applies the color and then passes it back to the useCase to format?  Doesn't seem ideal any way you do it.

rachaelblue:
Hey @levieggertcru, fixed the flow steps and close button.  Also confirmed with Jill, and she's ok with iOS 14 not having the blue language name.  Let me know what you think about whether it makes more sense to move that logic into the viewModel.

godtools/App/Features/AppLanguage/Presentation/ConfirmAppLanguage/ConfirmAppLanguageView.swift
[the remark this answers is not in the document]
levieggertcru: Oh that's interesting I hadn't thought of it doing that.  That's pretty neat actually.

godtools/App/Features/AppLanguage/Domain/UseCases/GetConfirmAppLanguageInterfaceStringsUseCase.swift
```diff
@@ -0,0 +1,91 @@
...
+        
+        self.getCurrentAppLanguageUseCase = getCurrentAppLanguageUseCase
+        self.getInterfaceStringInAppLanguageUseCase = getInterfaceStringInAppLanguageUseCase
+        self.getInterfaceStringRepositoryInterface = getInterfaceStringRepositoryInterface
+        self.localeLanguageName = localeLanguageName
+    }
+    
+    func getStringsPublisher(for selectedLanguage: AppLanguageCodeDomainModel) -> AnyPublisher<ConfirmAppLanguageInterfaceStringsDomainModel, Never> {
+        
+        return Publishers.CombineLatest4(
```
levieggertcru, line 29: Hey @rachaelblue I'm also thinking about deprecating/removing the use case ```GetInterfaceStringInAppLanguageUseCase```.  

I realized that this use case can get very cumbersome to use when we have a lot of interface strings to translate and it also will create more overhead to have a publisher for each interface string.  For example the ToolDetails has about 10 interface strings to stuff into a domain model. 

Instead we can ask the data layer interface to translate our interface strings and the input would be the language to translate strings into.  For example in Tool Details it needs interface strings translated into the tool language.

https://github.com/CruGlobal/godtools-swift/blob/develop/godtools/App/Features/ToolDetails/Data-DomainInterface/GetToolDetailsInterfaceStringsRepository.swift

I'm going to create a separate ticket for this too so we can get this PR merged in.

levieggertcru (approved):

godtools/App/Features/AppLanguage/Domain/UseCases/GetConfirmAppLanguageInterfaceStringsUseCase.swift
rachaelblue: So the use case would end up passing two separate strings, and then the viewModel would apply the color and combine them into 1?  That should work, and would be nice for the domain model to pass a regular string and not an attributed one.

levieggertcru: Yep the domain model would contain both the message string and the highlight string and then leave it up to the View to apply the highlighting.
We should be able to write a behavior test as well that expects the highlight string to be the newly selected app language in one test and for the other message the highlight string to be the current app language.

levieggertcru (dismissed):

levieggertcru (approved):

levieggertcru (approved):

PR #1802, opened 2023-10-30
GT-2175 Update download tool progress view for app language

levieggertcru:
This PR updates the download tool progress view for the app language.
- Translates interface strings in current app language.
- Sets layout direction left to right and right to left for current app language.
- Added ProgressTimer to help simulate a download progress over a length of time.
- Adds feature DownloadToolProgress and di container for this feature.

download-tool-arabic-1
download-tool-arabic-2
download-tool-en-1
download-tool-en-2


godtools/App/Features/DownloadToolProgress/Data-DomainInterface/GetDownloadToolProgressInterfaceStringsRepository.swift
```diff
@@ -0,0 +1,54 @@
...
+        let downloadMessageLocalizedKey: String
+        let downloadMessage: String
+        
+        let resourceType: ResourceType? = resource?.resourceTypeEnum
+        
+        if resourceType == .article || resourceType == .tract, let resourceId = resource?.id {
+
+            let isFavoritedResource: Bool = favoritedResourcesRepository.getResourceIsFavorited(id: resourceId)
+            
+            downloadMessageLocalizedKey = isFavoritedResource ? "loading_favorited_tool" : "loading_unfavorited_tool"
```
levieggertcru, line 34: Hey @rachaelblue here's a situation where I don't exactly want logic like this in the data layer repository that is implementing the domain interface.  This logic checks if a tool is favorited / unfavorited and if it's unfavorited will display this special messaging  

"The tool is downloading. To avoid waiting in the future, favorite this tool so it will be available offline"

I'd like this logic to be included in a behavior test so I will somehow have to make that known in the UseCase.  Another thinker...

rachaelblue: Hey @levieggertcru, I guess my initial understanding was that useCases _exist_ to hold business rules, like this one.  But based on our conversation yesterday, it sounded like useCases should act as a bridge between the view and data, and not really do any additional logic themselves.  Is that right?  I feel like the more we go back and forth on this, the more confused I become 🫠  My inclination is to just use repositories to grab raw data, and then use cases to perform business rules given the data.

levieggertcru: Hey @rachaelblue sorry for the confusion, it's been a bit difficult for me to know the best way to apply these business rules like this.

It's correct that the UseCase exists to hold business rules and that logic has to happen in the domain layer.  Figuring out how that works together with the data layer is where I'm a bit uncertain. 
I'm wondering if we should try to keep it all in the domain models when we initialize them, or if there will be situations where we have to do it in the UseCase.

Your inclination seems correct to me, use repositories to grab raw data and then UseCases or DomainModels to perform business rules given the raw data.  We just need to nail down how we want that to happen.  If we should introduce another type that encapsulates the raw data, or if the raw data should get injected into the initializer of the domain model.

rachaelblue: No worries, it is tricky.  Yeah I guess we just try injecting into the initializer and see if that works in most scenarios.

levieggertcru: Yeah I think injecting into the domain model is a good start.  That will be simpler and not introduce additional files.

PR #1818, opened 2023-11-07
GT-2203 filter page use cases

rachaelblue:
Hey @levieggertcru, I moved the filter search logic into its own class `StringSearcher` and created use cases for language/category that used that class.  Wasn't sure where a class like that should live... I put it in data but maybe it should be in domain?  It's not a useCase, but it does seem more like business logic than data...

Also, I fixed the searchText binding in here like you had in #1763, so we can probably close that one if you're good with it.


levieggertcru:
Thanks @rachaelblue I'll start reviewing this today.  I like the idea of creating the StringSearcher class for sharing the search filtering logic between the use cases.
With clean architecture we can still create those distinct high level classes for our screen views and view models, and for our use cases, and when it comes to sharing logic and not duplicating code create those smaller components to share between them like this one.

godtools/App/Features/AppLanguage/Domain/UseCases/SearchAppLanguageInAppLanguagesListUseCase.swift
```diff
@@ -17,7 +17,7 @@ class SearchAppLanguageInAppLanguagesListUseCase {
         self.getAppLanguagesListUseCase = getAppLanguagesListUseCase
     }
     
-    func getSearchResultsPublisher(for searchTextPublisher: CurrentValueSubject<String, Never>) -> AnyPublisher<[AppLanguageListItemDomainModel], Never> {
+    func getSearchResultsPublisher(for searchTextPublisher: AnyPublisher<String, Never>) -> AnyPublisher<[AppLanguageListItemDomainModel], Never> {
```
levieggertcru, line 21: Hey @rachaelblue I may come back to this later and see about moving getAppLanguagesListUseCase.getAppLanguagesListPublisher() out of this UseCase and instead injecting the list of languages as a publisher in the method call along with the searchTextPublisher.  The 1 advantage there is it would make writing tests a little be easier.  For now though can leave as is.

levieggertcru:
Hey @rachaelblue yep can go ahead and close that PR with the search text two-way binding.

godtools/App/Share/Data/StringSearcher/StringSearcher.swift
```diff
@@ -0,0 +1,43 @@
...
+//  StringSearcher.swift
+//  godtools
+//
+//  Created by Rachael Skeath on 11/7/23.
+//  Copyright © 2023 Cru. All rights reserved.
+//
+
+import Foundation
+
+class StringSearcher {
```
levieggertcru, line 11: A couple thoughts on this class.

1. It would be nice if this class could implement an interface so the use cases could utilize dependency inversion and point to an interface.

2. What do you think about making the naming of this class a little more unique?  Maybe something like  IgnoringCaseStringSearcher?  This way if we have more string searcher implementations that implement the interface the naming can be a little more distinct and hint a little more at how searching is accomplished.

godtools/App/DependencyContainer/AppDataLayerDependencies.swift
```diff
@@ -266,6 +266,10 @@ class AppDataLayerDependencies {
         )
     }
     
+    func getStringSearcher() -> StringSearcher {
```
levieggertcru, line 269: Right I think placement could be elsewhere other than the data layer since it does seem more functionality based.
To me I feel like the domain layer is a good candidate or even in the share/common.

godtools/App/Share/Data/StringSearcher/StringSearcher.swift
```diff
@@ -0,0 +1,43 @@
...
+    
+    func search<T: StringSearchable>(for searchText: String, in searchables: [T]) -> [T] {
```
levieggertcru, line 13: Nice I like the utilization of generics :)

levieggertcru (changes requested):
Hey @rachaelblue, this is looking good.

I've made some updates to the change requests from yesterday. 

 I'm really not 100% certain where StringSearcher fits best as far as which folder to place it in and if it should go in AppDomainLayerDependencies or not.

godtools/App/Share/Data/StringSearcher/StringSearcher.swift
levieggertcru: Hey @rachaelblue let's leave this as is and can come back to creating an interface.  I did say in our last video chat that I'm not too worried about dependency inversion right now.

Once you start implementing the new download language work for that I'd like you to use dependency inversion.

rachaelblue: I initially thought I'd do dependency inversion here, but because it wasn't a repository it didn't fit the pattern we had talked about with dependency inversion.  So my followup question would be-- should we be using dependency inversion for _any_ type of dependencies? (any type of dependencies within use cases?)

godtools/App/Share/Data/StringSearcher/StringSearcher.swift
rachaelblue: I love generics 😄

rachaelblue:
Hey @levieggertcru, moved `StringSearcher` and have it instantiated directly when allocating the useCase.

godtools/App/Share/Data/StringSearcher/StringSearcher.swift
levieggertcru: That's a good question for this instance here, originally I was thinking we would use di for any UseCase dependency, but since StringSearcher contains some business logic we need, now I'm wondering if we should just instantiate it directly in the UseCase itself.  It seems like we would need this logic included for behavior testing.

levieggertcru: It'd be good to write some behavior tests for the app languages list searching and tool category and language filtering to see if it's best to instantiate it directly in the UseCase, but that can be done in a separate ticket.

levieggertcru: With dependency inversion I really think it can be used anywhere.  I grabbed a quick definition of it here (https://www.geeksforgeeks.org/dependecy-inversion-principle-solid/)

```Any higher classes should always depend upon the abstraction of the class rather than the detail```

It fits really nicely into our behavior tests because we can easily mock the data since we only care about testing business rules and logic and aren't concerned about the data itself.

I think there are negatives as well to adding too much DI.  With more interfaces you add more abstraction which can make large projects harder to figure out as you sift through code trying to find the concrete implementation.

levieggertcru (approved):

godtools/App/Share/Data/StringSearcher/StringSearcher.swift
rachaelblue: Yeah, I do worry that having so many files makes the project difficult to navigate.  Made a ticket for writing those behavior tests:  https://jira.cru.org/browse/GT-2208

levieggertcru: Thanks @rachaelblue for creating that ticket.

I was hesitant about adding the domain layer because of the added files when using di.  However, I do like the behavior tests and my hope is they will be really helpful in creating a source of truth as to how the app should behave and function.

With di the primary focus will be in the UseCase layer.

PR #1826, opened 2023-11-15
GT-2184 Deprecating AppLanguageCodeDomainModel

levieggertcru:
In this PR I will be removing AppLanguageCodeDomainModel.swift.  

I'm realizing it wasn't the best idea to restrict the current app language to an app language code since there is additional details a language can have such as region and script.

Apple's Locale class it's created by an identifier which can then be broken into language code, language region, and language script.  I think our domain model should instead be based on either a language id similar to the Locale id or a type that encapsulates language id, language code, and language script since we support scripts in our languages such as Chinese Simplified and Chinese Traditional.  

(https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html)


levieggertcru:
Hey @gyasistory and @rachaelblue I went through this PR again and will merge it in to develop.  Most of the changes are changing the type from AppLanguageCodeDomainModel to AppLanguageDomainModel and some naming changes from appLanguageCode to appLanguage.
As I mentioned in the PR description using language code for the app language didn't seem best as it causes confusion with how Locale treats a language code in a BCP-47 language identifier.
To me it makes more sense to treat the app language as a BCP-47 language identifier since it does include a script code for our Chinese languages.

PR #1827, opened 2023-11-16
GT-2184 add chinese simplified and chinese traditional in app languages list language name

levieggertcru:
This PR updates the following:

- Add LocaleLanguageScriptName in order to fetch the script name for a language.
- Add AppLanguageNameDomainModel to format our names with a script like so ```language_name (script_name)```
- Update AppLanguageDataModel to be created with a language code and optional language script code.  The languageId was changed to a computed property which will return a BCP-47 language identifier.

chinese-simplifed-and-chinese-traditional


godtools/App/Features/AppLanguage/Domain/Entities/AppLanguageNameDomainModel.swift
```diff
@@ -0,0 +1,26 @@
...
+//  Copyright © 2023 Cru. All rights reserved.
+//
+
+import Foundation
+
+struct AppLanguageNameDomainModel {
+    
+    let name: String
+    
+    init(languageName: String, languageScriptName: String?) {
```
levieggertcru, line 15: Hey @rachaelblue here's another example where I was able to pass the raw data to the domain model and then include some business formatting rules to format the language name.  I think this will work.

chinese-simplifed-and-chinese-traditional

PR #1828, opened 2023-11-16
GT-2156 persist filter selections

rachaelblue:


godtools/App/Features/Dashboard/Domain/UseCases/GetToolFilterCategoriesUseCase/GetToolFilterCategoriesUseCase.swift
```diff
@@ -49,6 +49,23 @@ class GetToolFilterCategoriesUseCase {
         return createAnyCategoryDomainModel(translationLocaleId: translationLocaleId, filteredByLanguageId: nil)
     }
     
+    func getCategoryFilterPublisher(with categoryId: String?) -> AnyPublisher<CategoryFilterDomainModel?, Never> {
```
rachaelblue, line 52: I know I broke the rule that use cases are only supposed to have 1 method here.  This use case was built to return _all_ the category filter objects together, but now I'm in a position of needing only a _single_ filter object (the one category filter that's been persisted).  So I have two public methods in this use case-- one to get all the category filters, and one to get a single filter. 

I considered how this could be avoided by passing "raw data" into the domain model init, but instantiating `CategoryFilterDomainModel` requires lots of dependencies and logic (see comment below), so passing "raw data" into the init doesn't seem to solve the problem here.

Then I thought about splitting it into two separate use cases, one for each method, but we recently talked about the fact that use cases aren't supposed to require other use cases (the list use case couldn't use the other use case).  And I don't want the logic duplicated between the use cases.  So that was out too.

I think the goal of having only 1 public method in a use case was to ensure the use case only has 1 responsibility.  So I felt like exposing a method that was already necessary to perform the responsibility didn't really break the single responsibility goal.  Thoughts?

godtools/App/Features/Dashboard/Domain/UseCases/GetToolFilterCategoriesUseCase/GetToolFilterCategoriesUseCase.swift
```diff
@@ -63,21 +80,28 @@ class GetToolFilterCategoriesUseCase {
...
+        
+        let toolsAvailableText: String = getToolsAvailableText(toolsAvailableCount: toolsAvailableCount, localeId: translationLocaleId)
+        
+        return CategoryFilterDomainModel(
+            type: .category(id: categoryId),
+            translatedName: translatedName,
+            toolsAvailableText: toolsAvailableText,
+            searchableText: translatedName
+        )
+    }
```
rachaelblue, line 103: I was trying to figure out how to move some of this logic into the `CategoryFilterDomainModel` init, but was having trouble getting there.  Would you think of moving the `getToolsAvailableText()` method into the init instead?  But it can't require `localizationServices`...

godtools/App/Features/Dashboard/Domain/UseCases/GetToolFilterLanguagesUseCase/GetToolFilterLanguagesUseCase.swift
```diff
@@ -47,6 +47,21 @@ class GetToolFilterLanguagesUseCase {
         return createAnyLanguageDomainModel(translationLocaleId: translationLocaleId, filteredByCategoryId: nil)
     }
     
+    func getLanguageFilterPublisher(from languageId: String?) -> AnyPublisher<LanguageFilterDomainModel?, Never> {
```
rachaelblue, line 50: Same thing in this use case-- I now have a second public method to get just a single language filter object instead of the list.

godtools/App/Features/Dashboard/Presentation/Tools/ToolsViewModel.swift
```diff
@@ -60,11 +60,32 @@ class ToolsViewModel: ObservableObject {
...
+                
+                if let categoryFilter = categoryFilter {
+                    self.toolFilterCategorySelectionPublisher.send(categoryFilter)
+                }
+                
+                if let languageFilter = languageFilter {
+                    self.toolFilterLanguageSelectionPublisher.send(languageFilter)
+                }
+            }
+            .store(in: &cancellables)
```
rachaelblue, line 87: This is another problem.  The `UserFilterDomainModel` returns filter IDs only, not the actual objects.  So here I'm fetching the object once I get the IDs.  I was inclined to put this logic directly into `GetUserFiltersUseCase`, but that would mean another use case inside a use case.

levieggertcru (approved):
Hey @rachaelblue I just realized the comments I had left in here are still pending, I was hoping to send these before we had our meeting today.  However, it doesn't affect this PR and we can move forward. 

Also I'm going to see if I can update the ResourceModel and LanguageModel to aid in querying the number of tools available. 

Overall this is looking good!  Here's some overview on this PR and how it relates to clean architecture:

Presentation Layer:
- This looks good.  I do like creating a separate view and viewModel for each screen and we should definitely do that going forward for any new screen.  We can still re-use UI by creating smaller components.  

Data Layer:
- Looks good, I like that there is a repository class for storing user filters and that it encapsulates another class that is responsible for the persistence.  This follows the pattern of our data layer repositories.

Domain-Interface:
- This looks good.  I see you created a class to implement our domain layer interface.  This class creates a bridge to the data layer repository.  Now we could put this interface directly on UserFiltersRepository, however, I do like the idea of creating an additional class to implement the interface.  I think it keeps the data layer a little more clean from the domain layer by doing it this way.

Domain Layer:

Interfaces:
- This looks good.  I see you created an interface for user filter storage and retrieval which is what we want in the UseCase to accomplish di.

UseCases:
-  The use cases we will need to update a little more for dependency inversion, but that's expected.  I do see why you created the GetUserFiltersUseCase as a way to chunk that into a single UseCase and I do see the approach you took to utilize that UseCase.  By creating separate UseCases for the filters it does create more files and more interfaces which may not be what we want.  I don't want to get too nit picky on UseCase naming either.  I find UseCase naming to be a bit difficult in some circumstances and even hard to reason what the UseCase should actually be.

Entities:
- It's possible these change a little as we implement more dependency inversion.  I'm thinking when it comes to injecting the raw data into a domain model and how business rules manipulates that data.

godtools/App/Features/Dashboard/Presentation/Tools/ToolsViewModel.swift
levieggertcru: Hey @rachaelblue let's come back and revisit this one.  I'm going to work on updating the data layer to hopefully make getting the number of tools available a bit easier to query.  Then it would be good to re-visit some of the UseCases in this feature to see how we can update them for using dependency inversion and writing behavior tests too.

rachaelblue:
Thanks @levieggertcru, I'd like to think about these options some, too.  Let's plan to meet on Thursday afternoon to discuss if that works for you.

levieggertcru:
Hey @rachaelblue sounds good, let's plan for Thursday afternoon.

PR #1829, opened 2023-11-17
GT-2156 create tools filter feature folder

levieggertcru:
Hey @rachaelblue I created a feature folder for the tool filtering in Features/ToolsFilter/.

I think it will help to pull some of this out of the dashboard as our Dashboard feature folder is getting pretty big.

Also, I do want to start creating feature folders like this as new features come in per OKR so now is a good time to do it.


rachaelblue (approved):
Ok sounds good!

PR #1840, opened 2023-12-05
GT refactor lessons view model use case

levieggertcru:
This PR does a minor refactor to LessonsViewModel by adding a ViewLessonsUseCase which encapsulates the interface strings and lessons.


godtools/App/Features/Lessons/Domain/Interface/GetLessonsListRepositoryInterface.swift
```diff
@@ -12,5 +12,4 @@ import Combine
 protocol GetLessonsListRepositoryInterface {
     
     func getLessonsListPublisher(appLanguage: AppLanguageDomainModel) -> AnyPublisher<[LessonListItemDomainModel], Never>
-    func observeLessonsChangedPublisher() -> AnyPublisher<Void, Never>
```
levieggertcru, line 15: Was able to remove the observer from the interface and move this to the class that implements the interface.

PR #1845, opened 2023-12-08
GT-2214 downloadable languages UI

rachaelblue:
- built downloadable language UI with updated animating download progress wheel (I don't know if this is in Figma, Aaron had just mentioned it on [Slack](https://cru-main.slack.com/archives/CG3S8471V/p1700253437837579?thread_ts=1700253238.346459&cid=CG3S8471V))
- tried out the new "view" useCase concept for the LanguageDownloadable, all it returns right now is the nav title.  Will be creating a followup ticket to get the data for the rest of the page.
- updated SearchBar so it doesn't use GetInterfaceStringInAppLanguageUseCase anymore.  Also using a "view" useCase here now


godtools/App/Features/AppLanguage/Data-DomainInterface/GetDownloadableLanguagesInterfaceStringsRepository.swift
```diff
@@ -0,0 +1,34 @@
...
+//  Copyright © 2023 Cru. All rights reserved.
+//
+
+import Foundation
+import Combine
+
+class GetDownloadableLanguagesInterfaceStringsRepository: GetDownloadableLanguagesInterfaceStringsRepositoryInterface {
+    
+    private let localizationServices: LocalizationServices
+    private let localeLanguageName: LocaleLanguageName
```
levieggertcru, line 15: Hey @rachaelblue I did notice this dependency is not being used, not sure if that was intended?

godtools/App/Share/Data-DomainInterface/GetSearchBarInterfaceStringsRepository.swift
```diff
@@ -0,0 +1,34 @@
...
+//  Copyright © 2023 Cru. All rights reserved.
+//
+
+import Foundation
+import Combine
+
+class GetSearchBarInterfaceStringsRepository: GetSearchBarInterfaceStringsRepositoryInterface {
+    
+    private let localizationServices: LocalizationServices
+    private let localeLanguageName: LocaleLanguageName
```
levieggertcru, line 15: Same here this dependency is not being used.

godtools/App/Features/AppLanguage/Domain/UseCases/ViewDownloadableLanguagesUseCase.swift
```diff
@@ -0,0 +1,29 @@
...
+    private let getInterfaceStringsRepository: GetDownloadableLanguagesInterfaceStringsRepositoryInterface
+    
+    init(getInterfaceStringsRepository: GetDownloadableLanguagesInterfaceStringsRepositoryInterface) {
+        
+        self.getInterfaceStringsRepository = getInterfaceStringsRepository
+    }
+    
+    func viewPublisher(appLanguage: AppLanguageDomainModel) -> AnyPublisher<ViewDownloadableLanguagesDomainModel, Never> {
+        
+        return self.getInterfaceStringsRepository.getStringsPublisher(translateInAppLanguage: appLanguage).map {
```
levieggertcru, line 23: Hmm ```map``` is a bit cleaner here, I probably should be using ```map``` as well instead of ```flatMap```.  I'll see about changing some of the UseCases to use ```map``` instead.

godtools/App/Share/SwiftUI Views/SearchBar/SearchBarViewModel.swift
```diff
@@ -11,13 +11,37 @@ import Combine
...
+    private let getCurrentAppLanguageUseCase: GetCurrentAppLanguageUseCase
+    private let viewSearchBarUseCase: ViewSearchBarUseCase
+    
+    private var cancellables = Set<AnyCancellable>()
+    
+    @Published private var appLanguage: AppLanguageDomainModel = ""
+    
     @Published var cancelText: String = ""
-
-    init(getInterfaceStringInAppLanguageUseCase: GetInterfaceStringInAppLanguageUseCase) {
```
levieggertcru, line 16: Nice I'm glad you were able to update this.  Definitely would like to phase out the GetInterfaceStringInAppLanguageUseCase.

godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/Subviews/LanguageDownloadIcon.swift
```diff
@@ -0,0 +1,91 @@
...
+            drawProgressInCircle(progress: progress)
+            
+        case .downloaded:
+            
+            Image(systemName: "checkmark")
+                .imageScale(.small)
+        }
+    }
+    
+    @ViewBuilder private func drawProgressInCircle(progress: Double) -> some View {
```
levieggertcru, line 53: Nice work on this, I like the circular animation effect.  We may at some point want to pull some of this out into a more generic shared view if we have a need to re-use this, but for now this is great.

godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/DownloadableLanguagesView.swift
```diff
@@ -0,0 +1,43 @@
...
+                    }
+                    .listRowBackground(Color.clear)
+                }
+            }
+            .listStyle(.inset)
+            
+        }
+        .navigationBarBackButtonHidden(true)
+        .navigationTitle(viewModel.navTitle)
+        .environment(\.layoutDirection, ApplicationLayout.shared.layoutDirection)
```
levieggertcru, line 41: Nice got the environment modifier in.  I still need to see if there's a way we don't have to add these manually, it's going to be difficult to remember to add these.

godtools/App/Flows/Flow/FlowStep.swift
```diff
@@ -150,6 +151,9 @@ enum FlowStep {
     case backTappedFromConfirmAppLanguageChange
     case chooseAppLanguageFlowCompleted(state: ChooseAppLanguageFlowCompleted)
     
+    // downloaded languages
+    case backTappedFromDownloadedLanguages
```
levieggertcru, line 155: Nice got the back button tapped step.

levieggertcru (changes requested):
Hey @rachaelblue this is looking great.   
- The flow steps and naming looks good.
- Had the inputs coming in on the DownloadableLanguagesViewModel. 
- I like the domain modeling for the language download progress you have started, I think that should work well once we implement the data layer side.  
- UI is looking good.  
- DI is looking good.

All in all looking good.  Only thing I noticed was those LocaleLanguageName dependencies getting added that weren't being utilized.

godtools/App/Features/AppLanguage/Data-DomainInterface/GetDownloadableLanguagesInterfaceStringsRepository.swift
rachaelblue: Whoops, not sure what I was thinking there 🤔 Removed both of those references.

levieggertcru (approved):

PR #1851, opened 2023-12-14
GT-2227 create shared class for translating a language name

levieggertcru:
This PR adds GetTranslatedLanguageName.swift which is a class for performing some business formatting on translating our app language names.  It will first check Localizable.strings to see if the translation exists there, it will then fallback to using iOS Locale and attempt to get the language name.  If all else fails it will fallback to the TranslatableLanguage.swift protocol fallbackName.

- Adds GetTranslatedLanguageName.swift to App/Share/Data-DomainInterface/Supporting/
- Adds GetTranslatedLanguageName to AppDataLayerDependencies
- Updates LanguageModel and AppLanguageDataModel to implement TranslatableLanguage
- Removes GetAppLanguageName.swift


godtools/App/Share/Data-DomainInterface/Supporting/GetTranslatedLanguageName/GetTranslatedLanguageName.swift
```diff
@@ -0,0 +1,88 @@
...
+//  GetTranslatedLanguageName.swift
+//  godtools
+//
+//  Created by Levi Eggert on 12/13/23.
+//  Copyright © 2023 Cru. All rights reserved.
+//
+
+import Foundation
+
+class GetTranslatedLanguageName {
```
levieggertcru, line 11: Hey @rachaelblue here is the class for sharing the business formatting for translating our app language names that are in the app languages list, downloadable languages list, and will also be used in the tool settings language list.

This was the idea I had for creating more granular classes that share business formatting, logic, etc.  It doesn't implement any interface, but can still be shared in our classes that implement domain interfaces.

I'm still not 100% sure on where to store these.  For now I placed this in /App/Share/Data-DomainInterface/Supporting/.

rachaelblue: Great!  Love having logic like this pulled out in a way that's easy to reuse.

rachaelblue (approved):

PR #1855, opened 2023-12-18
GT Add method to track download translations from remote with progress

levieggertcru:
This PR adds a way to track the download progress of translations by comparing how many translations have been downloaded against the total number of translations to download.  

Example method of downloading tool translations for a given language code with progress.

```swift
private func downloadToolTranslations(for languageCode: BCP47LanguageIdentifier) -> AnyPublisher<Double, Error> {
        
        let includeToolTypes: [ResourceType] = [.article, .tract, .lesson, .chooseYourOwnAdventure]
        
        let tools: [ResourceModel] = resourcesRepository.getCachedResourcesByFilter(filter: ResourcesFilter(category: nil, languageCode: languageCode, resourceTypes: includeToolTypes))
       
        let translations: [TranslationModel] = tools.compactMap {
            translationsRepository.getLatestTranslation(resourceId: $0.id, languageCode: languageCode)
        }
        
        guard !translations.isEmpty else {
            return Just(1).setFailureType(to: Error.self)
                .eraseToAnyPublisher()
        }
        
        return translationsRepository.getTranslationManifestsFromRemoteWithProgress(
            translations: translations,
            manifestParserType: .renderer,
            includeRelatedFiles: true,
            shouldFallbackToLatestDownloadedTranslationIfRemoteFails: true
        )
        .map {
            return $0.progress
        }
        .eraseToAnyPublisher()
    }
```


levieggertcru:
Hey @rachaelblue I added a way to track the download progress when downloading an array of translations.

I'm going to look into doing another update to the TranslationsRepository, I don't think it includes a way to check which translation version has already been downloaded.

As of now this new method will always download a translation from remote which would be wasteful if the translation is already downloaded.

rachaelblue (approved):
Nice!  Thanks for the example, too.  This is simpler than I imagined it being 😅

PR #1859, opened 2023-12-19
GT-2226 update downloadable languages list

rachaelblue:


godtools/App/Flows/LanguageSettings/LanguageSettingsFlow.swift
```diff
@@ -91,7 +91,9 @@ extension LanguageSettingsFlow {
             flowDelegate: self,
             getCurrentAppLanguageUseCase: appDiContainer.feature.appLanguage.domainLayer.getCurrentAppLanguageUseCase(),
             viewDownloadableLanguagesUseCase: appDiContainer.feature.appLanguage.domainLayer.getViewDownloadableLanguagesUseCase(),
-            viewSearchBarUseCase: appDiContainer.domainLayer.getViewSearchBarUseCase()
+            viewSearchBarUseCase: appDiContainer.domainLayer.getViewSearchBarUseCase(),
+            downloadToolLanguageUseCase: appDiContainer.feature.appLanguage.domainLayer.getDownloadToolLanguageUseCase(),
+            removeDownloadedToolLanguageUseCase: appDiContainer.feature.appLanguage.domainLayer.getRemoveDownloadedToolLanguageUseCase()
```
levieggertcru, line 96: I don't think there's a way to remove downloaded languages from the list, however from the language settings those will be able to be removed.

The UseCases do take some time setting up, but it is nice having them because I can see high level what is going on in the ViewModel.

levieggertcru: Ignore this just saw your comment in DownloadableLanguagesViewModel.

godtools/App/Features/AppLanguage/Data-DomainInterface/DownloadToolLanguageRepository.swift
```diff
@@ -0,0 +1,43 @@
...
+    
+    init(downloadedLanguagesRepository: DownloadedLanguagesRepository) {
+        
+        self.downloadedLanguagesRepository = downloadedLanguagesRepository
+    }
+    
+    func downloadToolLanguage(languageId: String) -> AnyPublisher<Bool, Never> {
+        
+        // TODO: - actually perform download
+        return Future() { promise in
```
levieggertcru, line 24: Hey @rachaelblue once this PR (https://github.com/CruGlobal/godtools-swift/pull/1855) is merged you'll be able to implement translation downloading with progress.
There's also an example snippet in the description with what that might look like.

godtools/App/Features/AppLanguage/Data-DomainInterface/GetDownloadableLanguagesListRepository.swift
```diff
@@ -0,0 +1,100 @@
...
+        self.downloadedLanguagesRepository = downloadedLanguagesRepository
+        self.getTranslatedLanguageName = getTranslatedLanguageName
+        self.resourcesRepository = resourcesRepository
+        self.localizationServices = localizationServices
+    }
+    
+    func getDownloadableLanguagesPublisher(currentAppLanguage: AppLanguageDomainModel) -> AnyPublisher<[DownloadableLanguageListItemDomainModel], Never> {
+        
+        return Publishers.CombineLatest(
+            languagesRepository.getLanguagesChanged(),
```
levieggertcru, line 32: Nice got both observers in there.

godtools/App/Features/AppLanguage/Data-DomainInterface/GetDownloadableLanguagesListRepository.swift
```diff
@@ -0,0 +1,100 @@
...
+                    languageNameInAppLanguage: languageNameInAppLanguage,
+                    toolsAvailableText: toolsAvailableText,
+                    downloadStatus: downloadStatus
+                )
+            }
+        }
+        .eraseToAnyPublisher()
+    }
+    
+    func observeLanguagesChangedPublisher() -> AnyPublisher<Void, Never> {
```
levieggertcru, line 64: Hey @rachaelblue go ahead and exclude this exposed observer.  Now that we're pushing business logic to the interface implementation we'll leave it up to the implementation to observer data layer changes.  

Eventually I'm going to go through and remove any observers on the domain layer interfaces and push those to the data layer implementation.  This is a result of our last architecture discussion.

godtools/App/Features/AppLanguage/Data/DownloadedLanguagesRepository/Cache/RealmDownloadedLanguage.swift
```diff
@@ -13,6 +13,7 @@ class RealmDownloadedLanguage: Object {
     
     @objc dynamic var createdAt: Date = Date()
     @objc dynamic var languageId: String = ""
+    @objc dynamic var downloadProgress: Double = 0
```
levieggertcru, line 16: Hey @rachaelblue what is your thinking here on persisting the download progress?  Would this be because if there's only a partial download and then something interrupts the download process?

godtools/App/Features/AppLanguage/Domain/Entities/LanguageDownloadStatusDomainModel.swift
```diff
@@ -13,4 +13,22 @@ enum LanguageDownloadStatusDomainModel: Equatable {
     case notDownloaded
     case downloading(progress: Double)
     case downloaded
+    
+    init(downloadedLanguage: DownloadedLanguageDataModel?) {
```
levieggertcru, line 17: Hey @rachaelblue one rule in the domain layer is it's not supposed to know anything about the data layer.  

However, it's perfectly fine that the data layer knows about the domain layer.  I'd suggest either performing this logic in the interface implementation, or you could even create an extension off of DownloadedLanguageDataModel.

godtools/App/Features/AppLanguage/Domain/Interface/GetDownloadableLanguagesListRepositoryInterface.swift
```diff
@@ -0,0 +1,16 @@
...
+//  Copyright © 2023 Cru. All rights reserved.
+//
+
+import Foundation
+import Combine
+
+protocol GetDownloadableLanguagesListRepositoryInterface {
+    
+    func getDownloadableLanguagesPublisher(currentAppLanguage: AppLanguageDomainModel) -> AnyPublisher<[DownloadableLanguageListItemDomainModel], Never>
+    func observeLanguagesChangedPublisher() -> AnyPublisher<Void, Never>
```
levieggertcru, line 15: Left a comment else where, but after our last discussion on which architecture format to go with we won't need these exposed observers.  All data layer observing will now be pushed to the interface implementation.

godtools/App/Features/AppLanguage/Data-DomainInterface/GetDownloadableLanguagesListRepository.swift
```diff
@@ -0,0 +1,100 @@
...
+                let downloadStatus = self.getDownloadStatus(for: language.id)
+                
+                return DownloadableLanguageListItemDomainModel(
+                    languageId: language.id,
+                    languageNameInOwnLanguage: languageNameInOwnLanguage,
+                    languageNameInAppLanguage: languageNameInAppLanguage,
+                    toolsAvailableText: toolsAvailableText,
+                    downloadStatus: downloadStatus
+                )
+            }
```
levieggertcru, line 59: Hey @rachaelblue also see about sorting this list, I included the notes from Figma in the Jira ticket.  It seems they want sorting by download order first and then remaining non downloaded languages sorted alphabetically.  I'm not sure which language though?  I'd guess app language, but may want to confirm with Jill on the sorting.

godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/DownloadableLanguagesViewModel.swift
```diff
@@ -77,30 +75,27 @@ extension DownloadableLanguagesViewModel {
...
         switch downloadableLanguage.downloadStatus {
+            
         case .notDownloaded:
             
-            updateDummyLanguage(with: .downloading(progress: 0))
+            downloadToolLanguageUseCase.downloadToolLanguage(downloadableLanguage.languageId)
+                .sink { _ in
+                    
+                }
+                .store(in: &cancellables)
```
levieggertcru, line 86: The one caveat to storing a cancellable here is if the View is existed while a download process is happening, that download would be cancelled, which may not be what we want.  I'd confirm with Jill if that is the behavior they want.

One solution I found to keep publishers alive is to have a static cancellable that references the result of the publisher chain.  Not sure if there's a cleaner or preferred way, would be curious to investigate that more.

levieggertcru (changes requested):
Hey @rachaelblue this is looking really good!  I'm actually really liking the dependency inversion when it comes to reviewing the code.  It really helps to break up the ViewModel and isolate some of those business logic and rules in the interface implementation.

I left some comments on some minor updates and let me know if you want to include the translation downloading with progress in this PR or a separate one.  Looking great!

godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/DownloadableLanguagesViewModel.swift
levieggertcru: Actually static variable wouldn't work here because multiple languages could be downloading.

godtools/App/Features/AppLanguage/Data-DomainInterface/GetDownloadableLanguagesListRepository.swift
```diff
@@ -0,0 +1,100 @@
...
+            languagesRepository.getLanguagesChanged(),
+            downloadedLanguagesRepository.getDownloadedLanguagesChangedPublisher()
+        )
+        .map { _ in
+            
+            return self.languagesRepository.getLanguages().map { language in
+                                
+                let languageNameInOwnLanguage = self.getTranslatedLanguageName.getLanguageName(
+                    language: language,
+                    translatedInLanguage: language.id
```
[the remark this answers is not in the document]
rachaelblue: Hey @levieggertcru, what's the difference between `language.code` and `language.languageCode`?

godtools/App/Features/AppLanguage/Data-DomainInterface/DownloadToolLanguageRepository.swift
rachaelblue: Thank Levi.  Made a follow-up ticket for implementing downloading: https://jira.cru.org/browse/GT-2234

godtools/App/Features/AppLanguage/Data-DomainInterface/GetDownloadableLanguagesListRepository.swift
rachaelblue: Good point!  I like having the observer in the interface implementation with the rest of the logic, too.

godtools/App/Features/AppLanguage/Data/DownloadedLanguagesRepository/Cache/RealmDownloadedLanguage.swift
rachaelblue: This was more of an easy and temporary solution to maintain the download animation before the real download logic gets implemented 😄 I expect we don't want this persisted, although I'm not sure what happens in the case of a partial download... would it just restart the whole thing?

godtools/App/Features/AppLanguage/Data-DomainInterface/GetDownloadableLanguagesListRepository.swift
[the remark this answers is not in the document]
levieggertcru: The ```language.code``` we get from the mobile-content-api is basically the equivalent to iOS Locale identifier which contains the format ```languagecode-scriptcode_regioncode```.
The ```language.languageCode``` is just pulling the language code portion out of the Locale identifier.  Also named language designator.  That way when translating the language name we're just translating the language designator.  Then we pull the script code out on its own and translate that portion.
Here's the iOS Locale documentation on how they format those.(https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html)

godtools/App/Features/AppLanguage/Data/DownloadedLanguagesRepository/Cache/RealmDownloadedLanguage.swift
levieggertcru: I was suspecting that it was temporary.  We could track the progress, although I was thinking we'd just restart the download.
But, this just made me think of something else.  I'm thinking that whenever a user starts a download of a language, we'd want to store / track that downloadable language immediately in the repo.
There is going to be some additional logic that checks which languages have been "downloaded" in that repo, and then download any latest translations that haven't been downloaded.
I'm thinking we'd store that downloadable language immediately to the repo in the event something does interrupt the app, then next app launch the language will get picked up in our process that downloads latest translations for those.

godtools/App/Features/AppLanguage/Data-DomainInterface/GetDownloadableLanguagesListRepository.swift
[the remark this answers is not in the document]
rachaelblue: Oh interesting.  I never knew what the parts of the code were called (language designator, region designator) and how they worked.  Good to know!

godtools/App/Features/AppLanguage/Data/DownloadedLanguagesRepository/Cache/RealmDownloadedLanguage.swift
rachaelblue: Gotcha, so a flag or something that gets set on the Realm object once download completes?  I'm assuming that will happen in a later ticket?

godtools/App/Features/AppLanguage/Domain/Entities/LanguageDownloadStatusDomainModel.swift
rachaelblue: Ah good catch, still learning to think about which way the data is flowing.  Just moved into the interface implementation for now since the download status logic may change.

godtools/App/Features/AppLanguage/Data-DomainInterface/GetDownloadableLanguagesListRepository.swift
rachaelblue: Confirmed with Jill to sort by app language translation.

godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/DownloadableLanguagesViewModel.swift
rachaelblue: Confirmed with Jill that the download should complete even if the page is closed.  I tried to do some googling on retaining subscriptions until completion, but didn't find much.  Only alternative I saw was to create a custom publisher that makes sure deallocation doesn't occur until the task is complete.  Seems easier to just keep it static, but what do you think?  For now, I just made a static cancellables set.

godtools/App/Features/AppLanguage/Data-DomainInterface/GetDownloadableLanguagesListRepository.swift
```diff
@@ -0,0 +1,123 @@
...
+            
+        } else {
+            
+            return .downloading(progress: downloadedLanguage.downloadProgress)
+        }
+    }
+    
+    private func getSortOrder(language1: DownloadableLanguageListItemDomainModel, language2: DownloadableLanguageListItemDomainModel) -> Bool {
+        
+        if language1.isDownloaded && !language2.isDownloaded {
```
levieggertcru, line 112: Hey @rachaelblue I'm wondering if here we should be sorting those by when the language was downloaded.  We do track that ```createdAt: Date``` in DownloadedLanguageDataModel to know when a language was downloaded. That way the order would remain consistent.
I didn't see in Figma if it should be ascending or descending.  If they want most latest downloaded at the top or the first downloaded at the top.

godtools/App/Features/AppLanguage/Data/DownloadedLanguagesRepository/Cache/RealmDownloadedLanguage.swift
levieggertcru: It wouldn't be an additional attribute, it would instead be a matter of when we store a language as "downloaded" in the DownloadedLanguagesRepository since we delete those when a language is removed.

In our case I'd say it would be safer to store a language as "downloaded" immediately when the translation downloading begins.  Especially now that we know those should continue downloading in the background.  This way if the download process is interrupted by the app getting closed etc, we can still download translations for the language next time the app is launched.

If we wait to store a language as downloaded when all translations have completed, we run the risk of losing that a language as marked for download if for some reason the download process is interrupted.

levieggertcru: That can be part of GT-2234 for translation downloads.

godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/DownloadableLanguagesViewModel.swift
levieggertcru: Do you the link for the custom publisher?  I'm curious to see what that entails.

And for the static array I'm curious to see if those cancellable's get removed when the publisher completes.  Let's leave the static array.  I'm curious to test that some more when GT-2234 is completed.

godtools/App/Features/AppLanguage/Data-DomainInterface/GetDownloadableLanguagesListRepository.swift
levieggertcru: Hey @rachaelblue let's leave this sorting as is, all it says in the Figma is that downloaded languages should be at the top.  Going to confirm with Jill if there should be any extra sorting on those and that can be done in a separate ticket if needed.  The only concern I have is that the order of those downloaded languages at the top could appear random since it would depend on how languages are fetched from realm.

levieggertcru (approved):

godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/DownloadableLanguagesViewModel.swift
rachaelblue: Taking a second look, [this](https://forums.swift.org/t/combine-best-practices-with-memory-management-using-subjects-to-publish-values/38724/2) probably wouldn't solve our problem.  Although it is interesting, and possibly prone to memory leaks if you're not careful.  But yeah, seems like even if we did a custom class of some sort, we'd still end up needing something `static` to keep it alive.

godtools/App/Features/AppLanguage/Data-DomainInterface/GetDownloadableLanguagesListRepository.swift
rachaelblue: Ok yeah, the downloaded languages at the top will still sort alphabetically relative to each other, the way I have it right now.  But Figma doesn't show them alphabetically, so maybe it is by timestamp?  Would be good to check with Jill.

godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/DownloadableLanguagesViewModel.swift
levieggertcru: Thanks @rachaelblue, yeah I like the static approach better, curious to test that some more though to see if those cancellable's are removed from the static array reference.  That's my only concern is leaving those in memory.

levieggertcru: The static references for those single cancellable's I should check too.

godtools/App/Features/AppLanguage/Data-DomainInterface/GetDownloadableLanguagesListRepository.swift
levieggertcru: Okay nice I didn't realize those were sorting alphabetically.  Yeah double check with Jill.  Just as long as they are sorting in some manner would be preferable that way the list stays consistent each time a user taps into it.

rachaelblue: Jill said sort alphabetically, so we're good 👍

levieggertcru: Awesome thanks @rachaelblue

levieggertcru:
Hey @rachaelblue I was curious about the performance issue you mentioned in the downloadable languages list, and after running Time Profiler the biggest hit on the main thread is at LocalizableStringsBundle.stringForKey.  

Without the Time Profiler I probably wouldn't have expected that.  But it makes sense because basically we're loading 2 Bundle's for every language in that list when translating the current app language name and the language name.

In GetTranslatedLanguageName.swift the first thing that happens is a check in Localizable.strings for the translation.  That check will load a Bundle for the given translated language and then check Localizable.strings for the language.  I'm going to see if I can first check Locale instead of Localizable.strings.  I think that along should help with some of the performance.

Here's a couple screenshots that shows the call stack when downloadable languages list is opened:

downloadable-languages-profiler

downloadable-languages-profiler-expanded

rachaelblue:
😮 wow interesting, I would not have guessed that at all.  Thanks for investigating this @levieggertcru, I'll be curious to see if checking locale first fixes it.

PR #1861, opened 2023-12-21
GT Deprecate and obsolete TextWithLinks

levieggertcru:
This is pretty cool, you can mark something as ```obsoleted``` with iOS version.  This way when we move the minimum target to iOS 15 we will get an error here.


rachaelblue (approved):
Nice!  Also not sure why it took SwiftUI so long to support links 😄

PR #1866, opened 2023-12-27
GT update resource filtering for variants and meta

levieggertcru:


levieggertcru:
Hey @rachaelblue updated the ResourcesFilter.swift to include optional enum ```variants: ResourcesFilterVariant?```.

Also updated RealmResource.swift to include property ```isVariant: Bool``` and relationships ```defaultVariant: RealmResource?``` ```variants = List<RealmResource>()```.

When using ```.isNotVariant``` in the filter the filter will include RealmResources with isVariant = false.
When using ```.isVariant``` in the filter the filter will include RealmResources with isVariant = true.
When using ```.isDefaultVariant``` in the filter the filter will include RealmResources with isVariant = true and where RealmResource id is equal to metatool.defaultVariantId.

This will allow us to fetch resources that are not variants, resources that are variants, and resources that are only the default variants.

rachaelblue:
Ok nice!  Thanks @levieggertcru!  Looks easier to read this way, too 👍

levieggertcru:
@rachaelblue only downside is the method to filter those uses ```NSCompoundPredicate(andPredicateWithSubpredicates:``` 

May need to provide a way where we can individually grab those NSPredicates if we want to use ```OR``` to combine predicates.

PR #1884, opened 2024-01-12
GT-2234 perform translation downloads

rachaelblue:


godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/DownloadableLanguagesViewModel.swift
```diff
@@ -76,14 +114,34 @@ extension DownloadableLanguagesViewModel {
...
+                    case .finished:
+                        
+                        self?.activeDownloads.removeValue(forKey: languageId)
+                        
+                    case .failure(let error):
+                        
+                        // TODO: - what happens during a failure?
+                        
+                        self?.activeDownloads[languageId] = .notDownloaded
+                        self?.flowDelegate?.navigate(step: .showLanguageDownloadErrorAlert(error: error))
```
rachaelblue, line 139: Hey @levieggertcru, there was one time last week that the service was failing, so I added an alert for the moment so it's obvious when that happens.  Do we know what's supposed to happen instead (do we retry?)?  I can catch up with Jill/Daniel if needed and make a follow-up ticket.

godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/Subviews/DownloadableLanguageItemView.swift
```diff
@@ -51,9 +55,50 @@ struct DownloadableLanguageItemView: View {
...
+            
+            if downloadProgress < progressTarget {
+                
+                self.animationDownloadProgress? += 0.1
+                
+            } else if progressTarget >= 1 && downloadProgress >= 1 {
+                
+                timer.invalidate()
+                self.timer = nil
+            }
```
rachaelblue, line 101: Often times the download happens so fast that you can't even see the download progress wheel animate.  So this forces the animation to go slow enough to be visible even if the download has already finished, or pauses the animation if the download hasn't progressed enough.

godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/DownloadableLanguagesViewModel.swift
levieggertcru: Hey @rachaelblue hmm that's a good question. I feel like we should do something here.  I'm curious how @frett is handling languages that fail to download.

godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/Subviews/DownloadableLanguageItemView.swift
levieggertcru: Hey @rachaelblue that sounds good!  Sounds like a good idea to slow that down a bit for fast downloads.

godtools/App/Features/AppLanguage/Data/DownloadedLanguagesRepository/Cache/RealmDownloadedLanguage.swift
```diff
@@ -13,7 +13,7 @@ class RealmDownloadedLanguage: Object {
     
     @objc dynamic var createdAt: Date = Date()
     @objc dynamic var languageId: String = ""
-    @objc dynamic var downloadProgress: Double = 0
+    @objc dynamic var downloadComplete: Bool = false
```
levieggertcru, line 16: Hey @rachaelblue this looks good.  I didn't think we would need to track this here, but it makes sense because a user may add a language and either it fails to download or doesn't quite finish yet and we still want to track that as part of their downloadable languages.

godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/DownloadableLanguagesViewModel.swift
```diff
@@ -54,9 +56,45 @@ class DownloadableLanguagesViewModel: ObservableObject {
...
+
+// MARK: - Private
+
+extension DownloadableLanguagesViewModel {
+    
+    private func updateDisplayedLanguages(from downloadableLanguages: [DownloadableLanguageListItemDomainModel], activeDownloads: [BCP47LanguageIdentifier: LanguageDownloadStatusDomainModel]) {
+        
+        var updatedLanguages: [DownloadableLanguageListItemDomainModel] = Array()
+        
+        for downloadableLanguage in downloadableLanguages {
```
levieggertcru, line 84: My only concern with looping here is if we had a lot of downloadableLanguages you could see some UI slowdown here.  Would be good to do some performance tests.

godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/DownloadableLanguagesViewModel.swift
levieggertcru: I feel like we should do something here, maybe provide additional UI that shows the download failed with option to retry.  Let's definitely hold off for a separate ticket on that.  I feel like an alert message could get messy if multiple downloads are failing.

levieggertcru (approved):
Hey @rachaelblue this is looking good.  Just the timer needs to be stopped and started when that view is no longer visible on screen.  Feel free to do that in a separate task.

I also want to do some performance testing on the forloop I had commented on.  But first let's get the timer life cycle in and I still need to fix the performance slowdown on fetching the translated language name and also fix any other retain cycles like the one found here https://github.com/CruGlobal/godtools-swift/pull/1891.

godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/DownloadableLanguagesViewModel.swift
frett: I know it's a little bit of a delayed response, but on Android the DownloadManager will transparently retry in the background when it can. My download progress indicator is also linked to the number of translations currently downloaded, so it will show as incomplete until that last translation is downloaded successfully

frett: on Android the interaction between this view and the download manager is indirect.

This UI just toggles a setting in the database and displays what the database reports as already downloaded.

The DownloadManager monitors the database for changes, and proactively downloads translations for tools that are marked to be downloaded. Once it finishes downloading a translation it marks that in the database.

So those 2 components both interact with the database, but aren't dependent on each other

levieggertcru: Thanks @frett for the feedback.  I'm going to create a ticket to have our service do some type of retry.

PR #1916, opened 2024-01-25
GT-2255 pause download timer on background

rachaelblue:


godtools/App/Share/Extensions/View+AppLifecycle.swift
```diff
@@ -0,0 +1,31 @@
...
+//
+//  Created by Rachael Skeath on 1/25/24.
+//  Copyright © 2024 Cru. All rights reserved.
+//
+
+import SwiftUI
+
+extension View {
+    
+    // NOTE: SwiftUI supports lifecycle observation through `ScenePhase` (https://developer.apple.com/documentation/swiftui/scenephase),
```
levieggertcru, line 13: Hmm this is interesting, I haven't heard of this before.  Will definitely have to take a look into this.

levieggertcru (changes requested):
Hey @rachaelblue this is good.  I actually wasn't thinking about this scenario when the app is backgrounded, but I think this is still good to have for that specific case. 

We also want the timer to be stopped when the List item view is no longer on the screen due to scrolling through the list.  Similarly to when a UITableViewCell disappears and is recycled due to scrolling through the TableView.  

It might be the onAppear and onDisappear in SwiftUI?

godtools/App/Share/Extensions/View+AppLifecycle.swift
levieggertcru: Looks like we'd also want to move away from using UIApplicationDelegate and use the SceneDelegate.  Something I'd like to shift towards soon too. https://www.kairadiagne.com/2019/11/18/adopting-the-new-scene-delegate

levieggertcru (dismissed):
Hey @rachaelblue this is working good!  I started some downloads and backed out of the list mid download and the ViewModel was deallocated so that part is working.

I also noticed that when I tapped to download a language and backed out immediately the download would complete.  I could go back to the list and it would show as downloaded.

I did notice another issue which is separate and is something that is more of an edge case.  I'll try my best to explain.  

I tapped to download Arabic and mid download I backed out of the view and either the download failed or I quit the app right away so it never completed.  When going back into the list the icon appeared as not downloaded but in blue color like in the attached screenshot.

downloadable-languages

I noticed in the repository implementation the status was also landing on .downloading(progress: nil).  (https://github.com/CruGlobal/godtools-swift/blob/develop/godtools/App/Features/AppLanguage/Data-DomainInterface/GetDownloadableLanguagesListRepository.swift#L117)

Basically this is a case where the language is added but hasn't completed downloading either because the download failed, a user quit the app, network connection was lost etc.  Once we complete the work to download languages in this list on next app launch it should resolve this as long as someone has a good network connection.  But still a minor edge case I wanted to make aware.

rachaelblue:
Hey @levieggertcru, that was a good catch!  Easy enough fix though, so went ahead and threw it in here.

levieggertcru (approved):
Nice that did it.  I think that works since it hasn't fully downloaded yet.  
Once the work is in to download those in the background on next app launch this will be a smaller edge case.

download-progress-nil

PR #1917, opened 2024-01-26
GT-2123 confirm app language view

rachaelblue:
Small fixes based on Jill's comments in the ticket:
- make paragraphs the same font size
- button text should be translated in app language, not selected language
- switch order of buttons


levieggertcru (changes requested):
Hey @rachaelblue I hate to do this now, but I'm thinking this would be a good time to update the GetConfirmAppLanguageInterfaceStringsUseCase to point to an interface instead. I do want to phase out GetInterfaceStringInAppLanguageUseCase and GetInterfaceStringForLanguageRepositoryInterface.

And it would also be a good time to remove the GetCurrentAppLanguageUseCase dependency from the UseCase and instead inject the appLanguage as an input. 

Let me know if you have any questions on that.   

Also let me know on timing, it would be nice to make this change now before testing, but I also know you're on a shorter time schedule and if we have to do that work in a separate ticket I'm okay with that too.

rachaelblue:
Hey @levieggertcru, yeah you're right, good idea to make those updates before testing.  Will make those changes in this PR.

levieggertcru:
Awesome thank you.

rachaelblue:
Hey @levieggertcru, I went ahead and changed `GetConfirmAppLanguageInterfaceStringsUseCase` to `ViewConfirmAppLanguageUseCase` and made those changes you mentioned.

levieggertcru (approved):
Thanks @rachaelblue!  Nice to have those deprecated dependencies out of the way.

PR #1918, opened 2024-01-26
GT-2222 change tool primary to app language

levieggertcru:
This task adds selectedLanguageIndex: Int to base class MobileContentPagesViewModel.  This will allow us to control which language should start as selected.

- When opening a tool from the tools list, the tool will be opened with the app language as primary and the tool filter language as parallel and the parallel language will be selected.


levieggertcru:
Moved back to draft.  Might make a change on how the parallel language is marked for selection.

PR #1943, opened 2024-02-05
GT-2214 Undownload language button

rachaelblue:
Current logic is that tapping once on a downloaded language should show the red "x", and tapping a second time removes the download.  Waiting to hear back from Jill if there's supposed to be a way to cancel if a user gets the "x" but doesn't want to download.  Depending on what the solution is, I'm thinking that will just be a separate ticket.

### DIFF
--- a/godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/DownloadableLanguagesView.swift
+++ b/godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/DownloadableLanguagesView.swift
@@ -28,7 +28,7 @@ struct DownloadableLanguagesView: View {
                     
                     DownloadableLanguageItemView(downloadableLanguage: downloadableLanguage) {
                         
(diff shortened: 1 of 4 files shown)


levieggertcru:
Hey @rachaelblue sounds good, it seems there definitely needs to be something to reverse that.

I also have a change request which could be done in that separate ticket or this one.  Let me know which you'd prefer, but I was looking at ```DownloadableLanguagesViewModel.swift``` and instead of having a single input ```downloadableLanguageTapped``` I would like 2 separate inputs.  An input for downloading a language and an input for removing a language.

The ViewModel doesn't necessarily know how the View is structuring UI elements or that it's sharing the tap action for those, it simply declares what inputs are available to the View and it's up to the View to call those.

rachaelblue:
Hey @levieggertcru, since Jill's change to hide the "x" was simple enough (cancel after 3 seconds unless tapped again), I went ahead and did that here as well as splitting the input into two separate ones.

levieggertcru (approved):
Thanks @rachaelblue looks good.

PR #1960, opened 2024-02-21
GT-2159 translate account activity using app lang

rachaelblue:
Hey @levieggertcru, I removed `GetInterfaceStringInAppLanguageUseCase` from `AccountViewModel` and created `ViewAccountUseCase` to handle string translations.  I was debating whether the logic from the other use cases in `AccountViewModel` (like `GetUserAccountDetailsUseCase` and `GetGlobalActivityThisWeekUseCase`) should be moved into repositories and retrieved through `ViewAccountUseCase` too, but it seemed possible they could delay `ViewAccountUseCase` from returning as quickly.  Thoughts on this?


godtools/App/Flows/Menu/MenuFlow.swift
```diff
@@ -318,14 +318,16 @@ class MenuFlow: Flow {
     private func getAuthErrorAlertMessage(authError: AuthErrorDomainModel) -> AlertMessageType {
         
         let localizationServices: LocalizationServices = appDiContainer.dataLayer.getLocalizationServices()
+        let appLanguageLocaleId = appDiContainer.feature.appLanguage.domainLayer.getCurrentAppLanguageUseCase().getLanguage().localeId
```
rachaelblue, line 321: Hey @levieggertcru, is there a way you'd prefer to get the app language here rather than instantiating the use case and pulling it out?  Not sure I know how the flows fit into clean architecture.

levieggertcru:
Hey @rachaelblue that is a potential problem using CombineLatest, if one of the publishers are initially delayed it will delay everything else.
I think it would be nice to model everything within the ViewUseCase that is being viewed, however, I just don't know how realistic it will be.  Another situation is if we have a ViewUseCase that has a lot of data in it that is Viewable and it starts to get massive in scale.  I think these are 2 areas where we can chunk things out where it makes sense.

One way to solve the delay is to ensure we always have data persisted and immediately ready.  This could be a general rule we use throughout the app.

godtools/App/Flows/Menu/MenuFlow.swift
levieggertcru: The AlertMessageView has been an issue since UIAlertController requires the title and message on instantiation.  

In these situations since I really don't want to perform a sink and then call the alert message, I've been keeping a reference on the flow via Published property to keep the reference updated when the app language changes.
 
I'll either keep a reference to the app language or create a domain model that encapsulates the interface strings for the alert and keep a reference to that on flow.

In your situation I'd say it's fine to just keep a reference to the app language updated on the Flow.

Example from AppFlow (https://github.com/CruGlobal/godtools-swift/blob/develop/godtools/App/Flows/App/AppFlow.swift#L85)

godtools/App/DependencyContainer/AppDataLayerDependencies.swift
```diff
@@ -33,7 +33,13 @@ class AppDataLayerDependencies {
     }
     
     // MARK: - Data Layer Classes
-            
+    
+    func getAccountInterfaceStringsRepositoryInterface() -> GetAccountInterfaceStringsRepositoryInterface {
```
levieggertcru, line 37: This is the best place here for this dependency.  However, I still need to spend some time on organizing the features.  There is an AccountCreation feature, but now I'm wondering if we should just have a general Account Feature for account related things.

levieggertcru (approved):
Thanks @rachaelblue this is a great start for app language changes.

I'm going to make a separate task for refactoring for dependency inversion.

levieggertcru (changes requested):
Hey @rachaelblue I approved this, but I remember I had a comment on setting the Locale on Calendar.

godtools/App/Features/Menu/Data-DomainInterface/GetAccountInterfaceStringsRepository.swift
```diff
@@ -0,0 +1,51 @@
...
+        return Just(interfaceStrings)
+            .eraseToAnyPublisher()
+    }
+    
+    private func getGlobalAnalyticsTitle(localeId: BCP47LanguageIdentifier) -> String {
+    
+        let localizedGlobalActivityTitle = localizationServices.stringForLocaleElseEnglish(localeIdentifier: localeId, key: MenuStringKeys.Account.globalAnalyticsTitle.rawValue)
+        
+        let todaysDate: Date = Date()
+        let todaysYearComponents: DateComponents = Calendar.current.dateComponents([.year], from: todaysDate)
```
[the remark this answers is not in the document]
levieggertcru: Hey @rachaelblue ignore this request.  I ended up moving this to this ticket here -> https://jira.cru.org/browse/GT-2302

godtools/App/Features/AppLanguage/Domain/Interface/GetAppLanguageRepositoryInterface.swift
```diff
@@ -12,4 +12,5 @@ import Combine
 protocol GetAppLanguageRepositoryInterface {
     
     func getLanguagePublisher() -> AnyPublisher<AppLanguageDomainModel, Never>
+    func getLanguage() -> AppLanguageDomainModel
```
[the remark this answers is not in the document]
rachaelblue: Hey @levieggertcru, this is a rule I hadn't considered-- the repository interfaces should also only have a single method?

rachaelblue:
Hey @levieggertcru, sorry it's taken me a while to get back to this.  Is there any work that needs to happen here?  It looks like the change requests all got pulled out to separate tickets.

godtools/App/Features/AppLanguage/Domain/Interface/GetAppLanguageRepositoryInterface.swift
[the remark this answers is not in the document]
levieggertcru: Hey @rachaelblue let me get back to you on this question about interfaces having only a single method.  I noticed in your latest PR you have a couple methods when needing to fetch the any category.  

However, I still think this method would go away once updating the MenuFlow to reference the app language.

godtools/App/Flows/Menu/MenuFlow.swift
levieggertcru: Hey @rachaelblue no problem, these are the change requests for this PR.  
- Remove ```func getLanguage() -> AppLanguageDomainModel``` in GetCurrentAppLanguageUseCase and GetAppLanguageRepositoryInterface.  We only want the single publisher method on the UseCase. 
- See my comment above on referencing appLanguage from the MenuFlow.  That will resolve the need to call getLanguage() on the UseCase since you will have a live reference to the appLanguage.

rachaelblue: Thanks Levi!  Made those changes.

levieggertcru (approved):

PR #1985, opened 2024-03-06
GT-2158 search bar rules

rachaelblue:
- updated the 4 search bar locations (2 filter, 2 language settings) to follow the search rules Jill had listed
- updated all the search useCases to use DI interfaces/repositories
- existing useCases for the tool filter selection pages were deprecated and not functioning, so I removed them and created "view" useCases instead


godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/DownloadableLanguagesViewModel.swift
```diff
@@ -61,9 +63,16 @@ class DownloadableLanguagesViewModel: ObservableObject {
...
         Publishers.CombineLatest(
-            $downloadableLanguages.eraseToAnyPublisher(),
-            $activeDownloads.eraseToAnyPublisher()
+            $searchText.eraseToAnyPublisher(),
+            $downloadableLanguages.eraseToAnyPublisher()
         )
+        .flatMap({ searchText, downloadableLanguages in
+            
+            return Publishers.CombineLatest(
+                self.searchLanguageInDownloadableLanguagesUseCase.getSearchResultsPublisher(for: searchText, in: downloadableLanguages),
```
levieggertcru, line 72: These are difficult to spot, but we have to be careful to not capture self within a flatMap from our ViewModels.  I noticed it will create a retain cycle.  

searchLanguageInDownloadableLanguagesUseCase is simple since it's injected on init and is already in scope.

I'm wondering if you should move $activeDownloads.eraseToAnyPublisher() outside of the flatMap and include it with the searchText and downloadableLanguages?  Did you want to update display languages anytime activeDownloads is updated?

godtools/App/Features/ToolsFilter/Presentation/ToolFilterLanguageSelectionView/ToolFilterLanguageSelectionViewModel.swift
```diff
@@ -43,22 +45,37 @@ class ToolFilterLanguageSelectionViewModel: ObservableObject {
...
-            )
-            .receive(on: DispatchQueue.main)
-            .assign(to: &$languageSearchResults)
+        Publishers.CombineLatest(
+            $searchText.eraseToAnyPublisher(),
+            $allLanguages.eraseToAnyPublisher()
+        )
+        .flatMap { searchText, allLanguages in
+            
+            return self.searchToolFilterLanguagesUseCase.getSearchResultsPublisher(for: searchText, in: allLanguages)
```
levieggertcru, line 75: Capturing self here in flatMap will cause a retain cycle.  Easily resolved by using searchToolFilterLanguagesUseCase that is in the scope of init.

godtools/App/Features/ToolsFilter/Presentation/ToolFilterLanguageSelectionView/ToolFilterLanguageSelectionViewModel.swift
```diff
@@ -43,22 +45,37 @@ class ToolFilterLanguageSelectionViewModel: ObservableObject {
...
+        
+        $appLanguage.eraseToAnyPublisher()
+            .flatMap { appLanguage in
+                
+                return viewToolFilterLanguagesUseCase.viewPublisher(filteredByCategoryId: selectedCategory.id, translatedInAppLanguage: appLanguage)
+            }
+            .receive(on: DispatchQueue.main)
+            .sink { viewLanguageFiltersDomainModel in
+                
+                self.allLanguages = viewLanguageFiltersDomainModel.languageFilters
```
levieggertcru, line 60: Capturing self here will cause a retain cycle.  Will need to use capture list [weak self].

godtools/App/Features/ToolsFilter/Presentation/ToolFilterCategorySelectionView/ToolFilterCategorySelectionViewModel.swift
```diff
@@ -43,22 +45,37 @@ class ToolFilterCategorySelectionViewModel: ObservableObject {
...
-            )
-            .receive(on: DispatchQueue.main)
-            .assign(to: &$categorySearchResults)
+        Publishers.CombineLatest(
+            $searchText.eraseToAnyPublisher(),
+            $allCategories.eraseToAnyPublisher()
+        )
+        .flatMap { searchText, allCategories in
+            
+            self.searchToolFilterCategoriesUseCase.getSearchResultsPublisher(for: searchText, in: allCategories)
```
levieggertcru, line 75: Capturing self here in flatMap will cause a retain cycle. Easily resolved by using searchToolFilterCategoriesUseCase that is in the scope of init.

godtools/App/Features/ToolsFilter/Presentation/ToolFilterCategorySelectionView/ToolFilterCategorySelectionViewModel.swift
```diff
@@ -43,22 +45,37 @@ class ToolFilterCategorySelectionViewModel: ObservableObject {
...
+        
+        $appLanguage.eraseToAnyPublisher()
+            .flatMap { appLanguage in
+                
+                return viewToolFilterCategoriesUseCase.viewPublisher(filteredByLanguageId: selectedLanguage.id, translatedInAppLanguage: appLanguage)
+            }
+            .receive(on: DispatchQueue.main)
+            .sink { viewCategoryFiltersDomainModel in
+                
+                self.allCategories = viewCategoryFiltersDomainModel.categoryFilters
```
levieggertcru, line 60: Capturing self here will cause a retain cycle. Will need to use capture list [weak self].

godtools/App/Features/Dashboard/Presentation/Tools/ToolsViewModel.swift
```diff
@@ -95,24 +97,17 @@ class ToolsViewModel: ObservableObject {
...
-            .flatMap { userFilters in
+        $appLanguage.eraseToAnyPublisher()
+            .flatMap { appLanguage in
                 
-                return Publishers.CombineLatest(
-                    getToolFilterCategoriesUseCase.getCategoryFilterPublisher(with: userFilters.categoryFilterId),
-                    getToolFilterLanguagesUseCase.getLanguageFilterPublisher(from: userFilters.languageFilterId)
-                )
-                .eraseToAnyPublisher()
+                return self.getUserFiltersUseCase.getUserFiltersPublisher(translatedInAppLanguage: appLanguage)
```
levieggertcru, line 103: Capturing self here in flatMap will cause a retain cycle. Easily resolved by using getUserFiltersUseCase that is in the scope of init.

godtools/App/Features/Dashboard/Presentation/Tools/ToolsViewModel.swift
```diff
@@ -95,24 +97,17 @@ class ToolsViewModel: ObservableObject {
...
-                    self.toolFilterCategorySelectionPublisher.send(categoryFilter)
-                }
-                
-                if let languageFilter = languageFilter {
-                    self.toolFilterLanguageSelectionPublisher.send(languageFilter)
-                }
+            .receive(on: DispatchQueue.main)
+            .sink { userFilters in
+              
+                self.toolFilterCategorySelectionPublisher.send(userFilters.categoryFilter)
```
levieggertcru, line 109: Capturing self within sink will cause a retain cycle. Will need to use capture list [weak self].

godtools/App/Features/ToolsFilter/Presentation/ToolFilterCategorySelectionView/ToolFilterCategorySelectionViewModel.swift
levieggertcru, line 60: Capturing self here will cause a retain cycle. Will need to use capture list [weak self].

godtools/App/Features/ToolsFilter/Presentation/ToolFilterCategorySelectionView/ToolFilterCategorySelectionViewModel.swift
```diff
@@ -43,22 +45,37 @@ class ToolFilterCategorySelectionViewModel: ObservableObject {
...
         self.flowDelegate = flowDelegate
         
+        getCurrentAppLanguageUseCase
+            .getLanguagePublisher()
+            .assign(to: &$appLanguage)
+        
+        $appLanguage.eraseToAnyPublisher()
+            .flatMap { appLanguage in
+                
+                return viewToolFilterCategoriesUseCase.viewPublisher(filteredByLanguageId: selectedLanguage.id, translatedInAppLanguage: appLanguage)
```
levieggertcru, line 55: No retain cycle here since not capturing self on viewToolFilterCategoriesUseCase.

levieggertcru (changes requested):
Hey @rachaelblue I'm liking that this is moving towards use cases with dependency inversion.
I'm about 40% through reviewing this PR and figured I'd push a change request now regarding some areas where ```self``` is being captured in ViewModels.
I didn't realize how easy it is to create a retain cycle in the ViewModel and had gone through and cleaned up quite a few areas where self was captured in some prior pull requests.
Luckily flatMaps we can use useCases within the scope of init so those are easy to fix.
When calling sink we will have to be sure to use a capture list [weak self].

godtools/App/Features/ToolsFilter/Presentation/ToolFilterCategorySelectionView/ToolFilterCategorySelectionViewModel.swift
```diff
@@ -25,14 +25,16 @@ class ToolFilterCategorySelectionViewModel: ObservableObject {
...
     @Published var categorySearchResults: [CategoryFilterDomainModel] = [CategoryFilterDomainModel]()
     
-    init(getToolFilterCategoriesUseCase: GetToolFilterCategoriesUseCase, searchToolFilterCategoriesUseCase: SearchToolFilterCategoriesUseCase, storeUserFiltersUseCase: StoreUserFiltersUseCase, categoryFilterSelectionPublisher: CurrentValueSubject<CategoryFilterDomainModel, Never>, selectedLanguage: LanguageFilterDomainModel, getInterfaceStringInAppLanguageUseCase: GetInterfaceStringInAppLanguageUseCase, getCurrentAppLanguageUseCase: GetCurrentAppLanguageUseCase, viewSearchBarUseCase: ViewSearchBarUseCase, flowDelegate: FlowDelegate?) {
+    init(viewToolFilterCategoriesUseCase: ViewToolFilterCategoriesUseCase, searchToolFilterCategoriesUseCase: SearchToolFilterCategoriesUseCase, storeUserFiltersUseCase: StoreUserFiltersUseCase, categoryFilterSelectionPublisher: CurrentValueSubject<CategoryFilterDomainModel, Never>, selectedLanguage: LanguageFilterDomainModel, getInterfaceStringInAppLanguageUseCase: GetInterfaceStringInAppLanguageUseCase, getCurrentAppLanguageUseCase: GetCurrentAppLanguageUseCase, viewSearchBarUseCase: ViewSearchBarUseCase, flowDelegate: FlowDelegate?) {
```
levieggertcru, line 35: Hey @rachaelblue one more thing.  On the init() for ToolFilterCategorySelectionViewModel and ToolFilterLanguageSelectionViewModel make FlowDelegate non optional to avoid any confusion when initializing those ViewModels since we'd want a value there to process the FlowSteps.

godtools/App/Features/ToolsFilter/Presentation/ToolFilterCategorySelectionView/ToolFilterCategorySelectionViewModel.swift
```diff
@@ -11,7 +11,7 @@ import Combine
 
 class ToolFilterCategorySelectionViewModel: ObservableObject {
```
levieggertcru, line 12: Hey @rachaelblue I just noticed something else in both ToolFilterCategorySelectionViewModel.swift and ToolFilterLanguageSelectionViewModel.swift that could cause problems.  This is another one of those things in Combine we have to look out for.  It might be good to make a running list of these in the ReadMe.

The issue is when storing the selected category and language via publisher and using ```.store(in: &cancellables)``` where cancellables is a reference to the scope of the ViewModel.  Once that ViewModel is deallocated it will terminate any publishers in local cancellables.

godtools/App/Features/ToolsFilter/Presentation/ToolFilterCategorySelectionView/ToolFilterCategorySelectionViewModel.swift
```diff
@@ -25,14 +25,16 @@ class ToolFilterCategorySelectionViewModel: ObservableObject {
     
     let selectedCategory: CategoryFilterDomainModel
```
levieggertcru, line 26: Hey @rachaelblue I wasn't able to get a binding to work in what I had hoped for from ToolsViewModel to the SelectCategory and SelectLanguage ViewModels.  Let's stick with the approach of injecting the Subjects for now. 

The only change requests I'd suggest are updating both the selectedCategory and selectedLanguage to be Published properties and then update those when the category is tapped and language is tapped this way the View state will update.

godtools/App/Features/ToolsFilter/Presentation/ToolFilterCategorySelectionView/ToolFilterCategorySelectionViewModel.swift
rachaelblue: Thanks @levieggertcru, good point.  I added a set `staticCancellables`.  Definitely need to remember to think about scope when storing publishers.

levieggertcru (approved):

PR #2008, opened 2024-03-18
GT-2158 with custom observable object

levieggertcru:
Hey @rachaelblue let me know your thoughts on this approach.  

I was trying to use Binding with the View, but that wasn't working and then I remembered ObservableObject so I thought what if we wrap the selected category and selected language in an ObservableObject and use Published on those.  

This object can be injected into the View so the View can update state when the Published property changes and the one injecting those can still observe changes using ```observer.$value```.

The result of this was ```CombineObservableValue``` that way we can re-use with a generic type.

The main benefit is there isn't the need to create an extra property for updating state.  When injecting a CurrentValueSubject or AnyPublisher into a ViewModel, I think we end up having to create an additional Published property on the ViewModel to update the state of the View when those publishers update.

I still wish there were a way to somehow pass Published from a ViewModel as Binding to a View or Published to a ViewModel, but I'm not sure if that's possible with Published properties.


godtools/App/Features/ToolsFilter/Presentation/ToolFilterLanguageSelectionView/ToolFilterLanguageSelectionView.swift
```diff
@@ -13,10 +13,12 @@ struct ToolFilterLanguageSelectionView: View {
     private static let ultraLightGrey = Color.getColorWithRGB(red: 246, green: 246, blue: 246, opacity: 1)
     
     @ObservedObject private var viewModel: ToolFilterLanguageSelectionViewModel
-    
-    init(viewModel: ToolFilterLanguageSelectionViewModel) {
+    @ObservedObject private var selectedLanguage: CombineObservableValue<LanguageFilterDomainModel>
```
rachaelblue, line 16: Does `CombineObservableValue` only trigger changes if it's hooked up to a view?  Or would it be possible for it to live in the viewModel?  If it has to live in the view, I'm wondering what the difference is between this and using `@Published`/`@Binding` in tandem.

rachaelblue: For example, `DashboardViewModel` has a property `@Published var currentTab: Int` that gets passed into `DashboardTabBarItemView` as a `@Binding`, and the tab value can get updated from either end.  Seems kind of like the same idea?

levieggertcru: Hey @rachaelblue now that you mention it I think ViewModel would be more appropriate for MVVM.  Is that what you were thinking?

levieggertcru: Hmm I guess Binding would also work on the View with Published?  In that case CombineObservableValue wouldn't be necessary on the View.  Is that what you were thinking here?

levieggertcru: That's right Binding would be a better option and redundant to use CombineObservableValue.  Let me try Binding again, I remember I did get an error, just can't exactly remember what it stated.  I believe it had to do with scope since I was passing it through the Flow.

rachaelblue: Yeah, I was thinking as far as MVVM, the selectedLanguage should probably live in the viewModel.  I'm pretty sure that was my big issue in the beginning, that I didn't want to pass the value separately into the view.  Although I don't remember dealing with it needing to pass through the flow 🤔

levieggertcru: In my test I was passing CombineObservableValue from ToolsViewModel through the Flow to the Selection Views.

Let me try another approach using Binding on the Selection views and a Published on the ViewModel.  I'm curious what that might look like.

PR #2009, opened 2024-03-18
GT Add notes on UseCase best practices

levieggertcru:
Hey @rachaelblue updated the ReadMe to include those notes on Use Cases.

I'm also in the works of updating the ReadMe for dependency inversion (https://github.com/CruGlobal/godtools-swift/tree/GT-update-readme-for-dependency-inversion).


rachaelblue (approved):
Thanks @levieggertcru, this is helpful!

PR #2010, opened 2024-03-18
GT update readme for dependency inversion

levieggertcru:
Hey @rachaelblue made some updates to the ReadMe (https://github.com/CruGlobal/godtools-swift/tree/GT-update-readme-for-dependency-inversion) for dependency inversion.  I will need to add a little more to the folder structuring again and will create a ticket for that.
Let me know if anything doesn't make sense or anything I can include.


README.md
```diff
@@ -11,168 +13,170 @@ GodTools
...
+#### Conventions
 
-UseCase Best Practices:
+- [Classes](#classes)
+- [Project Folder Structure](#project-folder-structure)
 
-- Should have a single exposed method (public, internal) that takes zero or more inputs and produces a single output that is an AnyPublisher. 
-- UseCases can have private methods, however, as we move to dependency inversion I think private methods will become less and less.
-- Inputs should not be publisher types. Instead the ViewModel should react to changes which then triggers the UseCase.
-- Should not depend on UseCases. If specific Output from a UseCase is needed, instead inject that Output as an Input from the ViewModel.
```
rachaelblue, line 129: This line is a little confusing.  Maybe "If specific Output from a _different_ UseCase is needed,  ..."

README.md
```diff
@@ -11,168 +13,170 @@ GodTools
...
-- Should depend only on interfaces. Most of the time we depend on some type of Repository Interface where a Repository is simply a data storage and data access.
-- Would prefer that UseCases return a non Swift type and instead some type of DomainModel that encapsulates attributes related to the business requirements.
+##### Classes
 
-Responsibilities:
+- Class / Struct attributes should always be declared with the type.
+- Class level attributes should also be grouped by constant vs instance as well as by access level and mutability. Where constants are at the top and grouped by private first, then internal, then public.  Instance variables follow constants and are grouped by private first, then internal, then public.
```
rachaelblue, line 142: It might be helpful to have a screenshot of grouped attributes as an example.  It makes sense, but I think it's clearer in a picture than reading (or accidentally misreading) a list of the ordering.

rachaelblue (commented):
This is great to have in here.  Thanks @levieggertcru!

README.md
levieggertcru: Thanks @rachaelblue for the feedback.  I will be updating these areas.

levieggertcru: Hey @rachaelblue just updated both of those areas.  Please let me know if you have any additional comments or concerns or suggestions. 
For the attribute grouping I don't want to get too nit picky in this area.  Just as long as we have some consistency that is close to what is outlined in the ReadMe.  The main idea is that code is consistent which can be helpful when viewing other people's code in a project.  
I also realize there may be cases for attributes that aren't outlined in the ReadMe, example fileprivate and more special attributes such as Binding, State, etc.  As long as we hit the 3 main groupings static, instance, special, and from there do the best we can to fill in the inner groupings.

ReadMeAssets/attribute_grouping.png
rachaelblue, line 1: This is helpful to see it like this.  Thanks!

rachaelblue (approved):
Looks great, thanks for doing this @levieggertcru!

PR #2011, opened 2024-03-19
GT-2158 binding selected category and selected language

levieggertcru:
Hey @rachaelblue here's an example using Binding.  I prefer this approach to using the custom ```CombineObservableValue```.  I really don't want to introduce another type for SwiftUI and Combine.

The trick was I had to move your Button handling inside the Row views.  This way the selectedCategory and selectedLanguage could be set to the Row's category and language value.  This allows isSelected to also be handled within the Row views now that it is Binding to the selectedCategory and selectedLanguage.

Then in the ViewModel I added a sink on the Published properties for selectedCategory and selectedLanguage to update the CurrentValueSubject.


levieggertcru:
Also note in the ViewModels I did change the rowTapped method to categoryTapped and languageTapped to be a little more descriptive on the inputs.  The ViewModel really doesn't know the UI elements that make up the View, it just exposes what the expected input should be.

levieggertcru:
Actually, now I think we can make the categoryTapped and languageTapped private in the ViewModel since those inputs are through Binding.  Let me update that.

PR #2017, opened 2024-03-22
GT-2298 tools filter remove deprecations

rachaelblue:
Removed `GetInterfaceStringInAppLanguageUseCase` from the 2 filter selection pages and added repos to get the interfaceStrings in each.  All the other tasks listed in the ticket were completed in a previous PR.


levieggertcru (approved):
Thanks @rachaelblue for removing those deprecations.

PR #2022, opened 2024-03-25
GT-2273 language category filter fixes

rachaelblue:
Fixed some minor things Jill mentioned on the ticket: https://jira.cru.org/browse/GT-2273

Also updated `GetTranslatedToolLanguageAvailability` to allow it to take in two languages-- one is the language we're checking the availability of, and the other is the language the string will be translated into.  On lessons cards, it'll be the same language for both (app language), but on tools it'll reflect the language filter selection.


godtools/App/Features/Dashboard/Presentation/Tools/Subviews/ToolCard/ToolCardViewModel.swift
```diff
@@ -35,7 +35,7 @@ class ToolCardViewModel: ObservableObject {
                         
         name = tool.name
         category = tool.category
-        languageAvailability = tool.languageAvailability.availabilityString
+        languageAvailability = tool.languageAvailability?.availabilityString ?? ""
```
levieggertcru, line 38: Hey @rachaelblue as we move to dependency inversion and implementing business logic in the interface implementations, I do want to avoid adding fallback behavior like this in the ViewModel and instead keep logic like this strictly in the interface implementations.  
With languageAvailability being optional I'd suggest updating the ViewModel's Published property to be optional as well, this way it can directly consume data provided by the domain layer and we get the benefit that the View can also optionally render the language availability.

godtools/App/Features/Favorites/Data-DomainInterface/GetYourFavoritedToolsRepository.swift
```diff
@@ -16,16 +16,14 @@ class GetYourFavoritedToolsRepository: GetYourFavoritedToolsRepositoryInterface
     private let getTranslatedToolName: GetTranslatedToolName
     private let getTranslatedToolCategory: GetTranslatedToolCategory
     private let getToolListItemInterfaceStringsRepository: GetToolListItemInterfaceStringsRepository
-    private let getTranslatedToolLanguageAvailability: GetTranslatedToolLanguageAvailability
```
levieggertcru, line 19: Good call on removing the dependency, I think it aids in the visibility that favorited tools won't support language availability.

godtools/App/Features/Dashboard/Presentation/Tools/ToolsViewModel.swift
```diff
@@ -73,29 +73,34 @@ class ToolsViewModel: ObservableObject {
...
-                self?.spotlightTools = spotlightTools
+        Publishers.CombineLatest(
+            $appLanguage.eraseToAnyPublisher(),
+            toolFilterLanguageSelectionPublisher
+        )
+        .flatMap({ (appLanguage: AppLanguageDomainModel, toolFilterLanguage) -> AnyPublisher<(ViewToolsDomainModel, [SpotlightToolListItemDomainModel]), Never> in
+            
+            let selectedLanguageFilter = toolFilterLanguage.language
+            
+            return Publishers.CombineLatest(
```
levieggertcru, line 84: Hey @rachaelblue don't worry about making this change, but I just noticed the CombineLatest within the flatMap and I know we have this similar scenario throughout the app when the app language changes.
I'll need to test this to see what happens when Tools are updated in the RealmDatabase.  I don't think we'd get the update, I'll need to double check.  Shoot I didn't think about this.

levieggertcru (changes requested):
Hey @rachaelblue this looks good!

Just one change request on making the Published property optional on the ViewModel as well for the language availability.

godtools/App/Features/Dashboard/Presentation/Tools/ToolsViewModel.swift
rachaelblue: Yeah good question.  It seems like it should update since it's in the pipeline.  But also, I'm wondering if it might not retain those publishers inside the flatMap when we don't have a reference to them elsewhere.

godtools/App/Features/Dashboard/Presentation/Tools/Subviews/ToolCard/ToolCardViewModel.swift
rachaelblue: Ok yeah good point, that makes sense!  Will look out for that in the future as well.

godtools/App/Features/Dashboard/Presentation/Tools/ToolsViewModel.swift
levieggertcru: Yeah that could be because there isn't a reference.  Will do some testing today.

levieggertcru (approved):

levieggertcru (approved):

godtools/App/Features/Dashboard/Presentation/Tools/ToolsViewModel.swift
levieggertcru: Hey @rachaelblue just tested this by setting some breakpoints and manually triggering pull to refresh after the page had completed loading everything.  
Once the sync stored data in RealmDatabase the GetToolsRepository publisher fired again which fired the ViewToolsUseCase which then triggered in the ToolsViewModel sink.  So it's working which is good news.  
I also tested commenting out the UseCase references on the ViewModel and publishers were firing. 

In this scenario both the ViewToolsUseCase and GetToolsRepository implementations immediately return a Publishers.CombineLatest where the GetToolsRepository Publishers.CombineLatest triggers on database changes. 

I remember in the scenario that failed to trigger I had the Publishers.CombineLatest in the middle of a flatMap of a UseCase.  I wonder if it had to do something with it.  I'll see if I can test that scenario again.

rachaelblue: Ok nice!  Thanks for testing this.  Would be curious to see if there are edge cases to watch out for.

PR #2034, opened 2024-03-29
GT-2116 increase search bar tap area

rachaelblue:
Hey @levieggertcru, this was simple enough to accomplish using the `FocusState` property wrapper, which is really cool!  But it's only available on iOS 15+, and it looked like maybe the only possible way to accomplish the same thing on iOS 14 was to use `UIViewRepresentable`, and it seemed like a lot more effort than it would be worth.  So I copied the existing SearchBar implementation into a legacy version without `FocusState` and made the changes only for iOS 15+, which also allowed me to clean up several other things in the new version.

Also fixed some glitchiness with the "Cancel" button--  the SearchBarViewModel was getting re-initialized every time the view changed, so just needed to keep a reference to it.


levieggertcru (approved):
Thanks @rachaelblue the search is working really smooth.  Much better tap area too.

- FocusState looks really nice and a simple way to change where focus is at. 
- I agree, creating the UIViewRepresentable is definitely not worth the effort.  It seems now iOS 15+ has ironed out quite a bit.
- And yeah that does make sense about the sub ViewModel.  I've ran into that issue before when re-allocating subviews and sometimes you need to keep a stored reference on the parent ViewModel.

PR #2037, opened 2024-04-01
GT-2235 Use code instead of languageCode

levieggertcru:
The downloadable languages list was missing some languages and realized because ResourcesFilter languageCode was taking the LanguageModel languageCode (equivalent of locale language code) which is different than LanguageModel code (equivalent of locale identifier).  It's a bit confusing and I may need to look into a way to avoid that confusion.


rachaelblue (approved):
Thanks @levieggertcru, I'm writing that down about code vs languageCode. 😅 I've definitely mixed them up before (also localeId, languageId, and just id in some models), but usually nothing shows up at all so I realize the mistake.  It might be helpful to have explanations about the differences in the readme as well, also to make sure we're consistent.

levieggertcru:
Yeah it is confusing, and that specific case is in relation to the LanguageModel attributes (id and code) from the mobile-content-api which make it a bit more confusing.

PR #2040, opened 2024-04-03
GT-2294 Fix CYOA navigation

levieggertcru:
This PR makes some changes to renderer navigation and attempts to fix navigation for CYOA.  
- Moved CYOA navigation into ChooseYourOwnAdventureViewModel.swift and override methods ```getInitialPages(``` and ```getPageNavigationEvent(```.
- Updated navigation in base MobileContentPagesViewModel.swift to handle navigation for visible and hidden pages.
- Removed signal for removing pages in MobileContentPagesViewModel.swift.  Instead inserting and removing pages should happen when creating the MobileContentPagesNavigationEvent in ```getPageNavigationEvent(```.


godtools/App/Features/Dashboard/Presentation/ChooseYourOwnAdventure/ChooseYourOwnAdventureViewModel.swift
```diff
@@ -59,93 +59,112 @@ class ChooseYourOwnAdventureViewModel: MobileContentPagesViewModel {
...
+    override func getInitialPages(pageRenderer: MobileContentPageRenderer) -> [Page] {
+            
+        let allPages: [Page] = pageRenderer.getAllPageModels()
+        
+        if let introPage = allPages.first(where: {$0.id == "intro"}) {
+            return [introPage]
+        }
+        else if let firstPage = allPages.first {
+            return [firstPage]
+        }
```
frett, line 74: this logic should be the first page that isn't hidden

levieggertcru: Good call thanks @frett.

PR #2041, opened 2024-04-04
GT-2220 user filters update for realm

rachaelblue:


godtools/App/Features/ToolsFilter/Presentation/ToolFilterLanguageSelectionView/ToolFilterLanguageSelectionViewModel.swift
```diff
@@ -50,7 +48,21 @@ class ToolFilterLanguageSelectionViewModel: ObservableObject {
...
+            }
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] userFilters in
+            
+                self?.selectedLanguage = userFilters.languageFilter
+                self?.selectedCategory = userFilters.categoryFilter
+            }
+            .store(in: &cancellables)
+        
+        $appLanguage.eraseToAnyPublisher()
```
levieggertcru, line 62: Hey @rachaelblue it looks like here you will want to react to selected category changes as well to push into the use case.

godtools/App/Features/ToolsFilter/Presentation/ToolFilterCategorySelectionView/ToolFilterCategorySelectionViewModel.swift
```diff
@@ -50,7 +49,21 @@ class ToolFilterCategorySelectionViewModel: ObservableObject {
...
+            }
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] userFilters in
+            
+                self?.selectedLanguage = userFilters.languageFilter
+                self?.selectedCategory = userFilters.categoryFilter
+            }
+            .store(in: &cancellables)
+        
+        $appLanguage.eraseToAnyPublisher()
```
levieggertcru, line 63: Same here you'll want to react to selected language changes to push into the use case.

levieggertcru (changes requested):
Thanks @rachaelblue for updating this to use Realm.  I also see you were able to observe realm changes instead of going through the flow.  Either method is good.  

Just 2 similar change requests I noticed where you will need to observe the selected category and selected language changes to push to the use case.

godtools/App/Features/ToolsFilter/Presentation/ToolFilterLanguageSelectionView/ToolFilterLanguageSelectionViewModel.swift
rachaelblue: Hey @levieggertcru, the category won't get changed while you're on the language selection page.  You'd have to back out of the languages, change the category, and then reopen it.  So I don't think the category publisher would ever fire.  Same with the category page and language selection.

godtools/App/Features/ToolsFilter/Presentation/ToolFilterLanguageSelectionView/ToolFilterLanguageSelectionViewModel.swift
```diff
@@ -50,7 +48,21 @@ class ToolFilterLanguageSelectionViewModel: ObservableObject {
...
                 
-                return viewToolFilterLanguagesUseCase.viewPublisher(filteredByCategoryId: selectedCategory.id, translatedInAppLanguage: appLanguage)
+                return getUserToolFiltersUseCase.getUserToolFiltersPublisher(translatedInAppLanguage: appLanguage)
+                    .eraseToAnyPublisher()
+            }
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] userFilters in
+            
+                self?.selectedLanguage = userFilters.languageFilter
+                self?.selectedCategory = userFilters.categoryFilter
```
levieggertcru, line 58: Hey @rachaelblue I do see the selectedCategory is declared as mutable and is being changed here, so there is the potential that flatMap is called before selectedCategory is set.

rachaelblue: Good point.  Thanks!

levieggertcru (approved):

PR #2043, opened 2024-04-04
GT Add some behavior tests for tools category filter searching

levieggertcru:
Hey @rachaelblue this is my thinking on behavior tests for our business logic (domain interface implementations).  

This example includes tests for tools category filter searching.

These tests would mostly live in ```/godtoolsTests/App/Features/FeatureName/Data-DomainInterface/*``` to line up with the godtools target folder structure.

I will probably also include tests for the ```/App/Share/Data-DomainInterface/Supporting/*```.

This means for mocking the data layer dependencies we would need to either:

1. Create an interface to mock against.  One example of this is LocalizationServices, where GetOnboardingTutorialInterfaceStringsRepositoryTests.swift includes a mock and GetOnboardingTutorialInterfaceStringsRepository.swift points to the LocalizationServicesInterface.

That mock lives in ```/godtoolsTests/App/Share/Data/LocalizationServices/MockLocalizationServices/``` to line up with the godtools target folder structure.

2. Create a mock realm database via TestsInMemoryRealmDatabase and then create TestsDiContainer for fetching data layer classes that depend on a realm database.  An example of this is in GetDownloadToolProgressInterfaceStringsRepositoryTests.swift where I fetch a ResourcesRepository instance since creating an interface would probably be a lot of work.

The tools category filter searching was easy since the test data is injected as an input.

Let me know any thoughts or concerns you see. 

I'm hoping this is a good path and it will make sense as we add more and more tests.


rachaelblue (approved):
Hey @levieggertcru, this looks great!  Awesome how simple it is to mock and test.  I also really like the readability with Quick and Nimble.

PR #2057, opened 2024-04-17
GT-2298 and GT-2349 filter tools by category and language and add behavior tests for tools filtering

levieggertcru:
This PR adds the following:

- In ViewToolsUseCase add inputs for filter by category and filter by language.
- In GetToolsRepository add inputs for filter by category and filter by language.
- Adds Behavior Tests (GetToolsRepositoryTests) for filtering by category and filtering by language.
- In ResourcesRepository add getAllToolsListCount method and use this for fetching number of tools available. We gain some performance since there isn't any mapping to ResourceModel types.


godtools/App/Features/Dashboard/Presentation/Tools/ToolsViewModel.swift
```diff
@@ -67,34 +67,45 @@ class ToolsViewModel: ObservableObject {
...
-            self?.toolSpotlightSubtitle = domainModel.interfaceStrings.toolSpotlightSubtitle
-            self?.filterTitle = domainModel.interfaceStrings.filterTitle
-            
-            self?.spotlightTools = spotlightTools
-            
-            self?.allTools = domainModel.tools
-            self?.isLoadingAllTools = false
-        }
-        .store(in: &cancellables)
+                return getUserToolFiltersUseCase.getUserToolFiltersPublisher(translatedInAppLanguage: appLanguage)
```
rachaelblue, line 73: Just wondering, did you change this to call the `getUserToolFiltersPublisher` directly rather than listening for updates from `$toolFilterLanguageSelection.eraseToAnyPublisher()` because it'll likely use the dummy initial value the first time this chain evaluates?

rachaelblue (approved):
Nice tests!

godtools/App/Features/Dashboard/Presentation/Tools/ToolsViewModel.swift
levieggertcru: Hey @rachaelblue no, I would have left the ```$toolFilterLanguageSelection.eraseToAnyPublisher()```, but realized I also needed the category filter to push into the UseCase which is what caused this change. 

I'm glad you mentioned this because now that I think about it, it would be better to create a private variable for the toolFilterCategorySelection and include that rather than calling the use case again. Going to update this.

levieggertcru: That's another good point about the dummy initial value. I would like to keep that in the business logic, hmm.  The app language is also set to english by default.  Maybe I should just make those empty string values.

levieggertcru: 1 more change, I went ahead and set the category and language filter selection text to empty strings just for the sake of keeping as much as we can in the business logic.

rachaelblue (approved):

PR #2072, opened 2024-04-25
GT-2341 get distinct categories from realm

rachaelblue:
This was super nice!  Will have to remember that Realm feature.


rachaelblue:
Hey @levieggertcru, I made that change here.  It's not quite as nice since `distinct` doesn't make sense with an array of languages, but at least removes the overhead of mapping to ResourceModels.

levieggertcru:
Thanks @rachaelblue for looking into that and updating to exclude the mapping to ResourceModels.  Should get a bit of a performance boost :D

levieggertcru (approved):

PR #2148, opened 2024-05-29
GT-2266 default tool language

rachaelblue:
Hey @levieggertcru, it took me a while to wrap my head around this, especially because I kept swapping language ID for language code 😅 so let me know if I'm missing anything here.

To test:
Set app language to anything besides Vietnamese, go into tool details for "The Four" and favorite the Vietnamese version.  When you tap back out onto the favorites page, you'll see the tool in Vietnamese and be able to open it in Vietnamese.


godtools/App/Share/Data-DomainInterface/Supporting/GetTranslatedToolCategory/GetTranslatedToolCategory.swift
```diff
@@ -29,8 +29,19 @@ class GetTranslatedToolCategory {
...
     func getTranslatedCategory(resource: ResourceModel, translateInLanguage: BCP47LanguageIdentifier) -> String {
-                
-        let category: String = localizationServices.stringForLocaleElseEnglish(localeIdentifier: translateInLanguage, key: "tool_category_\(resource.attrCategory)")
+        
+        let localeId: String
+        
+        if resource.supportsLanguage(languageId: translateInLanguage) {
+            localeId = translateInLanguage
+        } else {
+            localeId = resource.attrDefaultLocale
```
levieggertcru, line 38: Hey @rachaelblue this looks right for the category to fallback to the default locale.

godtools/App/Share/Domain/UseCases/GetToolTranslationsFilesUseCase/DetermineToolTranslationsToDownload/DetermineToolTranslationsToDownload.swift
```diff
@@ -46,8 +46,9 @@ class DetermineToolTranslationsToDownload: DetermineToolTranslationsToDownloadTy
             translations.append(translation)
         }
         
-        if translations.isEmpty, let englishTranslation = translationsRepository.getLatestTranslation(resourceId: resourceId, languageCode: LanguageCodeDomainModel.english.value) {
-            translations = [englishTranslation]
+        if translations.isEmpty, let defaultTranslation = translationsRepository.getLatestTranslation(resourceId: resourceId, languageCode: resource.attrDefaultLocale) {
+            
+            translations = [defaultTranslation]
```
levieggertcru, line 51: I think this is a good place to provide the default translation if all else fails.

godtools/App/Features/ToolDetails/Data-DomainInterface/GetToolDetailsRepository.swift
```diff
@@ -47,8 +47,17 @@ class GetToolDetailsRepository: GetToolDetailsRepositoryInterface {
...
-        guard let translation = translationsRepository.getLatestTranslation(resourceId: toolId, languageCode: translateInLanguage) else {
+    
+        let translation: TranslationModel
+        if let appLanguagetranslation = translationsRepository.getLatestTranslation(resourceId: toolId, languageCode: translateInLanguage) {
+            
+            translation = appLanguagetranslation
+        }
+        else if let defaultTranslation = translationsRepository.getLatestTranslation(resourceId: toolId, languageCode: toolDataModel.attrDefaultLocale) {
+            
+            translation = defaultTranslation
```
levieggertcru, line 58: Hey @rachaelblue I'm not 100% certain on this here.  The Jira task wasn't very clear on what needed to display in the default locale vs app language.

I noticed the about description text is displayed in the default locale and the versions description text is in the app language.  I wonder if that is intended?

Also the tool name looks to be displaying in English in the version card.  There could be a bug here and I'm curious what the language translation requirements are for the version card interface strings.  I'm assuming it would be similar to the tool lists, the name and description would display in the app language if supported, otherwise fallback to the default locale.

tool-details-2

tool-details-3

rachaelblue: Hey @levieggertcru, Jill confirmed that only non-interface strings should be translated into the default language (she circled in red).  On the versions tab, the description is an interface string, so it should be in app language.
img_4238_720

I did update the translation logic on the versions tool card title since it was still defaulting to English.

levieggertcru (approved):

PR #2164, opened 2024-06-12
GT-2313 persist language choice for favorites

rachaelblue:
Hey @levieggertcru, I went back and forth on a couple things here because tool language choices should only be persisted/applied when you're managing tools on the Favorites tab itself (favorited tools on the Tools tab won't behave like that).  Ended up observing the current dashboard tab to help with this in spots where it's unclear if the flow is from favorites or not (like when you select tool details first, then open the tool).  Let me know if this makes sense.


levieggertcru:
Hey @rachaelblue that makes sense.  Probably not as clear, but you can also read the current tab directly from the DashboardView like so: ```getDashboardInNavigationStack()?.rootView.getCurrentTab()```.  Wouldn't have to hold an additional reference then.

rachaelblue:
Ok @levieggertcru I made a new feature `PersistUserToolLanguageSettings` and added a default value of false to the `shouldPersistToolSettings` flag.  Also made a followup ticket https://jira.cru.org/browse/GT-2411 as we discussed.

levieggertcru (approved):
Thanks @rachaelblue, I think pulling that out into a feature definitely helps in visibility.  
And that created ticket would also be good to get in at some point.

PR #2182, opened 2024-07-05
GT-2323 change dependency direction

levieggertcru:
This PR fixes a dependency direction where a data layer class was pointing to a business logic class.  

I had created a TranslatedLanguageNameRepository (data layer) and that was pointing to GetTranslatedLanguageName (business logic).  

TranslatedLanguageNameRepository was created to persist translated language names since there was a performance hit in our lists of languages that needed translating.  The performance hit was due to the fact that every localization file would be loaded for every language that needed translating.

I ended up changing TranslatedLanguageNameRepository to LocalizationLanguageNameRepository because it should really only deal with localized language names from one sky and I ended up simplifying the logic for now to only check against languages that are supported as one sky translations (https://github.com/CruGlobal/godtools-swift/blob/GT-2323-change-dependency-direction/godtools/App/Share/Data/LocalizationLanguageNameRepository/LocalizationLanguageNameRepository.swift#L14) 

To fix the dependency direction LocalizationLanguageNameRepository (data layer) will now be injected into GetTranslatedLanguageName (business logic).


gyasistory (dismissed):
Cursor_and_iPhone_15_Pro


There is an issue some ot the languages are not visible.

levieggertcru:
Thanks @gyasistory can you try now.  Fallback should be coming in.

gyasistory:
> Thanks @gyasistory can you try now. Fallback should be coming in.

Looks good to me see if you get @rachaelblue to take a look at it before you merge it.

rachaelblue (approved):

PR #2200, opened 2024-07-11
convert to language ID to check if language is supported

rachaelblue:
See Jill's comment on https://jira.cru.org/browse/GT-2266 about the category not translating correctly.

The existing logic was failing because a `BCP47LanguageIdentifier` was being passed into `resource.supportsLanguage(languageId:)`.  But `languageId` isn't equivalent to `BCP47LanguageIdentifier`.  But  `BCP47LanguageIdentifier` is equivalent to `languageCode` and `localeId`, right?  I feel like we need a better way to convert/standardize this.

### DIFF
--- a/godtools/App/DependencyContainer/AppDataLayerDependencies.swift
+++ b/godtools/App/DependencyContainer/AppDataLayerDependencies.swift
@@ -314,6 +314,7 @@ class AppDataLayerDependencies {
     
     func getTranslatedToolCategory() -> GetTranslatedToolCategory {
         return GetTranslatedToolCategory(
+            languagesRepository: getLanguagesRepository(), 
             localizationServices: getLocalizationServices(),
             resourcesRepository: getResourcesRepository()
         )
--- a/godtools/App/Share/Data-DomainInterface/Supporting/GetTranslatedToolCategory/GetTranslatedToolCategory.swift
+++ b/godtools/App/Share/Data-DomainInterface/Supporting/GetTranslatedToolCategory/GetTranslatedToolCategory.swift
@@ -11,11 +11,13 @@ import LocalizationServices
 
 class GetTranslatedToolCategory {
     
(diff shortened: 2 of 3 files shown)


levieggertcru:
Good morning @rachaelblue.  I agree it can be confusing.  One thing that helps is to keep in mind which object is being worked with.  When it comes to the mobile-content-api objects (ResourceModel, LanguageModel, TranslationModel, etc.), id's are the primaryKey tracked in the remote database.  So the LanguageModel id is the primaryKey and in this context languageId is the LanguageModel id attribute.

For me it was getting confusing with the LanguageModel code attribute since language code on Locale is different than the identifier on Locale. On LanguageModel the code attribute is the same thing as the Locale Identifier (BCP47LanguageIdentifier).  So here the type changed to a typealias BCP47LanguageIdentifier to help clarify that.

Again though, I think it can still be confusing when it comes to argument labels of type String.  We may have to create a type that wraps the value and pass those instead.  And maybe changing BCP47LanguageIdentifier to a struct would be helpful since we would get some extra help from the compiler when a plain string is passed on an argument of type BCP47LanguageIdentifier.

levieggertcru:
I'm thinking we could do the same for the mobile-content-api object's id attribute.  Wrap that in a struct.

levieggertcru (approved):
I changed the type for the code attribute in LanguagesRepository from String to BCP47LanguageIdentifier.  However, since BCP47LanguageIdentifier is a typealias the compiler wouldn't complain if a plain string were passed.  I think it would be good at some point to change BCP47LanguageIdentifier to a struct type.

rachaelblue:
Thanks @levieggertcru.  I made a followup ticket to make that change to BCP47LanguageIdentifier: https://jira.cru.org/browse/GT-2415

levieggertcru:
Thanks @rachaelblue for creating that.

PR #2201, opened 2024-07-12
GT-2405 lesson language filter tests

levieggertcru:


levieggertcru:
Hey @rachaelblue I started adding some behavior tests for the lessons filtering and I hit a interesting edge case that would be good to get clarification on.  Basically if the current app language doesn't have any lessons, then instead of the filter displaying the current app language, it will come back null.  Would be good to have Jill clarify this edge case as to if we still want to display the app language.

It will return nil on this line here (https://github.com/CruGlobal/godtools-swift/blob/GT-2405-Lessons-Language-Filter/godtools/App/Features/LessonFilter/Data-DomainInterface/GetLessonFilterLanguagesRepository.swift#L86).

This test here will test that behavior by setting the app language to french and there are 0 french tools in the test. (https://github.com/CruGlobal/godtools-swift/blob/GT-2405-lesson-language-filter-tests/godtoolsTests/App/Features/LessonFilter/Data-DomainInterface/GetUserLessonFiltersRepositoryTests.swift#L78)

levieggertcru:
lessons-filter-figma

levieggertcru:
NOTE: @levieggertcru @rachaelblue 
Will need to merge changes in https://github.com/CruGlobal/godtools-swift/pull/2205 in order to take advantage of the shared in memory realm instance for these tests.

PR #2205, opened 2024-07-17
GT user counter unit tests realm instance creator

levieggertcru:


levieggertcru:
Hey @rachaelblue below are a couple comments from realm's documentation regarding in memory realms.  

I'm thinking I will update the TestsInMemoryRealmDatabase to make sure it sends ```usesASingleSharedRealmInstance```.

"You can open a realm entirely in memory, which will not create a .realm file or its associated [auxiliary files](https://www.mongodb.com/docs/atlas/device-sdks/sdk/swift/realm-files/#std-label-ios-realm-file). Instead the SDK stores objects in memory while the realm is open and discards them immediately when all instances are closed."

"IMPORTANT
When all in-memory realm instances with a particular identifier go out of scope, Realm deletes all data in that realm. To avoid this, hold onto a strong reference to any in-memory realms during your app's lifetime."

(https://www.mongodb.com/docs/atlas/device-sdks/sdk/swift/realm-files/configure-and-open-a-realm/#open-an-in-memory-realm)

rachaelblue (approved):
Hey @levieggertcru, that makes sense that the strong reference is needed when in memory.  Thanks for looking into that and fixing this up!

PR #2227, opened 2024-07-26
GT-2350 filter category and language interface

rachaelblue:


godtools/App/Share/Common/StringSearcher/StringSearcher.swift
```diff
@@ -24,6 +24,19 @@ class StringSearcher {
         }
     }
     
+    func search(for searchText: String, in searchables: [StringSearchable]) -> [StringSearchable] {
```
levieggertcru, line 27: I'm curious how to best resolve StringSearcher now.  It would be best to have a single method rather than the 2 methods.

I'm wondering was it a bad design and suggestion by me to have an interface for ToolFilterCategoryDomainModel?

On StringSearcher I lean more towards having method ```func search<T: StringSearchable>(for searchText: String, in searchables: [T]) -> [T]``` rather than ```func search(for searchText: String, in searchables: [StringSearchable]) -> [StringSearchable]``` because of the generics.

levieggertcru: Sorry @rachaelblue, I think it was a bad idea to suggest an interface. 

Did some searching on when to use inheritance vs an interface and found a helpful article.  To me ```Any``` looks like a specialized version of the filter option.   What are your thoughts on that?

Screenshot 2024-07-26 at 11 21 54 AM

rachaelblue:
Hey @levieggertcru, removed those two interfaces and made the "any" versions subclasses.  Seeing how small the subclass is made me wonder if we even needed the subclass at all, they're really just initialized differently and that's the only difference.  Maybe nice to have it for clarity more than anything.  Let me know what you think.

godtools/App/Features/Dashboard/Presentation/Tools/Models/ToolFilterSelection.swift
```diff
@@ -10,6 +10,6 @@ import Foundation
 
 struct ToolFilterSelection {
```
levieggertcru, line 11: Hey @rachaelblue looks like this file can be removed?  I don't think it's in use anymore.

levieggertcru:
Hey @rachaelblue that ```Any``` has me thinking about modeling the domain layer in general in terms of categories and languages.  It was a special type since it didn't exist in resources.  For now, I'd say let's just leave it in the domain layer.  It definitely doesn't hurt to have some extra visibility of an ```Any``` category and language in our domain layer.  I'm also using in the tests here (https://github.com/CruGlobal/godtools-swift/pull/2228) when filtering for ```Any``` category and language.

levieggertcru (changes requested):
Thanks @rachaelblue for updating this.  I think the enum type would have been more difficult to maintain if adding and removing properties.
Just one comment on removing that file which I don't think is used anymore?
Also, go ahead and review the tests update and merge that into this branch which should resolve any issues with tests running (https://github.com/CruGlobal/godtools-swift/pull/2228).

levieggertcru (approved):

rachaelblue:
Good catch.  Removed that file and merged in tests @levieggertcru.  Thanks!

PR #2228, opened 2024-07-26
GT-2350 Update tests for new filter types on tools and lessons

levieggertcru:


rachaelblue (approved):
Thanks @levieggertcru for fixing these!

PR #2233, opened 2024-07-29
GT-2405 lesson navigation fix

rachaelblue:
Fixes:
1. When clicked, lessons should open up in the chosen filter language.
2. The lesson title should be written in the filtered language rather than the app language.


levieggertcru (approved):
Thanks Rachael looks good.

PR #2235, opened 2024-07-30
Update dependabot.yml to Add Swift Package Manager

gyasistory:
https://github.com/dependabot/dependabot-core/issues/1245#issuecomment-1650556093

https://github.com/dependabot/dependabot-core/pull/7525

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,13 @@ updates:
   assignees:
   - frett
   - levieggertcru
+- package-ecosystem: swift
+  directory: "/"
+  schedule:
+    interval: daily
+  reviewers:
+  - frett
+  - levieggertcru
+  assignees:
+  - frett
+  - levieggertcru


.github/dependabot.yml
```diff
@@ -10,3 +10,13 @@ updates:
   assignees:
   - frett
   - levieggertcru
+- package-ecosystem: swift
```
frett, line 13: The ticket you linked mentioned that the swift support is only for Package.swift and not XCode.

That's similar support that renovate has which doesn't currently cover apps using spm

frett: relevant dependabot ticket: https://github.com/dependabot/dependabot-core/issues/7694

frett (approved):
I'll approve since I don't think this will hurt anything, but I don't think this SPM support will actually do anything for the godtools-swift repo

.github/dependabot.yml
gyasistory: [Ok, I thought I found something.](https://github.com/dependabot/dependabot-core/pull/7525)

gyasistory: @levieggertcru I will leave it up to you whether or not to merge it in.

PR #2240, opened 2024-08-02
GT-2405 set lesson filter to app language

rachaelblue:


levieggertcru (approved):
Thanks @rachaelblue for making those changes!

PR #2250, opened 2024-08-14
GT-2199 tool options followup

rachaelblue:


godtools/App/Flows/Tract/TractFlow.swift
```diff
@@ -11,12 +11,12 @@ import GodToolsToolParser
...
 import LocalizationServices
 
-class TractFlow: ToolNavigationFlow, Flow {
+class TractFlow: ToolNavigationFlow, ToolSettingsNavigationFlow {
         
     private let appLanguage: AppLanguageDomainModel
     
-    private var toolSettingsFlow: ToolSettingsFlow?
     private var cancellables: Set<AnyCancellable> = Set()
+    internal var toolSettingsFlow: ToolSettingsFlow?
```
levieggertcru, line 19: Hey @rachaelblue just a side-note, I'm okay with omitting ```internal``` since it's the default access level (https://docs.swift.org/swift-book/documentation/the-swift-programming-language/accesscontrol#Default-Access-Levels).  However, being verbose never hurts either.   

One downside to protocol extension's is when there's a need to mutate at attribute in the extension, end up exposing attributes as in this case.

godtools/App/Services/Renderer/Views/MobileContent/Views/Pages/MobileContentPagesViewModel.swift
```diff
@@ -43,6 +43,8 @@ class MobileContentPagesViewModel: NSObject, ObservableObject {
     let pageNavigationEventSignal: SignalValue<MobileContentPagesNavigationEvent> = SignalValue()
     let incrementUserCounterUseCase: IncrementUserCounterUseCase
     
+    var toolSettingsObserver: ToolSettingsObserver?
```
levieggertcru, line 46: Hey @rachaelblue I'd say make this ```private(set) var``` since this is set in the ```setup``` method of the parent class.

levieggertcru (changes requested):
Thanks @rachaelblue for making these changes.  Just the 1 change request on making that toolSettingsObserver read-only.

godtools/App/Flows/Tract/TractFlow.swift
rachaelblue: Good point 🤦‍♀️

rachaelblue:
Thanks @levieggertcru.  I realized the observers weren't getting attached in the right order, so changing the language wasn't updating properly.  Fixed now.

levieggertcru (approved):

PR #2255, opened 2024-08-20
GT-2177 create domain model for highlighted text

rachaelblue:
Branched off of the cyoa feature.

### DIFF
--- a/godtools/App/Features/AppLanguage/Presentation/ConfirmAppLanguage/ConfirmAppLanguageViewModel.swift
+++ b/godtools/App/Features/AppLanguage/Presentation/ConfirmAppLanguage/ConfirmAppLanguageViewModel.swift
@@ -48,19 +48,30 @@ class ConfirmAppLanguageViewModel: ObservableObject {
             .switchToLatest()
             .receive(on: DispatchQueue.main)
             .sink { [weak self] domainModel in
-                
+                guard let self = self else { return }
                 let interfaceStrings = domainModel.interfaceStrings
                 
-                self?.messageInNewlySelectedLanguage = interfaceStrings.messageInNewlySelectedLanguage
-                self?.messageInCurrentLanguage = interfaceStrings.messageInCurrentLanguage
-                self?.changeLanguageButtonTitle = interfaceStrings.changeLanguageButtonText
-                self?.nevermindButtonTitle = interfaceStrings.nevermindButtonText
+                self.messageInNewlySelectedLanguage = getAttributedMessageString(highlightStringDomainModel: interfaceStrings.messageInNewlySelectedLanguageHighlightModel)
+                self.messageInCurrentLanguage = getAttributedMessageString(highlightStringDomainModel: interfaceStrings.messageInCurrentLanguageHighlightModel)
+                self.changeLanguageButtonTitle = interfaceStrings.changeLanguageButtonText
+                self.nevermindButtonTitle = interfaceStrings.nevermindButtonText
             }
             .store(in: &cancellables)
     }
     
     deinit {
         print("x deinit: \(type(of: self))")
+    }
+    
+    private func getAttributedMessageString(highlightStringDomainModel: ConfirmAppLanguageHighlightStringDomainModel) -> NSAttributedString {
+        
+        let formatStringAttributed = NSAttributedString(string: highlightStringDomainModel.formatString)
+        let highlightTextAttributed = NSAttributedString(
+            string: highlightStringDomainModel.highlightText,
+            attributes: [NSAttributedString.Key.foregroundColor: ColorPalette.gtBlue.uiColor]
+        )
+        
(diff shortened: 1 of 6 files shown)


godtools/App/Features/AppLanguage/Presentation/ConfirmAppLanguage/ConfirmAppLanguageViewModel.swift
```diff
@@ -48,20 +48,31 @@ class ConfirmAppLanguageViewModel: ObservableObject {
...
+                self.nevermindButtonTitle = interfaceStrings.nevermindButtonText
             }
             .store(in: &cancellables)
     }
     
     deinit {
         print("x deinit: \(type(of: self))")
     }
+    
+    private func getAttributedMessageString(highlightStringDomainModel: ConfirmAppLanguageHighlightStringDomainModel) -> NSAttributedString {
```
levieggertcru, line 66: Hey @rachaelblue I just realized I had left a task todo in the Jira task which shouldn't have been there which says: 

"We may want some type the ViewModel can expose that encapsulates both the text and text to highlight for the View to observe."  This isn't necessary since this is achieved by the newly created DomainModel ```ConfirmAppLanguageHighlightStringDomainModel```.

To complete this task I would like the ```getAttributedMessageString``` logic pushed directly into the View layer and the ViewModel can instead expose type ```ConfirmAppLanguageHighlightStringDomainModel``` for attributes ```messageInNewlySelectedLanguage``` and ```messageInCurrentLanguage```.  This would then move render logic directly to the responsibility of the View and the View simply observes the domain model ```ConfirmAppLanguageHighlightStringDomainModel```. 

On that note, one other thing I've been slowly shifting towards is exposing domain models directly to the View from the ViewModel.  We have quite a bit of instances where we map attributes from the DomainModel to attributes of the ViewModel which we don't need to do, but I understand that happened because we've slowly shifted toward dependency inversion and domain models.

An example where I expose the interface strings directly from the domain model to the View is here (https://github.com/CruGlobal/godtools-swift/blob/develop/godtools/App/Features/Account/Presentation/DeleteAccount/DeleteAccountViewModel.swift#L20).

In this scenario I did add a static method on the domain model that creates an instance with empty values this way I didn't have to have an optional type in the ViewModel. ```static func emptyStrings() -> DeleteAccountInterfaceStringsDomainModel```
That's the one downside I found to referencing domain models where I don't want an optional.

levieggertcru (changes requested):
Hey @rachaelblue this is looking much better!

I left a final comment to shift the render logic just a bit more over directly into the View layer.  The idea here is that the View could be swapped out and isn't locked into using NSAttributedString from the ViewModel when rendering highlight text.  It could renderer the highlight text in any manner it chooses.

levieggertcru (approved):
Thanks @rachaelblue looks good !

godtools/App/Features/AppLanguage/Presentation/ConfirmAppLanguage/ConfirmAppLanguageViewModel.swift
rachaelblue: Thanks @levieggertcru, sorry I missed the intent here!  That will be interesting moving forward, exposing the domain model directly into the view.  I am curious if that will change the way the view updates at all, when @Published wraps the whole struct rather than single values.

levieggertcru: That's a good point to make.  It would be worth investigating that.  Wondering if we could see any impacts on performance?

levieggertcru: I'm still curious if there would be a difference in how SwiftUI re-renders the view between a single published property pointing to a domain model vs multiple published properties pointing to the inner attributes of the domain model.  

If it's a single published property pointing to a domain model, would the View re-render all inner attributes of the DomainModel that it's referencing?  Regardless if those values have changed or not?

Where as, if there are multiple published properties, I'm wondering if the View would then know to only re-render published properties that have actually changed. 

I'm not quite sure how smart the View is when it comes to re-rendering state changes.

rachaelblue: Some interesting things in this article https://medium.com/@shashidj206/how-to-avoid-repeating-swiftui-view-updates-ec1fce0349a9.  They said if the model conforms to `Equatable` and `Hashable`, SwiftUI will know not to update unless the value changes.

levieggertcru: Thanks Rachael, I'm planning to look into that article more.

PR #2256, opened 2024-08-22
GT-2430 disable share links on cyoa

rachaelblue:


godtools/App/Flows/ToolSettings/Observer/Protocols/LinkShareable.swift
```diff
@@ -0,0 +1,13 @@
...
+//  LinkShareable.swift
+//  godtools
+//
+//  Created by Rachael Skeath on 8/20/24.
+//  Copyright © 2024 Cru. All rights reserved.
+//
+
+import Foundation
+
+protocol LinkShareable {
```
levieggertcru, line 11: Hey @rachaelblue these changes are looking good.  The View fix looks good too.  

I'm wondering if we should think about adding the dependencies needed into the protocols?  Going to think about that a bit.

godtools/App/Features/ToolSettings/Presentation/ToolSettings/Subviews/ToolSettingsOptions/ToolSettingsOptionsView.swift
```diff
@@ -28,15 +28,18 @@ struct ToolSettingsOptionsView: View {
             ScrollView(.horizontal, showsIndicators: false) {
                 HStack {
```
levieggertcru, line 29: I might make a separate ticket to add a ForEach here and do that later.  I should have originally done that.  We could then just check for toolOptions.isEmpty in ToolSettingsView which would be a little safer.

levieggertcru (approved):
Thanks @rachaelblue this looks good. I ended up making a separate ticket (https://jira.cru.org/browse/GT-2434) for adding a ForEach to the HStack for tool options.  I should've done that originally.  Not a high priority, but something that would be nice to update to in the future.

godtools/App/Flows/ToolSettings/Observer/Protocols/LinkShareable.swift
aaronlaib: This isn't me reviewing the PR haha, but @rachaelblue I really like how you set up Tool Options, allowing easy disable and enable of the different parts.

levieggertcru: Hey @aaronlaib welcome.  I agree.

rachaelblue: Sounds good @levieggertcru.  And thanks @aaronlaib!

PR #2259, opened 2024-08-27
GT-2436 add localized "network lost" string

rachaelblue:


levieggertcru (approved):
Thanks @rachaelblue for fixing this.

PR #2260, opened 2024-08-28
GT-1955 check for analytics event should trigger before tracking

rachaelblue:


godtools/App/Services/Renderer/Analytics/MobileContentRendererAnalytics.swift
```diff
@@ -27,12 +27,14 @@ class MobileContentRendererAnalytics {
     func trackEvents(events: [AnalyticsEvent], renderedPageContext: MobileContentRenderedPageContext) {
         
         for event in events {
+            if !event.shouldTrigger(state: renderedPageContext.rendererState) { break }
```
levieggertcru, line 30: Hey @rachaelblue for this do you want to use ```continue``` instead of ```break```?  Otherwise if any call to event.shouldTrigger evaluates to false it would prevent any remaining events from being checked against event.shouldTrigger.

rachaelblue: 🤦‍♀️ Thanks for catching that @levieggertcru

levieggertcru (approved):

PR #2295, opened 2024-10-14
GT-2438 lesson completion indicator

rachaelblue:


Podfile
```diff
@@ -13,7 +13,7 @@ use_frameworks!
 def shared_pods
   
   # CruGlobal pods
-  pod 'GodToolsShared', '1.0.1'
+  pod 'GodToolsShared', '1.1.0-SNAPSHOT'
```
frett, line 16: @levieggertcru let me know if you are close to releasing a build with this PR and I'll cut a release version of the GodToolsShared library

frett: I might back port the small change made for this to a 1.0.2 release. There some changes I'm needing to make to add support for the new cyoa page collection page type.

rachaelblue:
Thanks @levieggertcru!  Think this is ready for review.  I split out the "resume lesson" dialog into a subtask.

levieggertcru:
Hey @rachaelblue looks like tests will need to be updated to include the additional dependency changes in this PR.

godtools/App/DependencyContainer/AppDataLayerDependencies.swift
```diff
@@ -165,6 +165,14 @@ class AppDataLayerDependencies {
         )
     }
     
+    func getLessonListItemProgressRepository() -> GetLessonListItemProgressRepository {
```
levieggertcru, line 168: Hey @rachaelblue this dependency I'd like moved into the UserLessonProgressDataLayerDependencies.  I'm wondering if that's what you intended, but out of habit moved into here?  I know the specific feature dependency containers is something new we've been transitioning to.

godtools/App/Services/Renderer/Views/MobileContent/Views/Pages/MobileContentPagesViewModel.swift
```diff
@@ -44,7 +45,7 @@ class MobileContentPagesViewModel: NSObject, ObservableObject {
...
     
-    init(renderer: MobileContentRenderer, initialPage: MobileContentPagesPage?, resourcesRepository: ResourcesRepository, translationsRepository: TranslationsRepository, mobileContentEventAnalytics: MobileContentRendererEventAnalyticsTracking, getCurrentAppLanguageUseCase: GetCurrentAppLanguageUseCase, getTranslatedLanguageName: GetTranslatedLanguageName, initialPageRenderingType: MobileContentPagesInitialPageRenderingType, trainingTipsEnabled: Bool, incrementUserCounterUseCase: IncrementUserCounterUseCase, selectedLanguageIndex: Int?) {
+    init(renderer: MobileContentRenderer, initialPage: MobileContentPagesPage?, resourcesRepository: ResourcesRepository, translationsRepository: TranslationsRepository, mobileContentEventAnalytics: MobileContentRendererEventAnalyticsTracking, getCurrentAppLanguageUseCase: GetCurrentAppLanguageUseCase, getTranslatedLanguageName: GetTranslatedLanguageName, storeLessonProgressUseCase: StoreUserLessonProgressUseCase?, initialPageRenderingType: MobileContentPagesInitialPageRenderingType, trainingTipsEnabled: Bool, incrementUserCounterUseCase: IncrementUserCounterUseCase, selectedLanguageIndex: Int?) {
```
levieggertcru, line 48: Hey @rachaelblue was there a reason the ```StoreUserLessonProgressUseCase``` dependency needed to be added to the Parent class?  It seems like it should only be added to the LessonViewModel and it shouldn't be optional.

levieggertcru:
I'm not sure if this is intentional, but I noticed that the most forward progress isn't kept on a lesson.  For example, if I open a lesson and move forward say 5 pages and then close the lesson, my progress would be let's say 35%.  If I re-open that same lesson, go back to the first page of the lesson and close, my progress will say 0%.  I'm wondering should it still have said 35% which was my most forward progress on the lesson?

godtools/App/Features/UserLessonProgress/Data/UserLessonProgressRepository/Cache/RealmUserLessonProgress.swift
```diff
@@ -0,0 +1,21 @@
...
+//  Created by Rachael Skeath on 9/25/24.
+//  Copyright © 2024 Cru. All rights reserved.
+//
+
+import Foundation
+import RealmSwift
+
+class RealmUserLessonProgress: Object {
+    
+    @objc dynamic var lessonId: String = ""
```
levieggertcru, line 14: We can start using the new ```@Persisted``` declaration now that we're using Realm 10.10+.  Some of our realm objects are using it, I have a ticket in the backlog to update the rest of them.  (https://www.mongodb.com/docs/atlas/device-sdks/sdk/swift/model-data/object-models/#persisted-property-attributes)

levieggertcru:
Do you have this missing folder showing at Features/UserLessonProgress/Domain/Entities/ or is it just me?

missing-folder

godtools/App/Features/UserLessonProgress/Data/GetLessonListItemProgressRepository.swift
```diff
@@ -0,0 +1,71 @@
...
+    
+    private func formatProgressPercentage(progress: Double, appLanguage: AppLanguageDomainModel) -> String {
+        
+        let numberFormatter = NumberFormatter()
+        numberFormatter.numberStyle = .percent
+        numberFormatter.maximumFractionDigits = 0
+        numberFormatter.minimumFractionDigits = 0
+        numberFormatter.locale = Locale(identifier: appLanguage.localeId)
+        
+        return numberFormatter.string(from: NSNumber(value: progress)) ?? "\(progress)%"
```
levieggertcru, line 69: This logic would be really nice to move into a class in Share/Data-DomainInterface/Supporting/* in the event we want to share this business logic elsewhere.

godtools/App/Share/Views/PageNavigationCollectionView/PageNavigationCollectionView.swift
```diff
@@ -533,12 +533,15 @@ extension PageNavigationCollectionView {
...
             return false
         }
-                
-        collectionView.scrollToItem(
-            at: IndexPath(item: item, section: 0),
-            at: .centeredHorizontally,
-            animated: animated
-        )
+        
+        DispatchQueue.main.async {
```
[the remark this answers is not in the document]
rachaelblue: Hey @levieggertcru, yes this whole scroll method was getting triggered before the collection view was set up, so the scroll wouldn't happen and the collection view just always rendered to the first page.

godtools/App/Services/Renderer/Views/MobileContent/Views/Pages/MobileContentPagesViewModel.swift
rachaelblue: That's a good point.  I had it in the parent class because I use `pageDidAppear()` from the parent to trigger storing the progress, but I can just override that in LessonViewModel.

godtools/App/DependencyContainer/AppDataLayerDependencies.swift
levieggertcru: Hey @rachaelblue I'm going to mark this resolved since this is shared across features.

godtools/App/DependencyContainer/AppDataLayerDependencies.swift
```diff
@@ -417,6 +425,14 @@ class AppDataLayerDependencies {
         )
     }
     
+    func getUserLessonProgressRepository() -> UserLessonProgressRepository {
```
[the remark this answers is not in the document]
levieggertcru: Hey @rachaelblue I'm going to mark this resolved as well since this is shared across features.

godtools/App/Share/Views/PageNavigationCollectionView/PageNavigationCollectionView.swift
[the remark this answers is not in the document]
levieggertcru: Thanks @rachaelblue.  I'll take a closer look at this.  The renderer navigation is a bit finicky since it has to wait for layout constraints to complete in UIViewController viewDidLayoutSubviews (https://developer.apple.com/documentation/uikit/uiviewcontroller/1621398-viewdidlayoutsubviews).

godtools/App/Features/Lessons/Presentation/Lesson/LessonViewModel.swift
```diff
@@ -7,18 +7,22 @@
 //
 
 import UIKit
+import Combine
 
 class LessonViewModel: MobileContentPagesViewModel {
     
     private weak var flowDelegate: FlowDelegate?
+    private let storeLessonProgressUseCase: StoreUserLessonProgressUseCase?
```
levieggertcru, line 15: And now should no longer be optional.

godtools/App/Features/UserLessonProgress/Data/UserLessonProgressRepository/Cache/RealmUserLessonProgressCache.swift
```diff
@@ -0,0 +1,59 @@
...
+        let realm: Realm = realmDatabase.openRealm()
+        
+        let realmLessonProgress = RealmUserLessonProgress()
+        realmLessonProgress.lessonId = lessonProgress.lessonId
+        realmLessonProgress.lastViewedPageId = lessonProgress.lastViewedPageId
+        realmLessonProgress.progress = lessonProgress.progress
+        
+        do {
+            
+            try realm.write {
```
levieggertcru, line 51: I'd like our realm writes to start going through the RealmDatabase+Write.swift extension since there are tests for those extensions.   See. (https://github.com/CruGlobal/godtools-swift/blob/develop/godtools/App/Share/Data/RealmDatabase/RealmDatabase%2BWrite.swift#L63).

godtools/App/Features/UserLessonProgress/Data/UserLessonProgressRepository/Cache/RealmUserLessonProgressCache.swift
```diff
@@ -0,0 +1,59 @@
...
+        return realmDatabase.openRealm()
+            .objects(RealmUserLessonProgress.self)
+            .objectWillChange
+            .prepend(Void())
+            .eraseToAnyPublisher()
+    }
+    
+    func getUserLessonProgress(lessonId: String) -> UserLessonProgressDataModel? {
+        
+        if let realmLessonProgress = realmDatabase.openRealm().object(ofType: RealmUserLessonProgress.self, forPrimaryKey: lessonId) {
```
levieggertcru, line 32: I'd like our object reading to also go through the RealmDatabase+Read.swift extension since there are tests on those extensions. See (https://github.com/CruGlobal/godtools-swift/blob/develop/godtools/App/Share/Data/RealmDatabase/RealmDatabase%2BRead.swift#L58).

godtools/App/Flows/Lesson/LessonFlow.swift
```diff
@@ -43,18 +43,31 @@ class LessonFlow: ToolNavigationFlow, Flow {
             appLanguage: appLanguage,
             toolTranslations: toolTranslations
         )
-              
+        
+        let initialPageOrPreviousProgress: MobileContentPagesPage?
```
levieggertcru, line 47: Something I didn't think about is in what cases would we not want to open a lesson to last progress? Was that something that was discussed?

godtools/App/Services/Renderer/Views/MobileContent/Views/Pages/MobileContentPagesViewModel.swift
```diff
@@ -377,8 +377,19 @@ class MobileContentPagesViewModel: NSObject, ObservableObject {
         switch initialPage {
         
         case .pageId(let value):
-            return allPages.first(where: {$0.id == value})
```
levieggertcru, line 380: I need to think about this logic a little bit.  I'm not 100% certain if we want this change here in the base class.

godtools/App/Features/FeaturedLessons/Domain/Entities/FeaturedLessonDomainModel.swift
```diff
@@ -9,12 +9,12 @@
 import Foundation
 
 struct FeaturedLessonDomainModel: LessonListItemDomainModelInterface {
-        
     let analyticsToolName: String
     let availabilityInAppLanguage: ToolLanguageAvailabilityDomainModel
     let bannerImageId: String
     let dataModelId: String
     let name: String
+    var lessonProgress: LessonListItemProgressDomainModel
```
levieggertcru, line 17: This variable could be immutable too.  I don't think it's being mutated.

godtools/App/Features/UserLessonProgress/Data/GetLessonListItemProgressRepository.swift
```diff
@@ -0,0 +1,71 @@
...
+//  Created by Rachael Skeath on 10/14/24.
+//  Copyright © 2024 Cru. All rights reserved.
+//
+
+import Foundation
+import GodToolsToolParser
+import LocalizationServices
+import Combine
+
+class GetLessonListItemProgressRepository {
```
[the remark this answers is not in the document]
rachaelblue: Ah ok good point!  Was struggling with how this class fits in since there's no use case, but that makes sense.

godtools/App/Features/UserLessonProgress/Data/GetLessonListItemProgressRepository.swift
levieggertcru: Thanks @rachaelblue

godtools/App/Services/Renderer/Views/MobileContent/Views/Pages/MobileContentPagesViewModel.swift
levieggertcru: Was this added because we don't want to track a hidden page as part of lesson progress?

godtools/App/Share/Views/PageNavigationCollectionView/PageNavigationCollectionView.swift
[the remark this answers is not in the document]
levieggertcru: Okay have a fix here (https://github.com/CruGlobal/godtools-swift/pull/2298).  This was broken in all our tools.  I will need to create some navigation tests for initial renderer navigation.

godtools/App/Flows/Lesson/LessonFlow.swift
levieggertcru: Disregard this.  I see it's falling back to last progress only when initialPage is null on initialization. I read it incorrectly when asking that question.

godtools/App/Services/Renderer/Views/MobileContent/Views/Pages/MobileContentPagesViewModel.swift
rachaelblue: We can track a hidden page, we just don't want to open a hidden page in the case that a user switches languages and reopens a lesson, but the page they were on in the old language is hidden in the new language.  If that happens, we instead show the first non-hidden previous page.

Also if the user stops on the last visible page, we automatically restart at the first page when they reopen the lesson.

godtools/App/Features/Lessons/Presentation/Lesson/LessonViewModel.swift
```diff
@@ -7,16 +7,20 @@
 //
 
 import UIKit
+import Combine
 
 class LessonViewModel: MobileContentPagesViewModel {
     
     private weak var flowDelegate: FlowDelegate?
+    private let storeLessonProgressUseCase: StoreUserLessonProgressUseCase
+    private var storeLessonProgressCancellable: AnyCancellable?
```
levieggertcru, line 16: Sorry @rachaelblue one more thing to suggest is making this a static variable for the cancellable.  I know this isn't doing any network calls underneath, but want to be on the safe side in the event something changed in the future.  At some point we really do need some type of standard for our repository data writing and fetching for the underlying data.  Most of the time I could see us wanting to use a static ref when writing / posting network calls as we wouldn't want things to cancel if the user exits a screen and the cancellable ref is deallocated which terminates the request.

rachaelblue: If we do end up with network calls here in the future, we'd have to account for the case that a call is still running while the user opens a different lesson, and then we'd accidentally cancel the old call.  Just changing it back to a set won't fix that scenario, but not sure adding extra logic for that case is worth it at this point?

levieggertcru: Hey @rachaelblue that's a good point.  This will still be an issue. Need to do some research into this and come up with some standard. I'm going to make a note of this in my reminders.

levieggertcru (approved):
Thanks @rachaelblue!

PR #2298, opened 2024-10-22
GT-2438 Fix initial page navigation in renderer

levieggertcru:
Hey @rachaelblue looks like initial page navigation was broken across the board.  The collection view needed to be reloaded when performing an initial page navigation.  Once reloaded and the data model (pageModels) aren't mutated, page navigation can be performed without the need to reload the collection view.


rachaelblue (approved):
Thanks for looking into this @levieggertcru!

PR #2352, opened 2024-11-15
GT-2465 resume lesson modal

rachaelblue:


godtools/App/Features/UserLessonProgress/Domain/Interface/GetResumeLessonProgressModalInterfaceStringsRepositoryInterface.swift
```diff
@@ -0,0 +1,15 @@
+//
+//  GetResumeLessonProgressModalInterfaceStringsRepositoryInterface.swift
```
rachaelblue, line 2: Hey @levieggertcru, how do you feel about the naming convention making some names super long?  Was debating to take out the word "progress", but still gonna be pretty long either way 🤔

godtools/App/Features/UserLessonProgress/Presentation/ResumeLessonProgressModal/ResumeLessonProgressModal.swift
```diff
@@ -0,0 +1,78 @@
...
+    
+    var body: some View {
+        GeometryReader { geometry in
+            let totalSpaceAroundModal = modalInset * 2
+            let modalWidth = geometry.size.width - totalSpaceAroundModal
+            let totalSpaceAroundButtons = (buttonInset * 2) + buttonSpace
+            let buttonWidth = (modalWidth - totalSpaceAroundButtons) / 2
+            
+            ZStack {
+                if #available(iOS 15.0, *) {
```
rachaelblue, line 32: Also @levieggertcru, do you think we can bump the minimum deployment target to 15?

godtools/App/Features/UserLessonProgress/Domain/Interface/GetResumeLessonProgressModalInterfaceStringsRepositoryInterface.swift
levieggertcru: Hey @rachaelblue I know some names get super long.  I know we suffix a lot of key names such as UseCase, DomainModel, RepositoryInterface, etc. 

And some names get descriptive based on feature so there isn't collisions with other features. (If we could namespace that would help here). 

I think we could start omitting Repository from the interface naming.  

Maybe ```GetResumeLessonInterfaceStringsInterface``` or even ```GetResumeLessonStringsInterface```.  A tiny bit shorter.

godtools/App/Features/UserLessonProgress/Presentation/ResumeLessonProgressModal/ResumeLessonProgressModal.swift
```diff
@@ -0,0 +1,78 @@
...
+                    
+                    Text(viewModel.interfaceStringsDomainModel.subtitle)
+                        .font(FontLibrary.sfProTextRegular.font(size: 16))
+                        .foregroundColor(ColorPalette.gtGrey.color)
+                        .multilineTextAlignment(.center)
+                        .padding(.horizontal, buttonInset + 20)
+                        .padding(.bottom, 35)
+                    
+                    HStack(spacing: buttonSpace) {
+                        GTWhiteButton(title: viewModel.interfaceStringsDomainModel.startOverButtonText, width: buttonWidth, height: buttonHeight) {
```
levieggertcru, line 56: Hey @rachaelblue also on the GTWhiteButton and GTBlueButton can you inject a fontSize here of at least 15.

It defaults to 12 after inspecting the initializer which looks pretty small here.  

Looking at the Figma and I see size 15 is being used.  Sometimes I will even go a point or 2 larger than what Figma suggests if the screen size is small in Figma just depends.

rachaelblue:
Hey @levieggertcru, I originally was presenting the modal from the AppFlow and then navigating to the lesson after the user made a selection, but there's a slight delay when the lesson loads/presents, and it looked awkward to have the modal disappear and see no lesson for a second.  So my solution was to have the lesson and modal present nearly simultaneously so that the lesson is already behind the modal when the user makes a selection, and the page number gets updated if needed.  Visually this looks better, but does add the complication that the lesson opens before we know whether the user wants to resume or restart.  I tried making that happen from the lesson flow instead, but it does feel a little weird for the flow to be holding onto the viewModel in order to get page info/update the page.  Let me know what you think.

godtools/App/Features/UserLessonProgress/Presentation/ResumeLessonProgressModal/ResumeLessonProgressModal.swift
levieggertcru: We got the go ahead to bump to iOS 15 so feel free to make that change in this PR.

levieggertcru:
Hey @rachaelblue what if you push the LessonView onto the navigation stack with animated false and then dismiss the resume lesson modal?  Is there still some delay in the LessonView itself?  I know that view waits for the layout constraints to complete before doing anything navigation.

rachaelblue:
Hey @levieggertcru, that did work better to present LessonView without animation before dismissing the modal.  Felt a little weird moving some initialization around in the LessonFlow now, so let me know what you think.  Thanks!

levieggertcru (dismissed):

levieggertcru (approved):

levieggertcru (approved):

PR #2357, opened 2024-11-25
GT 2465 create extension for lesson flow views

levieggertcru:
Hey @rachaelblue I went ahead and made some updates to the LessonFlow and added a getLesson() method to allocate the LessonView.  I noticed it helps clean up the Flows by creating individual methods for allocating the Views. The renderer will also get allocated at that point.  I think that helped clean things up a bit more.  

Let me know if the init logic looks right.  

I also created a reference to the ToolTranslationsDomainModel and also added a computed property for the lesson: ResourceModel on the LessonFlow.  Some of those properties I don't think you were fully aware of which is understandable. 

Also removed some files marked for iOS 14 obsolete.  Nice to have those removed for iOS 15.


rachaelblue (approved):
Thanks @levieggertcru, looks great!  That cleaned things up much better.  I was struggling piecing it together (and not knowing where else to get some of those attributes).

PR #2385, opened 2025-01-07
GT-2442 Lesson Title Alignment

rachaelblue:

### DIFF
--- a/godtools/App/Features/FeaturedLessons/Domain/Entities/FeaturedLessonDomainModel.swift
+++ b/godtools/App/Features/FeaturedLessons/Domain/Entities/FeaturedLessonDomainModel.swift
@@ -14,7 +14,7 @@ struct FeaturedLessonDomainModel: LessonListItemDomainModelInterface {
     let bannerImageId: String
     let dataModelId: String
     let name: String
-    let nameLanguageDirection: LanguageModel.Direction
+    let nameLanguageDirection: LanguageDirectionDomainModel
     let lessonProgress: LessonListItemProgressDomainModel
 }
 
--- a/godtools/App/Features/FeaturedLessons/Domain/Entities/FeaturedLessonDomainModel.swift
+++ b/godtools/App/Features/FeaturedLessons/Domain/Entities/FeaturedLessonDomainModel.swift
@@ -14,7 +14,7 @@ struct FeaturedLessonDomainModel: LessonListItemDomainModelInterface {
     let bannerImageId: String
     let dataModelId: String
     let name: String
-    let nameDirection: LanguageModel.Direction
+    let nameLanguageDirection: LanguageModel.Direction
     let lessonProgress: LessonListItemProgressDomainModel
 }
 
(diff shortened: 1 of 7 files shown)


godtools/App/Features/FeaturedLessons/Domain/Entities/FeaturedLessonDomainModel.swift
```diff
@@ -14,6 +14,7 @@ struct FeaturedLessonDomainModel: LessonListItemDomainModelInterface {
     let bannerImageId: String
     let dataModelId: String
     let name: String
+    let nameLanguageDirection: LanguageModel.Direction
```
levieggertcru, line 17: Can you change this type to ```LanguageDirectionDomainModel```.

godtools/App/Features/Lessons/Domain/Entities/LessonListItemDomainModel.swift
```diff
@@ -15,6 +15,7 @@ struct LessonListItemDomainModel: LessonListItemDomainModelInterface {
     let bannerImageId: String
     let dataModelId: String
     let name: String
+    let nameLanguageDirection: LanguageModel.Direction
```
levieggertcru, line 18: Can you change this type to ```LanguageDirectionDomainModel```.

godtools/App/Features/Lessons/Domain/Entities/LessonListItemDomainModelInterface.swift
```diff
@@ -13,5 +13,6 @@ protocol LessonListItemDomainModelInterface {
     var availabilityInAppLanguage: ToolLanguageAvailabilityDomainModel { get }
     var bannerImageId: String { get }
     var name: String { get }
+    var nameLanguageDirection: LanguageModel.Direction { get }
```
levieggertcru, line 16: Can you change this type to ```LanguageDirectionDomainModel```.

levieggertcru (changes requested):
Hey @rachaelblue looks good.

Just need to change the LanguageModel.Direction type to LanguageDirectionDomainModel in the domain layer.

Kind of annoying mapping almost the same thing, but to keep with our architecture the domain layer isn't supposed to know anything about the data layer.  It's the lowest level.  It only knows of the data layer via the interface (dependency inversion).

godtools/App/Features/FeaturedLessons/Domain/Entities/FeaturedLessonDomainModel.swift
rachaelblue: Makes sense, thanks!  I do get mixed up when properties are so similar.  Just need to remember to think through the layers.

levieggertcru: Thanks, I'll need to remove the LanguageModel from LanguageDirectionDomainModel too since it really shouldn't know about the data layer.  Instead I'll create an extension on LanguageModel to convert to the LanguageDirectionDomainModel.

levieggertcru (approved):

PR #2386, opened 2025-01-08
GT-2489 state enum for language download icon

rachaelblue:


levieggertcru (approved):
Thanks Rachael I do like the single state for the icon and pushing some of that logic back into the DownloadableLanguageItemView.

PR #2450, opened 2025-02-10
GT-1454 implement play, stop, and toggle listeners on animations in renderer

levieggertcru:


godtools/App/Services/Renderer/Views/MobileContent/Views/ContentAnimation/MobileContentAnimationViewModel.swift
```diff
@@ -11,11 +11,21 @@ import GodToolsToolParser
 
 class MobileContentAnimationViewModel: MobileContentViewModel {
     
+    enum PlaybackState {
+        case pause
+        case play
+        case stop
+    }
+    
+    private static let eventToggleAnimation: String = "toggleAnimation"
```
frett, line 20: the `toggleAnimation` event id is just an arbitrary event id I created, in my sample content XML that event id is on both the playListeners and stopListeners and will trigger the corresponding listener based on if the animation is currently playing or not

frett: all that to say you shouldn't have a random event id hard-coded like this

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentAnimation/MobileContentAnimationViewModel.swift
```diff
@@ -37,11 +47,38 @@ class MobileContentAnimationViewModel: MobileContentViewModel {
             case .failure( _):
                 animatedViewModel = nil
             }
+            
+            playbackState = animationModel.autoPlay ? .play : .stop
```
frett, line 51: unfortunately `playbackState` may be a bit more complex then this. If the animation has `loop=false` it should automatically transition from .play to .stop once the animation finishes

frett: I know the lottie APIs on Android allow inspection to see if the animation is currently playing or not

godtools/App/Services/Renderer/Views/MobileContent/Views/ContentAnimation/MobileContentAnimationViewModel.swift
levieggertcru: Hey Daniel, I'm not crazy about it either.  I'll need to update the ViewModel to have a two-way binding so it knows when the animation is truly stopped.  Currently tracking the state only in the ViewModel which won't get updated when an animation finishes.

levieggertcru: Gonna work on adding that in.

frett (approved):

PR #2456, opened 2025-02-17
GT-2478 add a view model for downloadable language

levieggertcru:
Changes in this PR:

- Created a ViewModel for DownloadableLanguageItemView.  This helps split some of the responsibilities in the View by moving state driven logic into the ViewModel.  
- Created a shared common AnimateDownloadProgress.swift which encapsulates the logic for animating a download progress towards a download progress target.  
- Adding an ObservableObject to maintain the DownloadableLanguageItemView state when views are recycled.  This object is added to a static dictionary accessed by the language id. 
- Created a static reference to the language downloads and mark for removal timers this way those publishers will continue to execute in the background.
- Removed download progress from DownloadableLanguageListItemDomainModel.swift.  That is now tracked separately. I should've caught that sooner, but that goes against the dependency direction since the presentation layer is on top of the domain layer.


levieggertcru:
Hey @rachaelblue I think I need a break from the language download.  Making my head spin lol.  

I removed the rounding logic.  I think it will be okay since incrementing by 0.1.  Shouldn't have any floating point issues.  I 

I also had a static container that would hold onto a dictionary of CurrentValueSubject<Double?, Error> in order to persist the download progress between view recycling.  I ended up removing that and replacing it with another container that holds onto an ObservableObject.  That simplified a bit. 

I like the container because it can hold a dictionary of values and it can basically lazy load values by either returning the existing instance from the dictionary or creating a new instance if one doesn't exist.

levieggertcru:
Converted back to draft, need to drop in the download error handling.

godtools/App/Features/AppLanguage/Presentation/DownloadableLanguages/Subviews/DownloadableLanguageItem/DownloadableLanguageItemViewModel.swift
```diff
@@ -0,0 +1,256 @@
...
+            }
+            
+            return Just(iconState)
+                .eraseToAnyPublisher()
+        }
+        .assign(to: &$iconState)
+        
+        recycleState
+            .$downloadProgress
+            .sink(receiveValue: { [weak self] (progress: Double?) in
```
levieggertcru, line 88: Hey @rachaelblue note that I didn't add receive on main queue here because I noticed it would cause the UI to flicker when tapping multiple languages to download because I think the ViewModel was getting re-allocated when tapping additional languages.
The static observers will call receive on main when updating the recycle state object.

levieggertcru:
Hey @rachaelblue just a note, I'm going to look into persisting the download progress and errors to realm (GT-2524).  The more I think about it, to me that seems to be the best approach if we need to persist state outside the scope of the parent view / parent view model. 
Otherwise we have to keep an object for every sub item in memory which overtime can get expensive and then there's also the question where do you free up that memory.

PR #2457, opened 2025-02-19
GT-2478 When sync is called on ToolLanguageDownloader, mark all downloadable languages as downloaded

levieggertcru:
The ToolLanguageDownloader sync will run anytime the app is launched from a terminated state or launched from the background past 2 hours or more.

It will fetch downloadable languages from the DownloadedLanguagesRepository and download new data (translations, attachments, articles, etc.).

When this is run I think it's a good idea to mark any downloadable language in DownloadedLanguagesRepository as downloaded completed = true.  This way if a user was downloading some languages and they either closed the app or were on a slow network connection and closed the app, the next time they launch the app the sync would mark those as completed and begin downloading data again in the background.

If they were to open up the list of downloadable languages they would see those languages they tapped on as completed.

### DIFF
--- a/godtools/App/Features/AppLanguage/Data/DownloadedLanguagesRepository/Cache/RealmDownloadedLanguagesCache.swift
+++ b/godtools/App/Features/AppLanguage/Data/DownloadedLanguagesRepository/Cache/RealmDownloadedLanguagesCache.swift
@@ -122,4 +122,27 @@ class RealmDownloadedLanguagesCache {
         )
         .eraseToAnyPublisher()
     }
+    
+    func markAllDownloadsAsCompleted() {
+            
+        let realm: Realm = realmDatabase.openRealm()
+        
+        let nonCompletedDownloads = realm
+            .objects(RealmDownloadedLanguage.self)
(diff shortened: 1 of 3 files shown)


rachaelblue (approved):
What if the download never completes though?  I guess they can tap to remove it and try again.  Seems like an odd state to be in either way.

levieggertcru:
Hey @rachaelblue at some point the download will complete since launching the app triggers the sync to run on all the downloaded languages. 

Currently there isn't any tracking of partial downloads.  I'd like to look into adding some better error handling to display to the user and also retrying failed requests. 

But, at this moment it seems the best way to handle it is mark it as completed and let the sync clean up anything that is missing.

levieggertcru:
At the moment if a download does fail it will be removed from the downloaded languages repository.  

https://github.com/CruGlobal/godtools-swift/blob/develop/godtools/App/Features/AppLanguage/Data-DomainInterface/DownloadToolLanguageRepository.swift#L33

But, there is the case of someone starting a download and then maybe they kill the app while a download is still in progress.

PR #2458, opened 2025-02-19
GT-2478 Fix to ToolDownloader progress

levieggertcru:
This PR addresses 2 fixes in ToolDownloader when determining the progress.

1. Originally all of the translations and articles were counted as a single request (single publisher).  Those are now broken up into individual publishers and added to the ```allRequests: [AnyPublisher<Void, Error>]``` array which is used to determine progress.

2. I decided to add a check for the downloadCount against the number of running requests and clamp the progress to 1 when downloadCount reaches the number of running requests.


rachaelblue (approved):
Thanks @levieggertcru, I think clamping to 1 will help if somehow requests go out twice.  But hopefully that won't be happening anyway.

PR #2509, opened 2025-03-25
GT-1719 favorites drag and drop

rachaelblue:
Hey @levieggertcru, the one thing that's bugging me with this is that to move a tool you have to long press on it, but the tool background highlights when you do that.  Do you think it looks weird?  I feel like only the card itself should move and the background shouldn't do anything.


godtools/App/Share/Data/FavoritedResourcesRepository/Cache/Models/RealmFavoritedResource.swift
```diff
@@ -13,14 +13,9 @@ class RealmFavoritedResource: Object {
     
     @objc dynamic var createdAt: Date = Date()
     @objc dynamic var resourceId: String = ""
+    @objc dynamic var position: Int = 0
```
levieggertcru, line 16: Something that just came to mind here is user's who already have the app installed all of their favorited tools will be at position 0 which could cause some issues.

One route is to make the position optional and then first sort by position and fallback to sorting by createdAt date for positions that are null.

godtools/App/Features/Favorites/Presentation/AllYourFavoriteTools/AllYourFavoriteToolsViewModel.swift
```diff
@@ -190,4 +192,28 @@ extension AllYourFavoriteToolsViewModel {
         
         flowDelegate?.navigate(step: .toolTappedFromAllYourFavoritedTools(tool: tool))
     }
+    
+    func toolMoved(fromOffsets source: IndexSet, toOffset destination: Int) {
+        for index in source {
```
levieggertcru, line 197: Hey @rachaelblue I noticed something wonky when re-ordering a tool from position 1 to 0 and from position 0 to 1.  I'll include a video, but I think it has to do with the comment I made on making the position optional on the Realm Object.

Another thought that came to mind is I wonder if there will be any glitchy behavior when re-ordering the list and responding to realm changes since those changes have to dispatch to the main thread again.  I was thinking if that is an issue, maybe mutate the ViewModel Published property first before sending changes to the UseCase.

(https://sarunw.com/posts/swiftui-list-onmove/#move-items-in-data-source)
(https://developer.apple.com/documentation/Swift/MutableCollection/move(fromOffsets:toOffset:))

levieggertcru:
Hey @rachaelblue I think it would be nice if we could exclude the highlight background from the drag and drop, however, I'm not sure how difficult it may be to exclude that?  I'm curious to look at that some more in the Xcode UI debugger too so will look more at that.

godtools/App/Features/Dashboard/Presentation/Tools/Subviews/ToolCard/ToolCardView.swift
```diff
@@ -153,11 +153,13 @@ struct ToolCardView: View {
...
                                 toolDetailsTappedClosure?()
                             }
+                            .buttonStyle(.plain)
                             
                             GTBlueButton(title: viewModel.openButtonTitle, font: navButtonFont, width: navButtonWidth, height: navButtonHeight, titleHorizontalPadding: navButtonTitleHorizontalPadding, accessibility: .openTool) {
-
+                                
                                 openToolTappedClosure?()
                             }
+                            .buttonStyle(.plain)
```
levieggertcru, line 162: How did this go?  Did you notice that all the buttons were firing when changing to List?

godtools/App/Features/Favorites/Presentation/AllYourFavoriteTools/AllYourFavoriteToolsViewModel.swift
levieggertcru: Here is that video:

https://github.com/user-attachments/assets/fcb0ce09-29a6-4f7a-9a1f-617f6751256c

levieggertcru:
One more thing, can you remove this line at the top of the List?  Hopefully that is easy to remove.  

remove-line-at-top

godtoolsTests/App/Features/Favorites/Data-DomainInterface/ReorderFavoritedToolRepositoryTests.swift
```diff
@@ -0,0 +1,137 @@
...
+//  Copyright © 2025 Cru. All rights reserved.
+//
+
+import Foundation
+@testable import godtools
+import Combine
+import Quick
+import Nimble
+
+class ReorderFavoritedToolRepositoryTests: QuickSpec {
```
levieggertcru, line 15: It would be nice to include a couple more tests:
- When a tool is favorited.
- When a tool is unfavorited.

godtools/App/Features/Favorites/Domain/UseCases/ReorderFavoritedToolUseCase.swift
```diff
@@ -0,0 +1,24 @@
...
+
+class ReorderFavoritedToolUseCase {
+    
+    private let reorderFavoritedToolRepository: ReorderFavoritedToolRepositoryInterface
+    
+    init(reorderFavoritedToolRepository: ReorderFavoritedToolRepositoryInterface) {
+        self.reorderFavoritedToolRepository = reorderFavoritedToolRepository
+    }
+    
+    func reorderFavoritedToolPublisher(toolId: String, originalPosition: Int, newPosition: Int) -> AnyPublisher<[FavoritedResourceDataModel], Error> {
```
levieggertcru, line 20: Hey @rachaelblue I missed this one, but we will need a DomainModel being returned from the UseCase.  The DomainLayer shouldn't reference types in the data layer since that breaks the dependency direction.  
So you will want to map that in your interface implementation. I'd suggest create a simple domain model that wraps the dataModelId that way in your tests you can check against the data model id.

godtools/App/Share/Data/FavoritedResourcesRepository/Cache/Models/RealmFavoritedResource.swift
rachaelblue: I ended up going the route to validate each time a fetch occurs.  A user that already has the app installed wouldn't be able to reorder if we didn't "correct" the positions at some point (unless they removed and re-added their favorites).

godtools/App/Features/Favorites/Presentation/AllYourFavoriteTools/AllYourFavoriteToolsViewModel.swift
rachaelblue: This was likely because of the issue with having previous favorites all at position = 0.  Before adding the validation, there really wasn't a way to recover if a position was incorrect, the reorder would fail and you'd see a glitch.  Should be good now when reordering "old" favorites.

godtools/App/Features/Dashboard/Presentation/Tools/Subviews/ToolCard/ToolCardView.swift
rachaelblue: Yes!  Thanks for the heads up on this!  I thought I had broken something at first before I remembered what you said.  Weird that it defaults to that behavior.  I ended up moving the .buttonStyle to apply to the entire view rather than each subview.  Which in theory should have removed the row highlighting, according to https://stackoverflow.com/questions/56545444/how-to-remove-highlight-on-tap-of-list-with-swiftui.  Wondering if iOS 18 has a bug because it seems like no one can get the highlight to remove consistently.  I've gone around and around with it and can't get it to go away 😩

rachaelblue:
Hey @levieggertcru, I think this is ready for a second look.

godtools/App/Share/Data/FavoritedResourcesRepository/Cache/RealmFavoritedResourcesCache.swift
```diff
@@ -64,6 +64,14 @@ class RealmFavoritedResourcesCache {
             .map({FavoritedResourceDataModel(realmFavoritedResource: $0)})
     }
     
+    func getFavoritedResourcesSortedByPosition(ascendingOrder: Bool = true) -> [FavoritedResourceDataModel] {
```
levieggertcru, line 67: Hey @rachaelblue I notice this method doesn't run through the validation, only the publisher does.  And with this method exposed as internal it could be called and the list wouldn't be sorted.  I would suggest moving the validation to this method and the publisher call this method.

godtools/App/Share/Data/FavoritedResourcesRepository/Cache/RealmFavoritedResourcesCache.swift
```diff
@@ -117,11 +143,109 @@ class RealmFavoritedResourcesCache {
...
+    }
+    
+    // MARK: - Private
+    
+    private func validatePositionsForFavoritedResources(ascendingOrder: Bool) -> AnyPublisher<[FavoritedResourceDataModel], Never> {
+        let favoritesSortedByPosition = self.realmDatabase.openRealm()
+            .objects(RealmFavoritedResource.self)
+            .sorted(byKeyPath: #keyPath(RealmFavoritedResource.position), ascending: ascendingOrder)
+        
+        lazy var favoritesDataModels: [FavoritedResourceDataModel] = favoritesSortedByPosition.map({
```
levieggertcru, line 213: Ah didn't realize this was marked ```lazy```, at first I was thinking you created some extra overhead, but with ```lazy``` this wouldn't get invoked.  Was about to comment, but checked and looks good.

levieggertcru (changes requested):
Hey @rachaelblue this is looking really nice!  Sorting is running smooth.

I did notice a couple things.  One with the sorting logic on storage of new favorites.  See my comment there and let me know your thoughts on option 2.  I created a PR to see what I'm referring to.  (https://github.com/CruGlobal/godtools-swift/pull/2529)

Also with the validation, moving that method call out of the publisher.  This actually might help with option 2 when storing new favorited tools.

godtools/App/Features/Dashboard/Presentation/Tools/Subviews/ToolCard/ToolCardView.swift
levieggertcru: Ah says may apply a visual effect. (https://developer.apple.com/documentation/swiftui/plainbuttonstyle)

godtools/App/Share/Data/FavoritedResourcesRepository/Cache/RealmFavoritedResourcesCache.swift
```diff
@@ -90,22 +109,29 @@ class RealmFavoritedResourcesCache {
...
             
-            let favoritedResource = FavoritedResourceDataModel(id: ids[index], createdAt: createdAtDate)
+            let favoritedResource = FavoritedResourceDataModel(id: ids[index], createdAt: createdAtDate, position: 0)
             
             newFavoritedResources.append(favoritedResource)
         }
 
         return realmDatabase.writeObjectsPublisher { (realm: Realm) -> [RealmFavoritedResource] in
             
+            let existingFavorites = realm.objects(RealmFavoritedResource.self)
```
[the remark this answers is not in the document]
rachaelblue: Welp tests are failing now.  @levieggertcru you can hold off on reviewing for now.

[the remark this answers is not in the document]
levieggertcru: Hey @rachaelblue looks like the issue is when storing the new favorited tools, they aren't getting written to Realm.

I created an example here of writing those to realm.  I haven't tested this. (https://github.com/CruGlobal/godtools-swift/blob/GT-1719-favorites-write-objects/godtools/App/Share/Data/FavoritedResourcesRepository/Cache/RealmFavoritedResourcesCache.swift#L150)

Also note, I pass the same realm instance into your validate method, however, I don't think that step is necessary since those newly favorited objects are added to realm in the write transaction.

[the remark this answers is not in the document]
rachaelblue: Hey @levieggertcru, you're right, I missed that I was removing the piece that persisted the new favorite. 🤦‍♀️  But a separate issue occurred to me-- if we set all new tools to index 0 and then validate every time, we'd be wiping out the user's set order each time they add a favorite (since validate sorts by createdAt initially).  So I ended up reverting back to the logic here that updates existing favorites when new ones are added (moving the old ones down in the list), and fixed the issue with all new tools being at position 0.  Also added a test for storing an array of favorites instead of just 1.

rachaelblue:
Hey @levieggertcru, made those changes we discussed and ready for review.

godtools/App/Share/Data/FavoritedResourcesRepository/Cache/RealmFavoritedResourcesCache.swift
```diff
@@ -90,38 +90,141 @@ class RealmFavoritedResourcesCache {
...
+        let favoritesSortedByPosition = Array(
+            realm.objects(RealmFavoritedResource.self)
+            .sorted(byKeyPath: #keyPath(RealmFavoritedResource.position), ascending: ascendingOrder)
+            )
+        
+        if favoritesSortedByPosition.allSatisfy({ $0.position == 0 }) {
+            
+            let favoritesSortedByCreatedAt = realm
+                .objects(RealmFavoritedResource.self)
+                .sorted(byKeyPath: #keyPath(RealmFavoritedResource.createdAt), ascending: !ascendingOrder)
```
levieggertcru, line 207: Hey @rachaelblue I ran a fresh build on develop, pre migration, and added a tool to my favorited list.  I then ran this branch and I noticed my tool order was in reverse. 

It took me a second to see why, but right here on ascendingOrder the flag is getting flipped ```! ascendingOrder ```.

godtools/App/Features/ToolShortcutLinks/Data-DomainInterface/GetToolShortcutLinksRepository.swift
```diff
@@ -25,7 +25,7 @@ class GetToolShortcutLinksRepository: GetToolShortcutLinksRepositoryInterface {
     
     func getLinksPublisher(appLanguage: AppLanguageDomainModel) -> AnyPublisher<[ToolShortcutLinkDomainModel], Never> {
         
-        return favoritedResourcesRepository.getFavoritedResourcesSortedByCreatedAtPublisher(ascendingOrder: false)
+        return favoritedResourcesRepository.getFavoritedResourcesSortedByPositionPublisher(ascendingOrder: false)
```
levieggertcru, line 28: For tool shortcuts make sure it matches the favorited list order.  Ascending defaults to true so you will want to remove this flag.

levieggertcru (changes requested):
Thanks @rachaelblue, I like this direction better.  I like that the createdAt is removed too.

Just 2 little things I noticed.

godtools/App/Share/Data/FavoritedResourcesRepository/Cache/RealmFavoritedResourcesCache.swift
levieggertcru: I would suggest putting ```false``` there and removing the argument on the method.  We will always want to migrate those so the newest is first.

levieggertcru: Then you will need to update your methods ```getFavoritedResourcesSortedByPosition and getFavoritedResourcesSortedByPositionPublisher``` to sort by position if you want to use the ascendingOrder argument on those methods.

levieggertcru: Hey @rachaelblue that sounds good!

levieggertcru (approved):
Hey @rachaelblue looks good!  Getting correct migration order.

PR #2525, opened 2025-04-01
GT-2307 add green/red color to language supported text

rachaelblue:


godtools/App/Features/ToolDetails/Presentation/ToolDetails/Subviews/VersionsCard/ToolDetailsVersionsCardView.swift
```diff
@@ -69,6 +69,8 @@ struct ToolDetailsVersionsCardView: View {
                             VStack(alignment: .trailing, spacing: 5) {
                                 
                                 Text(viewModel.languages)
+                                    .foregroundColor(ColorPalette.gtLightGrey.color)
```
levieggertcru, line 72: Good catch applying that back to the child view

levieggertcru (approved):

PR #2534, opened 2025-04-07
Opt In Notification Feature

JasonBennett41:
The Opt In Notification feature is designed as a reusable component, attachable to a view via .overlay modifier:

```
DashboardView{}.overlay(
                alignment: Alignment.bottom,
                content: {
                    if viewModel.isOptInNotificationActive {
                        OptInNotificationView(
                            viewModel:
                                viewModel.getOptInNotificationViewModel()
                        ).transition(.move(edge: .bottom))

                    }
                }
            )
```

Within the view model, shouldPromptNotificationSheet() checks that:

1. The user has completed onboarding
2. The user has not already granted notification permissions
3. The user has not been prompted recently*
4. The user has not exceeded the maximum amount of prompts*

*The 'time' and 'maxPrompts' variables are intended to be integrated with FirebaseRemoteConfig, but are hard coded as of this PR

```
func shouldPromptNotificationSheet() async {

        refreshNotificationStatus()

        let lastPrompted =
            optInNotificationRepository.getLastPrompted()
            ?? Date.distantPast

        let promptCount =
            optInNotificationRepository.getPromptCount()

        let currentDate = Date()
        let calendar = Calendar.current
        let twoMonthsAgo = calendar.date(
            byAdding: .month,
            value: -2,
            to: currentDate
        )!

        guard isOnboardingLaunch == false else { return }
        guard notificationStatus != "Approved" else { return }
        guard promptCount <= 5 else { return }

        if (notificationStatus == "Denied"
            || notificationStatus == "Undetermined")
            && lastPrompted < twoMonthsAgo
        {

            await MainActor.run {
                isActive = true
            }

            optInNotificationRepository.recordPrompt()
        }

    }
```
    
If a user meets all the prerequisites, they will encounter the Opt In Notification sheet.  The sheet itself responds to user input by showing the system permission request dialog (for a first time prompt) or showing a custom alert dialog (if a user has already interacted with the first time request).  The prompt is recorded by storing 'lastPromptedOptInNotification' and 'optInNotificationPromptCount' in user defaults for use in determining the need for future prompts.
    
To ensure the parent view (in this case DashboardView) reacts to 'isActive' status, a stream is used in parent view model to listen for changes in the opt in notification view model:

```
dashboardPresentationLayerDependencies.optInNotificationViewModel
            .$isActive
            .receive(on: DispatchQueue.main).sink { [weak self] isActive in
                guard let self = self else { return }

                withAnimation {
                    self.isOptInNotificationActive = isActive
                }
            }.store(in: &cancellables)
```


godtools/App/Features/Dashboard/Presentation/Dashboard/DashboardPresentationLayerDependencies.swift
```diff
@@ -9,83 +9,142 @@
...
         return LessonsViewModel(
             flowDelegate: unwrappedFlowDelegate,
-            resourcesRepository: appDiContainer.dataLayer.getResourcesRepository(),
-            getCurrentAppLanguageUseCase: appDiContainer.feature.appLanguage.domainLayer.getCurrentAppLanguageUseCase(), 
-            getUserLessonFiltersUseCase: appDiContainer.feature.lessonFilter.domainLayer.getUserLessonFiltersUseCase(),
-            viewLessonsUseCase: appDiContainer.feature.lessons.domainLayer.getViewLessonsUseCase(),
-            trackScreenViewAnalyticsUseCase: appDiContainer.domainLayer.getTrackScreenViewAnalyticsUseCase(),
-            trackActionAnalyticsUseCase: appDiContainer.domainLayer.getTrackActionAnalyticsUseCase(),
-            attachmentsRepository: appDiContainer.dataLayer.getAttachmentsRepository()
+            resourcesRepository: appDiContainer.dataLayer
```
levieggertcru, line 51: Hey @JasonBennett41 thank you for all the work on this feature and thank you for taking a stab at clean architecture by creating use cases.  Creating those will definitely add some extra development work so I appreciate that.

I'm curious, does your Xcode run some sort of linter?  I noticed it appeared that there was some auto formatting happening.  I'm curious to know what sort of tool you ran.  Maybe it's something to incorporate into this project.

levieggertcru: The only negative to the auto formatting is it does show a lot of additional changes in this PR.  However, I can quickly work through it and skim past it.  You may want to disable that for future client work just to avoid adding the additional PR changes.

levieggertcru (changes requested):
Hey @JasonBennett41 would love if you could revert those auto formatting changes.  I'm not sure what happened, but there's some crazy formatting happening that's making it difficult to review what actually changed in some of these files.

godtools/App/Features/Dashboard/Presentation/Dashboard/DashboardPresentationLayerDependencies.swift
JasonBennett41: I see, sorry about that I didn't really think twice about it.  I just run 'ctrl + shift + i' as I am working to "Format File with 'swift-format'", it's a native xcode tool rather than any sort of extension.  To be honest, I am unsure about how to revert the formatting aside from maybe just copy-pasting from previous commits.  Would you like me to try that out?

levieggertcru: Ah I didn't know of that command.  From a quick scan it doesn't look like too many files were heavily modified.  

If you could revert the following 4 files that would really help.  They appear to have the most formatting changes.

 
Screenshot 2025-04-08 at 2 48 39 PM

JasonBennett41: Okay, give me a few hours I'm at an appointment right now. Will get that taken care of 👍🏽

levieggertcru: Sounds good, thank you.

JasonBennett41: Okay, I've unformatted the requested files and it looks like it worked out.  Let me know if there's anything else

godtools/am.lproj/Localizable.strings
```diff
@@ -94,6 +96,16 @@
 /* Please keep to a 9 character limit, longer phrases may get cut off */
 "favorites.favoriteLessons.details" = "ዝርዝሮች";
 
+"optInNotification.title" = "ምክር እና ደጋፊነት ይውሰዱ";
```
levieggertcru, line 99: Hey @JasonBennett41 wish I had spotted this so you didn't have to create all of the additional translations. With interface strings we only need to upload the english strings and OneSky translators will take care of the rest.  This is something I will make a note of in the GodTools ReadMe.

godtools/App/Features/Dashboard/Presentation/Dashboard/DashboardPresentationLayerDependencies.swift
levieggertcru: Hey @JasonBennett41 I missed one file.  Could you also unformat AppFlow.swift.

godtools/am.lproj/Localizable.strings
JasonBennett41: Yes I was made aware of this after I went in and added them. Was told they would just be overridden by the OneSky translations. No worries though it didn't take that long to input them anyways

godtools/App/Features/Dashboard/Presentation/Dashboard/DashboardPresentationLayerDependencies.swift
JasonBennett41: Yes I will do it

JasonBennett41: @levieggertcru just got the AppFlow reverted

levieggertcru: Thank you

levieggertcru (changes requested):
Hey it's looking good.  Just a few minor changes I noted. 

I really like that you took a stab at our very layered architecture.

I am thinking about one more update and that's integrating this with our coordinator pattern where the coordinator is responsible for all navigation decisions and navigation via UINavigationController.     

In that update the logic to trigger the initial presentation of the opt-in notification view would happen in the coordinator instead of the ViewModel.  Then any user actions from the presentation layer would get delegated to the coordinator.

I'm still debating this just because with how the opt-in notification view is presented from the dashboard view.

The solution you have now works well and you were able to delegate the additional dialog to the AppFlow. I also don't imagine this flow getting any more complex.

godtools/App/Features/OptInNotification/Data/LastPromptedOptInNotificationRepository/Cache/OptInNotificationsUserDefaultsCache.swift
```diff
@@ -0,0 +1,77 @@
...
+            ) as? String
+
+        else {
+            return nil
+        }
+
+        let lastPromptedDate: Date =
+            OptInNotificationsUserDefaultsCache.dateFormatter.date(
+                from: lastPrompted
+            ) ?? Date()
```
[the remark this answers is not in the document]
JasonBennett41: So for this, distantPast is there in case of a null value being returned by the cache (indicating the user has not been shown the prompt before), in order to pass the 'lastPrompted < twoMonthsAgo' condition and activate the bottom sheet.

The Date() default thinking in cache was "should never happen but if it does, that indicates some previous prompt happened and we should default to a value which will not trigger the bottom sheet", however I think your suggestion sounds more intuitive.

Returning null will result in a fallback to Date.distantPast, but this is okay since the guard of notificationStatus will prevent an erroneous prompt to a user who has already granted permissions.

Sounds like I need to add logic which:

1) In the case of 'Approved' status: deletes bad data, and re-records the date to prevent future issues

2) For not-approved status: deletes bad data only with the assumption that re-recording is unnecessary since it will be recorded in shouldPromptNotificationSheet()

Sound good?

godtools/App/Features/OptInNotification/Domain/Interface/GetCheckNotificationStatusInterface.swift
```diff
@@ -0,0 +1,15 @@
...
+//  Created by Jason Bennett on 3/31/25.
+//  Copyright © 2025 Cru. All rights reserved.
+//
+
+import Combine
+import Foundation
+
+protocol GetCheckNotificationStatusInterface {
+
+    func permissionStatusPublisher() -> AnyPublisher<String, Never>
```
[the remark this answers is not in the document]
JasonBennett41: Sure thing

godtools/App/Features/OptInNotification/Presentation/OptInNotificationView/OptInNotificationViewModel.swift
```diff
@@ -0,0 +1,225 @@
...
+                [weak self]
+                permissionGranted in
+                guard let self = self else { return }
+
+                if permissionGranted {
+                    // Theoretically should never happen because a user who has granted permissions shouldn't end up in this view
+
+                    self.isActive = false
+
+                } else {
```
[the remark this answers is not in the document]
JasonBennett41: Got it

godtools/App/Features/OptInNotification/Presentation/OptInNotificationView/OptInNotificationView.swift
```diff
@@ -0,0 +1,155 @@
...
+struct OptInNotificationView: View {
+    @ObservedObject private var viewModel: OptInNotificationViewModel
+
+    init(viewModel: OptInNotificationViewModel) {
+        self.viewModel = viewModel
+    }
+
+    var body: some View {
+
+        VStack {
```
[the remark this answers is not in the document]
JasonBennett41: So I actually had this bit in there, but I ended up removing since everything in the view is centered.  There didn't seem to be a difference for centered text, see attached examples I just spun up.

with_modifier

without_modifier

godtools/App/Flows/App/AppFlow.swift
```diff
@@ -1131,6 +1138,42 @@ extension AppFlow {
...
+        appDiContainer.feature.optInNotification.domainLayer.getViewOptInDialogUseCase().viewPublisher(appLanguage: appLanguage)
+            .receive(on: DispatchQueue.main).sink {
+                [weak self]
+                (domainModel: ViewOptInDialogDomainModel) in
+
+                guard let weakSelf = self else {
+                    return
+                }
+
+                let view = weakSelf.getOptInNotificationDialogView( domainModel: domainModel, userDialogReponse: userDialogReponse)
```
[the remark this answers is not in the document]
JasonBennett41: Do any actions need to be taken here?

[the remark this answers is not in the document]
levieggertcru: Nothing needed here.

godtools/App/Features/OptInNotification/Data/LastPromptedOptInNotificationRepository/Cache/OptInNotificationsUserDefaultsCache.swift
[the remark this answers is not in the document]
levieggertcru: Hey Jason I see your thinking behind that now and why that was intentional.  You know, let's just leave it as is.  It's really unlikely for this case to ever occur unless there is some type of programmer error by writing an invalid string.

godtools/App/Features/OptInNotification/Presentation/OptInNotificationView/OptInNotificationView.swift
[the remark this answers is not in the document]
levieggertcru: Hey Jason I'd say just go ahead and add it anyways.  I'm going to look at this some more too.  Maybe it's inheriting from the DashboardView.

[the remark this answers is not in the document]
JasonBennett41: Roger that 👍🏽

godtools/App/Features/OptInNotification/Data/LastPromptedOptInNotificationRepository/Cache/OptInNotificationsUserDefaultsCache.swift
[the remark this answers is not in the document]
JasonBennett41: I changed this to the following:

```
guard let lastPromptedDate: Date =
                OptInNotificationsUserDefaultsCache.dateFormatter.date(
                    from: lastPrompted
                )
        else {
            assertionFailure("An error occurred while parsing \(lastPromptedCacheKey) from cache")
            return nil
        }
```

This way in the unlikely event of a parsing failure, a user who has not granted permissions would still receive the prompt as long as they haven't exceeded 'maxPrompts'.  In this scenario, the bad data would be overridden in shouldPromptNotificationSheet().  The only foreseeable downside to this would be that it negates the time interval, meaning if bad data was written in yesterday's user session, the prompt may be shown again in today's session (and thus the bad data fixed).  Seems preferable to the alternative which would result in a user never being shown the prompt if they have bad data in cache.

The bad data would persist for users who have already enabled notifications, but they would never receive a mistaken prompt.  It would just continue to fall through to the assertionError block but shouldPromptNotificationSheet() will terminate execution at this point:

`guard notificationStatus != "Approved" else { return }`

Is this suitable or would you like me to revert to original / add additional re-record logic for the cache?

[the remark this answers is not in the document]
JasonBennett41: @levieggertcru Requested changes have been performed in the latest commit.  I believe the only pending question is whether you are happy with the current handling of cache parsing failure.  Let me know if there's anything else you need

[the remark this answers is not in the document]
levieggertcru: Yes, I'm liking this cache change better.

levieggertcru (approved):
Hey Jason looks good!  Thank you

PR #2538, opened 2025-04-09
GT-1719 remove invalid test

levieggertcru:
Removing this test because this is a scenario that should never take place.

In this test the existing favorited tools were all at position 0, however, they didn't have a set createdAt date which would never be a scenario in the app.  When storing tool in favorites, the createdAt date is set correctly.


levieggertcru:
Hey @rachaelblue this was the hang-up I had mentioned.  Will merge this change in so be sure to pull.

PR #2539, opened 2025-04-09
Favorited Changed Publisher Test

rachaelblue:
Hey @levieggertcru, here's the bug I was seeing before I moved the onChanged publisher inside the `getFavoritedResourcesSortedByPositionPublisher()` method.

### DIFF
--- a/godtools/App/Features/Favorites/Data-DomainInterface/GetYourFavoritedToolsRepository.swift
+++ b/godtools/App/Features/Favorites/Data-DomainInterface/GetYourFavoritedToolsRepository.swift
@@ -28,12 +28,13 @@ class GetYourFavoritedToolsRepository: GetYourFavoritedToolsRepositoryInterface 
     
     func getToolsPublisher(translateInLanguage: AppLanguageDomainModel, maxCount: Int?) -> AnyPublisher<[YourFavoritedToolDomainModel], Never> {
         
-        return Publishers.CombineLatest3(
+        return Publishers.CombineLatest4(
+            favoritedResourcesRepository.getFavoritedResourcesChangedPublisher(),
             resourcesRepository.getResourcesChangedPublisher(),
             getToolListItemInterfaceStringsRepository.getStringsPublisher(translateInLanguage: translateInLanguage),
             favoritedResourcesRepository.getFavoritedResourcesSortedByPositionPublisher()
         )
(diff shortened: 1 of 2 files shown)


godtools/App/Features/Favorites/Data-DomainInterface/GetYourFavoritedToolsRepository.swift
```diff
@@ -28,12 +28,13 @@ class GetYourFavoritedToolsRepository: GetYourFavoritedToolsRepositoryInterface
     
     func getToolsPublisher(translateInLanguage: AppLanguageDomainModel, maxCount: Int?) -> AnyPublisher<[YourFavoritedToolDomainModel], Never> {
         
-        return Publishers.CombineLatest3(
+        return Publishers.CombineLatest4(
+            favoritedResourcesRepository.getFavoritedResourcesChangedPublisher(),
```
levieggertcru, line 32: Hey @rachaelblue the issue here is that ```favoritedResourcesRepository.getFavoritedResourcesSortedByPositionPublisher()``` sits on the same level as ```favoritedResourcesRepository.getFavoritedResourcesChangedPublisher()```.  So when the collection change is triggered, fetching the favorited resources won't be called here.

Here's an example where you observe changes on a realm collection and then in flatMap fetch the results again. (https://github.com/CruGlobal/godtools-swift/blob/develop/godtools/App/Features/AppLanguage/Data-DomainInterface/GetDownloadedLanguagesListRepository.swift#L33).

I do like the comment you made about expecting a publisher to basically be alive and continually respond with changes.  To me that makes sense so go ahead and observing changes within the Realm cache.  I'm going to make a ticket and see about other areas where we observe collection changes and see if those can be moved into the realm cache level where there are publishers.

Another thing I need to double check is invoking ```prepend()```  when observing collection changes.  There may be situations when you attempt to observe collection changes it isn't triggered where invoking ```prepend()``` will cause it to trigger right away.

From their docs:
(https://www.mongodb.com/docs/atlas/device-sdks/sdk/swift/react-to-changes/#register-a-collection-change-listener)

```
You can register a notification handler on a collection within a realm.

Realm notifies your handler:

After first retrieving the collection.

Whenever a write transaction adds, changes, or removes objects in the collection.

Notifications describe the changes since the prior notification with three lists of indices: the indices of the objects that were deleted, inserted, and modified.
```

levieggertcru: I created this ticket to review where some of those collection change observers are.  (https://jira.cru.org/browse/GT-2555)

rachaelblue: Gonna go ahead and close out this ticket.

PR #2540, opened 2025-04-11
GT Ruby version 3.4.0

levieggertcru:


frett:
@levieggertcru I did some quick googling earlier, it doesn't appear that fast Lane supports ruby 3.4.x yet without some workarounds.

I can find the open issue back if it would help.

I feel like this probably just needs to wait for now

levieggertcru:
Okay thanks @frett for looking at that!  Yeah looks like here (https://github.com/fastlane/fastlane/issues/29183)

PR #2546, opened 2025-04-14
GT-1793 add support for italics in renderer

levieggertcru:
I believe using ```size: 0``` is okay here.  Seems to maintain font size. (https://developer.apple.com/documentation/uikit/uifont/init(descriptor:size:))

Before:

before-italic

After:

supporting-italic

### DIFF
--- a/godtools/App/Services/Renderer/Views/MobileContent/Views/ContentText/MobileContentTextViewModel.swift
+++ b/godtools/App/Services/Renderer/Views/MobileContent/Views/ContentText/MobileContentTextViewModel.swift
@@ -27,6 +27,14 @@ class MobileContentTextViewModel: MobileContentViewModel {
         super.init(baseModel: textModel, renderedPageContext: renderedPageContext, mobileContentAnalytics: mobileContentAnalytics)
     }
     
+    private var isBold: Bool {
+        return textModel.textStyles.contains(.bold)
+    }
+    
+    private var isItalic: Bool {
+        return textModel.textStyles.contains(.italic)
+    }
+    
     var startImage: UIImage? {
         
         guard let resource = textModel.startImage else {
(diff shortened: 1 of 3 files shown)


godtools/App/Services/Renderer/Views/MobileContent/Views/ContentText/MobileContentTextViewModel.swift
```diff
@@ -126,14 +134,8 @@ class MobileContentTextViewModel: MobileContentViewModel {
...
     private func getFontWeight() -> UIFont.Weight {
-             
-        let textStyles: Set<Text.Style> = textModel.textStyles
-        
-        if textStyles.contains(.bold) {
-            return .bold
-        }
         
-        return .regular
+        return isBold ? .bold : .regular
```
frett, line 138: Not in the scope of this PR, but I've deprecated `TextStyle.Bold` in favor of `text.fontWeight`.

in the shared library I handle the fallback of bold -> font-weight for you, so once you implement support for `font-weight` you will no longer need to check for the `bold` text style

frett: I had wanted to handle the raw font-weight to `UIFont.Weight` conversion in the shared module for you, but I couldn't find the `UIFont.Weight` constants when I looked

levieggertcru: Hey thanks @frett.  Do you know if there's a ticket for upgrading to use ```text.fontWeight```?

frett: there is tickets created for it yet, the only use case for it currently is web. And I don't have an easy implementation path on Android with the current renderer views.

PR #2560, opened 2025-04-21
GT-2411 should persist tool settings

rachaelblue:
Hey @levieggertcru, wasn't sure how to go about this without breaking the normal pattern.  We only need the use case if the bool is true, so I wondered if it'd work to just pass the use case wrapped in a model instead.  Thoughts?


levieggertcru:
Hey @rachaelblue had to refresh myself on this task.  But, it appears that these persisted language settings are tied to the user's favorited tool correct?

I think it would be best to inject an interface into TractViewModel for persisting language settings choices.  This way in the future if there's a need to persist tool language settings based on another condition we can do so from the Flow.  For example, maybe we want to persist tool language settings that are based on the all tools.

Can you create an interface ```PersistToolLanguageSettingsInterface``` with a method for persisting the tool languages that TractViewModel can point to instead.

rachaelblue:
Hey @levieggertcru, yeah it's only supposed to persist the settings if viewing from the favorites tab.  Updated to use an interface.

levieggertcru (approved):
Thanks @rachaelblue!

PR #2563, opened 2025-04-23
GT 2567 fix realm crash

levieggertcru:
Changes in this PR:

- Updated favorited tools realm migration to use Combine Publisher and make call to writePublisher extension. I think this will resolve the crash.
- Use the fetch publisher instead of onChange publisher in FavoritedToolsLatestToolDownloader since the fetch publisher will respond to changes. 

gt-realm-crash

### DIFF
--- a/godtools/App/Share/Data/FavoritedResourcesRepository/Cache/RealmFavoritedResourcesCache.swift
+++ b/godtools/App/Share/Data/FavoritedResourcesRepository/Cache/RealmFavoritedResourcesCache.swift
@@ -56,25 +56,18 @@ class RealmFavoritedResourcesCache {
             .object(ofType: RealmFavoritedResource.self, forPrimaryKey: id) != nil
     }
     
-    func getFavoritedResourcesSortedByPosition() -> [FavoritedResourceDataModel] {
-        
-        return getFavoritesSortedByPosition()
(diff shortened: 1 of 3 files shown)


godtools/App/Share/Data/FavoritedResourcesRepository/Cache/RealmFavoritedResourcesCache.swift
```diff
@@ -191,43 +184,48 @@ class RealmFavoritedResourcesCache {
...
     
     @available(*, deprecated)
-    private func migrateExistingFavoritesToPositions() -> [RealmFavoritedResource] {
-        let realm = realmDatabase.openRealm()
-        let favoritesSortedByCreatedAt = realm
-            .objects(RealmFavoritedResource.self)
-            .sorted(byKeyPath: #keyPath(RealmFavoritedResource.createdAt), ascending: false)
-                    
-        do {
-            try realm.write {
```
levieggertcru, line 218: Hey @rachaelblue I'm about 95% certain this write crash was caused because the realm onChange is also triggered on write transactions. So Realm complained that there was a write within a write.

I've seen this happen before elsewhere.  That's one of the things that prompted me to create the RealmDatabase extensions and the ```writePublisher``` because that publisher will always create a new realm instance on another thread so it's confined within it's own thread.

PR #2569, opened 2025-04-25
GT-2505 resume lesson progress modal

rachaelblue:


godtools/App/Flows/ToolNavigation/ToolOpenedFrom.swift
```diff
@@ -0,0 +1,18 @@
...
+
+import Foundation
+
+enum ToolOpenedFrom {
+    case dashboardLessons
+    case dashboardFavoritesFeaturedLesson
+    case dashboardFavoritesFavoritedTool
+    case dashboardTools
+    case deepLink
+    case learnToShare
```
rachaelblue, line 17: Hey @levieggertcru, do you think these should be more granular?  Like distinguishing between opening from toolDetails,  whether learnToShare came from tools or favorites, etc.

levieggertcru: Hey @rachaelblue no let's not worry about getting too granular right now.  That will be enough to cover this related task.  

If we do need more granularity for any reason we can introduce that then.

levieggertcru (approved):

PR #2589, opened 2025-05-07
Remote Config Integration & Opt In Notification Dialog removal

JasonBennett41:
Removed secondary dialog in favor of dynamic 'allow notifications' vs 'notification settings' button.  To accomplish this, a boolean 'isInitialPrompt' is added to the ViewOptInNotificationDomainModel, and the use case fetches the value for this.

Plugged in the remote config repository which seems to be properly fetching the values.

Also removed all the logic for the OptInNotificationDialog, made some minor adjustments to the layout, and replaced the notification graphic pngs.


godtools/App/Share/Data/RemoteConfigRepository/RemoteDatabase/FirebaseRemoteConfigWrapper.swift
```diff
@@ -47,7 +47,10 @@ class FirebaseRemoteConfigWrapper: RemoteConfigRemoteDatabaseInterface {
         
         return RemoteConfigDataModel(
             globalActivityIsEnabled: remoteConfig.configValue(forKey: "ui_account_globalactivity_enabled").boolValue,
-            toolContentFeaturePageCollectionPageEnabled: remoteConfig.configValue(forKey: "tool_content_feature_page_collection_page_enabled").boolValue
+            toolContentFeaturePageCollectionPageEnabled: remoteConfig.configValue(forKey: "tool_content_feature_page_collection_page_enabled").boolValue,
+            optInNotificationEnabled: remoteConfig.configValue(forKey: "ui_opt_in_notification_enabled").boolValue,
+            optInNotificationPromptLimit: Int(truncating: remoteConfig.configValue(forKey: "ui_opt_in_notification_prompt_limit").numberValue),
+            optInNotificationTimeInterval: Int(truncating: remoteConfig.configValue(forKey: "ui_opt_in_notification_time_interval").numberValue)
```
levieggertcru, line 53: Good call on the Int(truncating to be safe.

godtools/App/Features/OptInNotification/Presentation/OptInNotification/OptInNotificationViewModel.swift
```diff
@@ -14,6 +14,8 @@ import UserNotifications
 
 class OptInNotificationViewModel: ObservableObject {
 
+    var isInitialPrompt: Bool = false
```
levieggertcru, line 17: Hey @JasonBennett41 looks like this will need to be a Published property since the SwiftUI View is accessing this property.

levieggertcru (changes requested):
Hey @JasonBennett41 looks good.  Just 1 change request on making ```var isInitialPrompt: Bool = false``` a Published property.  I noticed the SwiftUI View was accessing that value.

levieggertcru (approved):

PR #2591, opened 2025-05-07
GT-2598 update logic to convert colormath colors to UIColors

frett:


levieggertcru:
Hey @frett so far looks good to me.  I noticed that running locally the build fails for me, however in GitHub Actions the build didn't fail which is strange.

Locally it's failing on the Manifest navBarControlColor property which says it's of type ```id``` in the header. 

build-failed-locally

navBarControlColor-type-id

frett:
Still very much a work in progress, even the shared parser updates are ongoing. I probably pushed some other stuff I was working on after the gha build which updated the snapshot dependency you ended up pulling

frett:
I've had to do bulk type updates to all the color properties because of interdependence between them, but have only finalized the text color property. I'm probably referencing the text color or primaryTextColor, but haven't converted to the right type at the moment

levieggertcru:
Ah that makes sense as to why it wasn't failing in GitHub Actions.  Forgot that I could be pulling in a different snapshot.

levieggertcru:
Hey @frett these changes look good, for some reason though I was still getting errors when running locally.  I'll post some screenshots today.  It was giving me an error in places where ```toUIColor()``` was invoked because the attribute invoking toUIColor was of type ```UIColor```.

frett:
@levieggertcru ok, I'm planning on waiting to merge this until the cyoa web work gets merged anyway. I want to make sure they have everything they need before this work lands

levieggertcru:
Hey @frett does the Podfile need to be updated? 

Got this error running a ```pod install``` after a ```pod repo update```.

Screenshot 2025-05-14 at 4 22 15 PM

levieggertcru:
Hey @frett this one has merge conflicts with the Podfile.  Do you want to keep those changes or merge from develop?

frett:
@levieggertcru I'm waiting to address the conflicts until the shared PR is merged

frett:
@levieggertcru I updated the PR to use 1.2.0-SNAPSHOT, which should be more stable

levieggertcru (approved):

PR #2613, opened 2025-05-16
GT-2621 Add language code ha to app languages

levieggertcru:


levieggertcru:
Once merged to dev will need to run OneSky workflow to pull in localizable strings for ```ha```

PR #2623, opened 2025-05-20
GT Starting to add compose UIViewController in MobileContentStackView

levieggertcru:


levieggertcru:
So far was able to build and run the project.  @frett is looking into next path forward to resolve an Exception.

"Uncaught Kotlin exception: kotlin.IllegalStateException: Error: `Info.plist` doesn't have a valid `CADisableMinimumFrameDurationOnPhone` entry. 
This will result in an inadequate performance on iPhones with high refresh rate. 
Add `<key>CADisableMinimumFrameDurationOnPhone</key><true/>` entry to `Info.plist` to fix this error. 
If you don't have a separate plist file, add the entry to the target from within Xcode: Project -> Targets -> Info -> Custom iOS Target Properties.                            
Or set `ComposeUIViewController(configure = { enforceStrictPlistSanityCheck = false }) { .. }`, if it's intended. "

PR #2645, opened 2025-06-04
GT-2287 Lock ScrollView height to content size

levieggertcru:
This PR updates the offline languages view to lock the scrollview height to the content size.

offline-languages-many

offline-languages-none

offline-languages-2


levieggertcru:
Hey @rachaelblue thought I would re-visit this since I just faced a similar scenario updating the ToolSettingsView so the ScrollView wraps its content.  It's a bit janky, but, I came across a stack overflow solution for having a ScrollView wrap it's inner content.  The trick is to use a background modifier with GeometryReader to calculate the content size and store that in a state variable.

```swift
.background(
                        GeometryReader { geo -> Color in
                            DispatchQueue.main.async {
                                contentSize = geo.size
                            }
                            return Color.clear
                        }
)
```
Then on the ScrollView set the maxHeight to the contentSize height.

```swift
.frame(maxHeight: contentSize.height)
```

PR #2666, opened 2025-06-27
GT-2578 QR code activity item

rachaelblue:
Hey @levieggertcru, I ended up moving `ToolScreenShareFlow` up a level so it's owned by `TractFlow` instead of `ToolSettingsFlow`.  I was running into the issue that the settings view needed to be dismissed when the UIActivityView was open, but doing so deallocated ToolScreenShareFlow, which is what handles the QR code activity tap.


levieggertcru:
Thanks @rachaelblue I'll take a look too.  I know those view presentations can sometimes get in the way.

Was the issue with tapping the QR code activity item was wanting to dismiss the tool settings?  Not sure I understood 100%.

levieggertcru:
Ohh yeah I see, the UIActivityViewController dismissing ToolSettings.

levieggertcru:
Yeah the choose language and share image modals looks good since those are full screen.

choose-language-modal

share-image-modal

levieggertcru:
Yeah those Flow Completed States I added as a way to tell the parent flow that a flow has completed so it can set the child flows to null for clean up.  I wish that clean up could be more automatic.

PR #2668, opened 2025-06-30
GT-2578 fix to completing tool screen share flow

levieggertcru:
Hey @rachaelblue I was actually thinking something else when you originally mentioned the issue you were having.  I didn't realize it was these 2 lines causing tool settings to be closed.
- (https://github.com/CruGlobal/godtools-swift/blob/develop/godtools/App/Flows/ToolScreenShare/ToolScreenShareFlow.swift#L136)
- (https://github.com/CruGlobal/godtools-swift/blob/develop/godtools/App/Flows/ToolSettings/ToolSettingsFlow.swift#L106)

I shouldn't have been closing the tool settings flow when the session was created because the share session UIActivityViewController was still in memory.

I realized we don't want to move that to the TractFlow because the LessonFlow and CYOAFlow implement the ToolSettingsFlow.

All in all I ended up fixing the ToolScreenShareFlow completion logic so it's completed when a user either closes that share modal or completes sharing the QR code.

This means that the boolean flag for checking if the QR code is presented isn't needed so I went ahead and removed that.

I'm also glad you added the completion items handler for UIActivityViewController (https://developer.apple.com/documentation/uikit/uiactivityviewcontroller/completionwithitemshandler-swift.typealias) because I didn't know that existed.  I noticed an issue where tapping the QR code would cause both the QR tap to trigger and dismiss activity view controller to trigger on the ViewModel.  Looking at the complete items handler it will tell us what activity was tapped, so I updated to move the QR tapped to be triggered from the handler instead.


rachaelblue (commented):
Hey @levieggertcru, thanks for fixing this up!  I missed that Lessons and CYOA used the tool settings too, sorry!  Should the "share screen" button on Settings hide once the share session starts?  Since Settings doesn't dismiss automatically now, it's possible to press "Share screen" again after the UIActivityVC dismisses.

levieggertcru:
Hey @rachaelblue, I was thinking of something there too, or changing the icon to make it look active.  At the moment an alert pops up if you attempt to tap the button again.  I'll check with Jill on that and see what she says.

levieggertcru:
And honestly I forgot that the LessonFlow and CYOAFlow had implemented that as well until I was poking around in there some more.

PR #2669, opened 2025-07-03
GT-2656 generate qr code image

rachaelblue:
Hey @levieggertcru, I ended up doing two of the subtasks here-- GT-2656 and GT-2657.  I also pulled your custom modal code from tool settings out into a separate view for reusability.


godtools/App/Flows/Flow/FlowStep.swift
```diff
@@ -212,7 +212,7 @@ enum FlowStep {
     case didCreateSessionFromCreatingToolScreenShareSession(result: Result<WebSocketChannel, TractRemoteSharePublisherError>)
     case cancelTappedFromCreateToolScreenShareSessionTimeout
     case acceptTappedFromCreateToolScreenShareSessionTimeout
-    case shareQRCodeTappedFromToolScreenShareSession
+    case shareQRCodeTappedFromToolScreenShareSession(shareUrl: String)
```
rachaelblue, line 215: Hey @levieggertcru, would you prefer to regenerate the shareUrl for the QR code view instead of passing it?

godtools/App/Share/SwiftUI Views/GTModalView.swift
```diff
@@ -0,0 +1,76 @@
...
+//  GTModalView.swift
+//  godtools
+//
+//  Created by Rachael Skeath on 7/3/25.
+//  Copyright © 2025 Cru. All rights reserved.
+//
+
+import SwiftUI
+
+struct GTModalView<Content: View>: View {
```
levieggertcru, line 11: I like this.  It would be good to use this for the opt-in notification view and tool settings view.  I'm going to create a ticket for that.

levieggertcru (approved):

PR #2674, opened 2025-07-15
GT-2681 create swift tests for supporting files

levieggertcru:


godtoolsTests/App/Share/Data-DomainInterface/Supporting/GetTranslatedLanguageName/GetTranslatedLanguageNameTests.swift
```diff
@@ -6,203 +6,216 @@
...
+    }
+    
+    @Test(
+        """
+        Given: User is viewing the french language name.
+        When: Translating the french language name in spanish and forceLanguageName is false and a translation is available in the app bundle's string phrases.
+        Then: Translate the french language name using the app bundle's string phrases.
+        """
+    )
+    func usesAppBundlesStringPhrases() {
```
levieggertcru, line 134: Hey @rachaelblue I'm starting to add tests for the Share/Data-DomainInterface/Supporting.  I transitioned the current tests from Quick and Nimble to Swift Tests.  What do you think of supplying Given, When, Then in the ```Test``` description?

One thing I do like about Quick and Nimble is we don't have to define a method.  In SwiftTests I have to define a method which can get wordy.

levieggertcru: Ahh yeah the comment would be a good place for additional clarification.

PR #2685, opened 2025-07-21
string searchable test

rachaelblue:


godtoolsTests/App/Share/Common/StringSearcher/StringSearcherTests.swift
```diff
@@ -0,0 +1,58 @@
...
+    private static let englishSearchable = MockStringSearchable(searchableStrings: ["English", "English"])
+    private static let albanianSearchable = MockStringSearchable(searchableStrings: ["Albanian", "shqip"])
+    private static let croatianSearchable = MockStringSearchable(searchableStrings: ["Croatian", "hrvatski"])
+    private static let englishBahrainSearchable = MockStringSearchable(searchableStrings: ["English (Bahrain)", "English (Bahrain)"])
+    private static let frenchSearchable = MockStringSearchable(searchableStrings: ["French", "français"])
+    private static let spanishSearchable = MockStringSearchable(searchableStrings: ["Spanish", "español"])
+    
+    @Test(
+        """
+        Given: A list of StringSearchable items, where each StringSearchable contains a list of searchable strings
```
levieggertcru, line 33: Hey @rachaelblue thanks for getting started on these tests! 

Some things to note, so far for tests we have 3 categories:

1. Unit Tests (TDD) - Still needs planning and documenting.
2. Behavior Tests (BDD) - Documented in the ReadMe (https://github.com/CruGlobal/godtools-swift/pull/2681)
3. UI Navigation Tests - Still needs planning and documenting.

The (Given, When, Then) I see mainly used in Behavior Tests.  Which would fall into testing app behavior and business rules.  Mostly these are domain-interface implementations.  I still need to figure out how we can test more complex interactions where multiple pieces are working together. 

I do see StringSearcher as more of a unit test since it's generic and placed in our Share/Common/. However, I'm fine with using (Given, When, Then) here since it does a good job at setting the stage and outcome.  I'm just wondering if we should use some other language for unit tests.  What's your experience with unit tests and TDD?  I still need to do more research here myself.  But, I would like to document something in the ReadMe for unit tests.

Lastly, looking at StringSearcher.swift there are a couple additional tests I think would be good to add so we're covering all the outcomes.
1. An empty search string should give back an empty array (https://github.com/CruGlobal/godtools-swift/blob/develop/godtools/App/Share/Common/StringSearcher/StringSearcher.swift#L17).
2. It appears that StringSearcher will ignore case sensitivity on both search string and searching against.  Some tests that case sensitivity should be ignored would also be good.  Your test does do this, I think it would be nice to make that more explicitly known in the ```Then```. (https://github.com/CruGlobal/godtools-swift/blob/develop/godtools/App/Share/Common/StringSearcher/StringSearcher.swift#L23).

levieggertcru (changes requested):

godtoolsTests/App/Share/Common/StringSearcher/StringSearcherTests.swift
```diff
@@ -0,0 +1,58 @@
...
+        When: The StringSearchable list is searched with a search term string
+        Then: StringSearcher should return a list of StringSearchables that have a searchable string matching the search term.
+        """,
+        arguments: [
```
levieggertcru, line 37: Really nice setup of arguments. I like this in Swift Tests.

godtoolsTests/App/Share/Common/StringSearcher/StringSearcherTests.swift
levieggertcru: With the Unit Tests let's not do the (Given, When, Then).  That was something I was reading in behavior tests.  For now let's just have wordy function names.  I need to do some more research into Unit Tests, maybe there's something other developers are doing for that.

For Unit Tests there's a couple things:

1. Tests are written to ensure 100% code coverage. For example, in StringSearcher there is an if conditional to check for an empty search string (https://github.com/CruGlobal/godtools-swift/blob/develop/godtools/App/Share/Common/StringSearcher/StringSearcher.swift#L17C1-L18C1) which returns an empty array. We can ensure that the return line is tracked by adding a test that an empty search string gives back empty results.

2. Tests are written to cover the intended functionality. One piece of functionality I noticed is StringSearcher is that it ignores case sensitivity.  So writing tests to ensure that case sensitivity is ignored.  This could include an all caps search string, and all lower case, mixed case, etc.  I think having the single letter search string is also nice.  Maybe in the future we want to update StringSearcher to take a minimum search length and then an additional test can be added for that.

It's hard to say where to stop sometimes, I think as long as you're covering all possible outcomes and intended functionality that's the main thing. And throwing good inputs at the test to ensure that functionality passes.

A benefit of that is unit tests become self documenting.  Similarly to behavior tests in where we're essentially documenting the business rules.  Unit tests we're documenting how a piece of code should function. 

TDD is also helpful in this process because the failing tests are considered and written ahead of time before implementing any code.  And then code is written to pass those tests.  It can be helpful to set that up ahead of time.

levieggertcru: Hmm yeah StringSearcher doesn't support a different alphabet.  So you're saying in that case an English Search String comes in against a list of Chinese Searchables?

levieggertcru: I wouldn't test that unless that functionality is desired.

godtoolsTests/App/Share/Common/StringSearcher/StringSearcherTests.swift
```diff
@@ -0,0 +1,56 @@
...
+    @Test(
+        "Search a list of StringSearchables with a search term string and return case-insensitive matches",
+        arguments: [
+            TestArgument(searchTerm: "English", expectedSearchResults: [englishSearchable, englishBahrainSearchable]),
+            TestArgument(searchTerm: "Bangla", expectedSearchResults: []),
+            TestArgument(searchTerm: "SH", expectedSearchResults: [englishSearchable, albanianSearchable, englishBahrainSearchable, spanishSearchable]),
+            TestArgument(searchTerm: "french", expectedSearchResults: [frenchSearchable]),
+            TestArgument(searchTerm: "ESPAÑOL", expectedSearchResults: [spanishSearchable]),
+            TestArgument(searchTerm: "ai", expectedSearchResults: [englishBahrainSearchable, frenchSearchable]),
+            TestArgument(searchTerm: "", expectedSearchResults: [englishSearchable, albanianSearchable, croatianSearchable, englishBahrainSearchable, frenchSearchable, spanishSearchable]),
```
levieggertcru, line 40: Hey @rachaelblue another thing I like to do is organize my tests by functionality or intent.  

For example, because the empty search string isn't part of the case sensitivity logic or main search logic, instead it returns immediately with the searchables, I'd create another test method with something like "empty search string should return with original list of searchables". 

Doing this high-lights each piece of functionality or intent of the StringSearcher.

godtoolsTests/App/Share/Common/StringSearcher/StringSearcherTests.swift
rachaelblue: Hey @levieggertcru, I added a case for Amharic, since it's an example where we have 2 different alphabets in the same array.  You should be able to search for "Amharic" or "አማርኛ" and get the same result.

levieggertcru: Ohh very nice!

godtoolsTests/App/Share/Common/StringSearcher/StringSearcherTests.swift
rachaelblue: This is a good idea, I was thinking having all the test arguments lumped together could be hard to follow.

PR #2693, opened 2025-07-25
GT-2692 show last page of screen share tutorial

levieggertcru:


rachaelblue (commented):
Hey @levieggertcru, there's something a little awkward with the navigation here in some places.  I think because there's so many different modals opening and closing.  The biggest weird thing is that sometimes when I press the "share link" button on the "get started" tutorial page, there's a delay after the tutorial closes to open the activity controller.  It's not consistent though so I'm not sure 🤔

levieggertcru:
Thanks @rachaelblue I'm seeing that too.  A fairly noticeable delay before the share activity view controller is presented.  Sometimes its short and sometimes longer.  I'm going to look at that in this Jira task (https://jira.cru.org/browse/GT-2671).  Polishing some of the navigation between tool settings flow and tool screen share flow.

PR #2696, opened 2025-07-25
GT-2700 start migrate quick and nimble to swift tests

levieggertcru:
Hey @rachaelblue this is the very start of Jira task (https://jira.cru.org/browse/GT-2698) which contains 2 sub-tasks for migrating the quick and nimble tests to Swift Tests.

- Adding folder ```godtoolsTests/BehaviorTests/App/**```
- Migrates GetAppLanguagesInterfaceStringsRepositoryTests to Swift Tests.


godtoolsTests/App/Features/AppLanguage/Data-DomainInterface/GetLanguageSettingsInterfaceStringsRepositoryTests.swift
```diff
@@ -6,21 +6,172 @@
...
     
-    override class func spec() {
+    @Test(
+        """
+        Given: User is viewing the language settings.
+        When: The app is switched from English to Spanish.
+        Then: The interface strings should be translated into Spanish.
+        """
+    )
+    @MainActor func interfaceStringsAreTranslatedWhenAppLanguageChanges() async {
```
levieggertcru, line 22: Hey @rachaelblue because the TestsRealmDatabase by default uses a single realm instance on the main thread, there will be a crash when the realm instance is observing collection changes.
```*** Terminating app due to uncaught exception 'RLMException', reason: 'Can only add notification blocks from within runloops.'
terminating due to uncaught exception of type NSException```

To ensure there isn't a crash the Test will need to be marked with the ```MainActor``` to keep execution on the main thread.

rachaelblue: Thanks @levieggertcru!

PR #2730, opened 2025-08-07
GT-2687 user account details tests

rachaelblue:


levieggertcru:
Thanks @rachaelblue that's good to know about the interfaces.  That will be very handy.

godtoolsTests/BehaviorTests/App/Features/Account/Data-DomainInterface/GetUserAccountDetailsRepositoryTests.swift
```diff
@@ -0,0 +1,143 @@
...
+                    confirmation()
+                    userAccountDetails = result
+                }
+                .store(in: &cancellables)
+            
+            let locale = Locale(identifier: argument.appLanguage.rawValue)
+            let createdAtDateString = Self.getDateFormatter(locale: locale).string(from: Self.userCreatedAt)
+            let joinedOnStringExpected = "\(argument.joinedOnString) \(createdAtDateString)"
+            
+            #expect(userAccountDetails?.name == Self.userFullName)
```
levieggertcru, line 59: Hey @rachaelblue the joined on date test looks good to me.

I'm wondering for the name if we should break that into a separate test?  I was looking at the logic in UserAccountDetailsRepository and there is some specific logic for breaking down the name (https://github.com/CruGlobal/godtools-swift/blob/develop/godtools/App/Features/Account/Data-DomainInterface/GetUserAccountDetailsRepository.swift#L67).

levieggertcru: Actually, maybe the name test is fine as is.  Some of these tests to me are hard to define.

godtoolsTests/Mock/App/Share/Data/MobileContentAuthTokenRepository/Cache/MockMobileContentAuthTokenKeychainAccessor.swift
```diff
@@ -0,0 +1,48 @@
...
+    private var userId: String?
+    private var authToken: MobileContentAuthTokenDataModel?
+    private var appleRefreshToken: String?
+    
+    func setUserId(_ userId: String?) {
+        self.userId = userId
+    }
+}
+
+extension MockMobileContentAuthTokenKeychainAccessor: MobileContentAuthTokenKeychainAccessorInterface {
```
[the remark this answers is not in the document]
rachaelblue: The file was originally AI generated, yes.  I moved the interface implementation into the extension and simplified the methods.

godtoolsTests/BehaviorTests/App/Features/Account/Data-DomainInterface/GetUserAccountDetailsRepositoryTests.swift
rachaelblue: Well now that you say that, I'll go back and add some extra name tests.  Didn't hit all of those cases.  I need to remember to check the code coverage for each file!

levieggertcru: Okay thanks it doesn't hurt!
I was getting boggled thinking about it for a second because I know some of that logic is dependent on the data layer and what it gives us. Where as the use case domain model itself is just a single name field.

rachaelblue: Yeah I gotta remember to look at the repo logic itself, and not just the fields in the domain model.

levieggertcru: What makes it more tricky is some of those rules we need defined from the product or business analytics side.  Those are the tests we really want to capture in behavior tests. Typically I get that from Figma if there are detailed comments on app behavior.  Or, sometimes it's just up to us developers.

levieggertcru (approved):

PR #2733, opened 2025-08-11
GT-2722 create tests for share common

levieggertcru:


godtoolsTests/App/Share/Common/FileCache/FileCacheTests.swift
```diff
@@ -0,0 +1,253 @@
...
+//
+//  Created by Levi Eggert on 8/12/25.
+//  Copyright © 2025 Cru. All rights reserved.
+//
+
+import Testing
+@testable import godtools
+import Foundation
+
+@Suite(.serialized)
```
levieggertcru, line 13: Hey @rachaelblue here I run the tests serially instead of parallel since accessing the shared FileCache could cause unexpected results if running in parallel when adding and deleting the temp_directory. 

Some pretty good info on migrating from XCTests to Swift Tests (https://developer.apple.com/documentation/testing/migratingfromxctest)

godtoolsTests/App/Share/Common/FileCache/FileCacheTests.swift
```diff
@@ -0,0 +1,253 @@
...
+        
+        switch fileCache.getRootDirectory() {
+        case .success( _):
+            break
+        case .failure(let error):
+            throw error
+        }
+                
+        let optionalFileData: Data? = "text data".data(using: .utf8)
+        let fileData: Data = try #require(optionalFileData)
```
levieggertcru, line 106: @rachaelblue this is also a neat way to unwrap an optional using require and fail the test if the value is null since we can't proceed any further.

PR #2800, opened 2025-10-02
GT-2698 migrate quick and nimble

rachaelblue:


godtools/App/Share/Data/FavoritedResourcesRepository/Cache/RealmFavoritedResourcesCache.swift
```diff
@@ -149,6 +149,9 @@ class RealmFavoritedResourcesCache {
     func reorderFavoritedResourcePublisher(id: String, originalPosition: Int, newPosition: Int) -> AnyPublisher<[FavoritedResourceDataModel], Error> {
         
         return realmDatabase.writeObjectsPublisher { realm in
+            
+            guard realm.object(ofType: RealmFavoritedResource.self, forPrimaryKey: id)?.position == originalPosition else { return [] }
```
levieggertcru, line 153: Hey @rachaelblue I was curious if this fixed a bug or another issue?

It makes sense that it may not be good to continue if the originalPosition is no longer matching the position in the database.

levieggertcru (approved):

PR #3274, opened 2026-08-21
GT-3041 task clean up first pass

levieggertcru:
First pass of Task clean up.
- Uses Task.detached for one off tasks that need to complete.
- Uses weak self capture list when capturing task to prevent retain cycles.

### DIFF
--- a/godtools/App/Features/Shareables/Presentation/ShareShareable/ShareShareableViewModel.swift
+++ b/godtools/App/Features/Shareables/Presentation/ShareShareable/ShareShareableViewModel.swift
@@ -41,7 +41,9 @@ extension ShareShareableViewModel {
     
     func pageViewed() {
         
-        Task {
+        let incrementUserCounterUseCase: IncrementUserCounterUseCase = self.incrementUserCounterUseCase
+        
+        Task.detached {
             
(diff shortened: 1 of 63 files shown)


godtools/App/Features/Shareables/Presentation/ShareShareable/ShareShareableViewModel.swift
```diff
@@ -41,7 +41,9 @@ extension ShareShareableViewModel {
     
     func pageViewed() {
         
-        Task {
+        let incrementUserCounterUseCase: IncrementUserCounterUseCase = self.incrementUserCounterUseCase
+        
+        Task.detached {
```
levieggertcru, line 46: Hey @rachaelblue just a heads up, I'm doing some task clean up as the codebase shifts from Combine to async await.  In Combine we kept a static reference to cancellables to keep subscriptions alive beyond ViewModel life cycle for background work.  Using Task.detached is the equivalent of that.  It runs on a global actor and will stay alive beyond the ViewModel life cycle.  The caveat is you can't access self anymore so you have to capture those locally as in this example here. incrementUserCounterUseCase is captured locally and used in Task.detached.

PR #3298, opened 2026-08-25
Fix JSON:API filter parameter names

frett:
I don't believe you are using the `resource_type` parameter, but if you are that should be updated to `filter[resource-type]` after https://github.com/CruGlobal/mobile-content-api/pull/2036 lands. I also implemented multi-value support for the resource type, it takes it as a comma separated list, `filter[resource-type]=tract,cyoa`

This PR is currently safe to merge before the API PR lands. The bare parameters are not supported by the JSON:API spec, and the API already supports the correct `filter[*]` parameters that are part of the JSON:API spec. The `filter[resource-type]=a,b` parameter is not supported until after the API PR merges


frett:
it looks like you might have the `filter[` + `]` name part extracted to your request-operation-ios library, so I'm going to close this

PR #3304, opened 2026-08-28
GT-2839 selected country floats to top

rachaelblue:


godtools/App/Flows/LocalizationSettings/LocalizationSettingsFlow.swift
```diff
@@ -57,17 +57,17 @@ final class LocalizationSettingsFlow: GTFlow {
...
-            }
-            
             if userShouldConfirmSelectedCountry {
                
                 presentView(
                     view: getLocalizationSettingsConfirmationView(selectedCountry: countryListItem),
                     animated: true
                 )
             }
+            else if shouldStoreCountryWhenSelected {
```
levieggertcru: Hey @rachaelblue I see why this change was made since userShouldConfirmSelectedCountry equals true and is never set from the app anywhere.  Looks like some left over dead code from a prior change.  Can you remove that attribute userShouldConfirmSelectedCountry on the Flow since it's always true.  Can also remove storing selected country here and just present the getLocalizationSettingsConfirmationView.

godtoolsUITests/App/Flows/LocalizationSettingsFlow/LocalizationSettingsFlowTests.swift
```diff
@@ -47,4 +47,96 @@ final class LocalizationSettingsFlowTests: BaseFlowTests {
         
         assertIfScreenDoesNotExist(screenAccessibility: .localizationSettings)
     }
+    
+    func testConfirmingSelectedCountryMovesItToTopOfList() {
```
levieggertcru: Hey @rachaelblue the behavior tests look good in the use case tests.

This testing here in the flow doesn't belong and these changes can be reset. Flow tests are strictly for testing navigation when tapping buttons and items from the UI, asserting that the screen to navigate to exists.

levieggertcru (changes requested):
Hey @rachaelblue looks good.  Just an update needed on some left over dead code and removing the flow test.

rachaelblue:
Thanks @levieggertcru, finally got back this and made those changes.

levieggertcru (approved):

PR #3305, opened 2026-08-28
GT-2926 Show Personalization Unavailable

rachaelblue:
I think the business logic here changed from what we expected.  Jill's comment on the ticket:

Currently if you have a country and language selected but there is no data for that combination it shows the "Displaying localized Lesson list" (blue background). We would like it to instead show the "Personalization unavailable" (grey background - "Fallback 2 - option b" in the figma design). Anytime there is no data available for any combination the Personalization unavailable option should be displayed.


levieggertcru (approved):
Hey @rachaelblue this looks like a good start.

